PR #70, opened 2026-05-16
fix: improve ffmpeg detection on Windows, increase yt-dlp timeout, add proxy setting

SummerSec:
## Summary

Three backend fixes for download/extraction reliability:

### 1. FFmpeg detection on Windows
`find_ffmpeg()` only checked macOS well-known paths. Added `D:\ffmpeg\bin\ffmpeg.exe`, `C:\ffmpeg\bin\ffmpeg.exe`, and `C:\Program Files\ffmpeg\bin\ffmpeg.exe` so Windows users with standalone ffmpeg installs are detected even when PATH is not inherited by the backend subprocess.

### 2. yt-dlp download timeout
Increased `socket_timeout` from 30s → 120s and added `retries: 3` / `fragment_retries: 3` for flaky connections.

### 3. Error message improvement
Audio extraction errors no longer show as empty string `""` — when ffmpeg stderr is empty, the exit code is used; when the exception has no message, the exception type is shown instead.

### 4. Proxy support
Added `HTTP_PROXY` / `HTTPS_PROXY` to the allowed env-var keys in `/system/set-env` so users can set a download proxy for yt-dlp and HuggingFace.

## Test plan
- [ ] Backend starts without ffmpeg-related errors on Windows
- [ ] YouTube URL import does not timeout on slow connections
- [ ] POST /system/set-env with key=HTTP_PROXY succeeds

SummerSec

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring HTTP and HTTPS proxy settings

* **Bug Fixes**
  * Enhanced network resilience with improved timeout and retry logic for downloads
  * Improved error messages for audio extraction failures
  * Expanded ffmpeg detection across macOS, Windows, and Linux systems

[Review Change Stack](https://app.coderabbit.ai/change-stack/debpalash/OmniVoice-Studio/pull/70?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/api/routers/system.py
+++ b/backend/api/routers/system.py
@@ -518,7 +518,7 @@ async def set_env_var(body: dict):
     The value is set on os.environ for the running process.
     For persistence across restarts, users should set it in their shell profile.
     """
-    ALLOWED_KEYS = {"HF_TOKEN", "TRANSLATE_API_KEY"}
+    ALLOWED_KEYS = {"HF_TOKEN", "TRANSLATE_API_KEY", "HTTP_PROXY", "HTTPS_PROXY"}
     key = body.get("key", "")
     value = body.get("value", "")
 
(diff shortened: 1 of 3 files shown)


SummerSec:
Merged into PR #66 which now includes all changes.

PR #85, opened 2026-05-18
fix: Windows port conflict + HF_HUB_OFFLINE for offline model loading

caaaaaleb:
## Summary

Three fixes to improve reliability on Windows and restricted networks:

### 1. Windows `kill_orphan_on_port` implementation
The Windows version was an empty function, causing stale backend processes to occupy the port. When a new backend spawned, uvicorn couldn't bind → shutdown → model loading continued in half-dead process → confusing "Could not import module AutoModel" error.

**Fix:** Implement `kill_orphan_on_port` for Windows using `netstat -ano` to find the listening PID and `taskkill /PID /F` to terminate it.

### 2. `HF_HUB_OFFLINE=1` at import time
On restricted networks, lazy imports deep in the HuggingFace stack could trigger `httpx.ConnectTimeout` before `local_files_only=True` takes effect.

**Fix:** Set `HF_HUB_OFFLINE=1` in `main.py` before ANY library import, and in `model_manager._load_model_sync()` with proper save/restore.

### 3. `local_files_only=True` for `from_pretrained`
Added `local_files_only=True` to `OmniVoice.from_pretrained()` call as belt-and-suspenders alongside `HF_HUB_OFFLINE`.

## Tested

- Backend starts and stays running on Windows
- Model loads successfully via Tauri spawn
- `/system/info` returns correct model checkpoint info
- TTS generation works end-to-end

## Summary by CodeRabbit

* **Documentation**
  * README replaced with a Simplified Chinese, Windows-focused user guide: quick-start, install modes (MSI/source/dev), build/MSI notes, ports (3900/3901), FAQ and repo layout.

* **New Features**
  * Full English + Simplified Chinese UI with persistent language toggle.
  * Windows start scripts for production and dev that run checks, launch services, and open the browser.

* **Bug Fixes**
  * Offline model loading and Windows startup safeguards to avoid HF/Triton issues.
  * Auto-kill of orphaned Windows processes on occupied ports.

* **Chores**
  * Dev server bound to 127.0.0.1; MSI build artifacts ignored.

[Review Change Stack](https://app.coderabbit.ai/change-stack/debpalash/OmniVoice-Studio/pull/85?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


debpalash:
Thanks @caaaaaleb — the Windows `kill_orphan_on_port` implementation and the `HF_HUB_OFFLINE` / `local_files_only` belt-and-suspenders are both real fixes we want, and the `_gpu_pool` → `_get_gpu_pool()` catch is excellent work (the module-level `__getattr__` shim was breaking under PyInstaller).

However the PR also rewrites the README into Simplified Chinese, flips the i18n `fallbackLng` from `en` to `zh`, and ships ~3,500 lines of UI string refactoring across 40+ JSX files — those changes affect every user globally and need their own review.

**Could you split this into:**

1. **The four backend/Rust bugfixes** — `backend/main.py`, `backend/services/model_manager.py`, `frontend/src-tauri/src/backend.rs` Windows `kill_orphan`, and the `_get_gpu_pool` fix. We'll merge this quickly.
2. **i18n work** — `zh.json` (+1086 lines), the 40+ JSX `t()` wrappings, and `README.zh-CN.md` added **alongside** the existing English README (not replacing it). Keep `fallbackLng: 'en'` so non-Chinese users don't get Chinese fallback text.
3. **`start.bat` / `start-dev.bat`** scripts as a third PR — the `taskkill /f` against any PID on `:3900` is a useful affordance but should be audited (a user running another service on that port loses it without confirmation).

**One small nit on (1):** in the Windows `kill_orphan_on_port` netstat parser, the `:PORT ` needle with trailing space is built but never actually matched against — falls through to `port_suffix` for every line. Simplify to a single `local_addr.ends_with(&format!(":{}", port))` check.

**Coordination:** The `HF_HUB_OFFLINE=1` work in `main.py` is complementary to our Phase 3 mirror-cascade plan (no conflict). The Windows `kill_orphan` fix unblocks Windows users today and can land independently. The vite `host: '127.0.0.1'` change parallels PR #84's bind-address audit — fine, but worth calling out.

Happy to walk through the split if helpful. Looking forward to landing the backend fixes — they're solid work.

debpalash:
Triage (v0.3.0 line): this PR is currently **conflicting with `main`** (58 files, +4006/-1861) and overlaps the active plan-02 (Windows runtime integrity) and plan-03 (installer network resilience) work. Two of its themes are being addressed root-cause-first there:
- **Windows port conflict** → will be covered by plan-02 (#129)
- **HF_HUB_OFFLINE / offline model loading** → relates to plan-01 (#128, just landed a WinError-448 cache fix in #137) and plan-03 (#130)

Could you **rebase onto current `main`** and split it down to just the parts not already covered by those plans? A focused, rebased PR is much easier to absorb than the current broad diff. Flagging rather than closing — keen to keep the Windows fixes that aren't duplicated.

debpalash:
The three fixes here are genuinely valuable — the Windows `kill_orphan_on_port` implementation and `HF_HUB_OFFLINE` offline loading directly serve the "first-run that actually works on Windows / restricted networks" goal.

The problem is scope: this PR touches **58 files / +4006 / −1861**, but ~50 of those are whole-frontend reformatting churn (App.jsx, AudioTrimmer, every component) unrelated to the fixes. That looks like fork drift / an editor reformat. We can't review or land that safely — it'd conflict heavily with `main` and risks regressing recently-shipped work.

Could you scope it down to just the fix files? Based on the diff that's roughly:
- `frontend/src-tauri/src/backend.rs` — Windows `kill_orphan_on_port` (`netstat -ano`)
- `backend/main.py` + `backend/services/model_manager.py` — `HF_HUB_OFFLINE` offline loading
- `backend/core/personalities.py` (if part of the fix)

A clean branch off current `main` with only those changes (and no blanket reformat) merges easily. If it's faster for you, I'm happy to re-implement the Windows port-conflict + offline-loading fixes directly with you credited as reporter — just confirm which path you prefer. Either way we want these in v0.3.0.

debpalash:
Thanks so much for this, @caaaaaleb — the Windows fixes here are genuinely valuable and we want them in. 🙏

We've salvaged the two safe, high-value pieces into a clean PR (#156), with you credited as co-author on the commit:

- **The Windows `kill_orphan_on_port` implementation** (`netstat -ano` → `taskkill`) — this was a no-op before, so it's a real fix.
- **The Windows Triton / torch.compile disable** in `backend/main.py` (guarded to `win32`, using `setdefault` so it never overrides a user-set value).

A few things we intentionally **didn't** carry over, and why:

- **The forced `HF_HUB_OFFLINE="1"` + `local_files_only=True`** — this is the main reason we couldn't take the PR as-is. It would break first-run model downloads for every user, and downloading models on first use is core to how the app works. Offline mode needs to stay **opt-in** (only when a user sets `HF_HUB_OFFLINE` themselves). If you've been hitting `httpx` connect timeouts on a restricted network, that's a real problem we'd love to solve properly — likely via an opt-in offline toggle and/or the `uv`/HF mirror fallbacks we're already planning — so any detail on your setup would help.
- **The `model_manager.py` torch.compile / gpu-pool changes** — `main` already has an equivalent (and a bit more capable) version of these: `should_torch_compile()` gates on Triton availability and falls back to eager on Windows, and the lazy `_get_gpu_pool()` is already in place. So your goal there is already covered.
- **The README rewrite and the frontend `.jsx` formatting churn** — out of scope for a Windows-fix change; keeping the diff focused makes it much easier to review and ship.

We're leaving this PR open for you / a maintainer to close at your discretion. Thank you again for the contribution — the port-conflict and Triton fixes are landing because of you.

debpalash:
Your Windows fixes shipped via #156 (merged) — the `netstat`-based `kill_orphan_on_port` and the win32 Triton/torch.compile disable, credited to you. We intentionally did not take the unconditional `HF_HUB_OFFLINE=1`/`local_files_only=True` (it breaks first-run model downloads for everyone — offline should stay opt-in) nor the README rewrite/formatting churn. Thanks @caaaaaleb — closing as superseded by #156.

PR #125, opened 2026-05-26
deploy: expose port on 0.0.0.0, add OMNIVOICE_IDLE_TIMEOUT=0

abhid:
## Summary

Two quality-of-life fixes for self-hosted deployments:

### 1. Expose port on `0.0.0.0` instead of `127.0.0.1`
Allows the container to be reached from the local network / reverse proxy without needing extra Docker networking config. Users who want loopback-only can still override the host binding in their own compose override.

### 2. Add `OMNIVOICE_IDLE_TIMEOUT=0`
Prevents the server from unloading models during periods of inactivity. Essential for always-on deployments where startup latency on the first request is unacceptable.

Both CPU and GPU service profiles updated.

## Summary by CodeRabbit

* **Chores**
  * Updated deployment configuration to enable network access from other machines on the local area network by default, rather than restricting access to the local machine only.
  * Modified idle timeout setting to prevent automatic service shutdown during periods of inactivity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/debpalash/OmniVoice-Studio/pull/125?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -31,7 +31,7 @@ services:
     container_name: omnivoice-studio
     profiles: ["cpu"]
     ports:
-      - "127.0.0.1:3900:3900"
+      - "0.0.0.0:3900:3900"
     volumes:
       - omnivoice-data:/app/omnivoice_data
     environment:
@@ -46,6 +46,7 @@ services:
       # OMNIVOICE_BIND_HOST=0.0.0.0 here only opens the container's own
       # interface. The backend default is 127.0.0.1 (see backend/main.py).
       - OMNIVOICE_BIND_HOST=0.0.0.0
+      - OMNIVOICE_IDLE_TIMEOUT=0
     healthcheck:
       test: ["CMD", "curl", "-sf", "http://localhost:3900/health"]
       interval: 30s
@@ -63,7 +64,7 @@ services:
     container_name: omnivoice-studio-gpu
     profiles: ["gpu"]
     ports:
-      - "127.0.0.1:3900:3900"
+      - "0.0.0.0:3900:3900"
     volumes:
       - omnivoice-data:/app/omnivoice_data
     environment:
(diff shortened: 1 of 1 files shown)


debpalash:
The `OMNIVOICE_IDLE_TIMEOUT=0` addition is fine and useful. The blocker is the host-port change `127.0.0.1:3900:3900` -> `0.0.0.0:3900:3900`: that makes the Docker deploy LAN-reachable by default, and the API has no built-in auth (SECURITY.md: "Do not expose OmniVoice to the internet without authentication"). The file's own comment notes the `127.0.0.1` mapping is what "keeps LAN reachability off by default." Could we keep the loopback default and document `0.0.0.0` as opt-in (a commented line, or a separate `docker-compose.lan.yml` override) — ideally putting a reverse proxy with auth in front? Splitting the two changes lets the idle-timeout part merge immediately.

debpalash:
Thanks for this — the **`OMNIVOICE_IDLE_TIMEOUT=0`** addition is good and we'll take it (sensible for always-on self-host deploys).

The **`0.0.0.0` host mapping** I can't take as the default, though. The backend has loopback-gated admin endpoints (settings, token, etc. via `require_loopback`) and the TTS/dub routes are unauthenticated by design — the local-first model assumes the service is reachable only from the same machine. The compose deliberately maps `127.0.0.1:3900:3900`; the very comment your diff edits says it "keeps LAN reachability off by default." Flipping the default to `0.0.0.0` would silently expose an unauthenticated service to the LAN for everyone who runs the shipped compose.

Could you rework it so LAN exposure is **opt-in**?
- Keep the default mapping `127.0.0.1:3900:3900`.
- Document the LAN path: either a commented-out `# - "0.0.0.0:3900:3900"` line users can uncomment, or a `${OMNIVOICE_PUBLISH_HOST:-127.0.0.1}:3900:3900` mapping driven by an env var (default loopback).
- Keep `OMNIVOICE_IDLE_TIMEOUT=0` as-is.

That lands both QoL wins without changing the secure-by-default posture. Happy to merge once it's opt-in.

debpalash:
Thanks @abhid! We're declining the raw host-mapping flip to `0.0.0.0:3900` — it would silently expose every default `docker compose up` to the LAN (and the edited comment no longer matched the behavior). Instead we're shipping this properly as an explicit, user-driven **network-sharing toggle**: loopback by default, PIN-gated LAN exposure with a QR code, plus Tailscale for private remote access. Design: `docs/superpowers/specs/2026-05-30-network-sharing-design.md`. The `OMNIVOICE_IDLE_TIMEOUT=0` idea for long-running deploys is a supported env var operators can already set. Closing as superseded by that feature.

abhid:
@debpalash Thanks for the heads up. That makes total sense.

Issue #129, opened 2026-05-27
[plan-02] Windows Runtime Integrity — complete venv + safe accel gating


debpalash:
# [plan-02] Windows Runtime Integrity — complete venv + safe accel gating

## Defect
The packaged/bootstrapped Windows venv makes unsafe assumptions about its environment. Transitive runtime deps are missing (e.g. `setuptools`/`pkg_resources` that `ctranslate2`→`whisperx` import), and `torch.compile` is enabled where Triton has no Windows build — both fail only at *inference* time, surfacing as confusing errors (a fake "OOM", or a hard `ModuleNotFoundError` mid-transcription).

## Children
- #116 — `ModuleNotFoundError: No module named 'pkg_resources'` from `ctranslate2.__init__` via `whisperx` during chunk transcription (recurrence of the previously-closed #58)
- #65 — `torch.compile(mode="reduce-overhead")` requires Triton at runtime; Triton is unavailable on Windows → masked as TTS OOM. Suggested gate: `importlib.util.find_spec("triton")` before compile
- (note) the MSI "missing modules / Triton" half of #122 is the same family

## Fix sequence
1. Pin `setuptools` (provides `pkg_resources`) as an explicit runtime dep; verify it resolves in the bootstrapped venv.
2. Add a post-bootstrap venv integrity check that imports the critical chain (`ctranslate2`, `whisperx`, `torch`) and reports a clear actionable error if anything is missing.
3. Gate `torch.compile` on `find_spec("triton")`; fall back to eager mode with an INFO log on platforms without Triton.
4. Add an installer smoke test that imports the ASR/TTS critical path on Windows before the build is published.

## Test matrix
| OS | Accel | Required behavior |
|---|---|---|
| Windows + CUDA, no Triton | torch.compile path | skips compile, eager inference succeeds |
| Windows packaged venv | ASR transcribe | `pkg_resources`/`ctranslate2` import OK, segments produced |
| macOS MPS / Linux CUDA+Triton | both | unchanged |

## Out of scope
HF cache traversal (→ plan-01). The "no error in logs" transparency defect (→ plan-04).

debpalash:
Shipped — fix merged in #138 on the v0.3.0 line. Closing this plan-master as completed; its routed child reports close alongside it.

Issue #116, opened 2026-05-22
[Bug] transcription prodused no segments. no module pkg resources


gitmeatarru:
i try dubbing video

<img width="1849" height="1413" alt="Image" src="https://github.com/user-attachments/assets/a3e03a23-0103-493a-abab-f0a674cb6c66" />
<img width="882" height="196" alt="Image" src="https://github.com/user-attachments/assets/f2de138e-d443-44a0-b0ab-8d20187defc5" />

What you expected to happen.

## Screenshots / Logs

  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load

  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked

  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked

  File "<frozen importlib._bootstrap_external>", line 940, in exec_module

  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed

  File "C:\Users\user\AppData\Local\com.debpalash.omnivoice-studio\project\.venv\Lib\site-packages\whisperx\asr.py", line 5, in <module>

    import ctranslate2

  File "C:\Users\user\AppData\Local\com.debpalash.omnivoice-studio\project\.venv\Lib\site-packages\ctranslate2\__init__.py", line 8, in <module>

    import pkg_resources

ModuleNotFoundError: No module named 'pkg_resources'

2026-05-22 14:52:22,158 WARNING [omnivoice.api] Chunk 164/173 error: No module named 'pkg_resources'

2026-05-22 14:52:22,177 INFO [omnivoice.asr] whisperx loading ASR large-v3 on cuda (float16)

2026-05-22 14:52:22,183 ERROR [omnivoice.api] chunk transcribe failed (backend=whisperx)

Traceback (most recent call last):

  File "C:\Users\user\AppData\Local\com.debpalash.omnivoice-studio\project\backend\api\routers\dub_core.py", line 368, in _transcribe_chunk

    r = _asr_backend.transcribe(tmp.name, word_timestamps=True)

        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users\user\AppData\Local\com.debpalash.omnivoice-studio\project\backend\services\asr_backend.py", line 203, in transcribe

    self._ensure_asr()

[...]

Issue #58, opened 2026-05-15
[Bug] Transcription produced no segments. No module named 'pkg_resources'


sharagakidd:
idk what is it. I click on dub, upload the video, it transcribes for 10 seconds, and then I get an error. I have no idea what to do with this, please help me.

2026-05-15 09:24:04,627 WARNING [omnivoice.api] Chunk 1/1 error: No module named 'pkg_resources'
2026-05-15 09:24:04,627 ERROR [omnivoice.api] transcribe yielded 0 segments (job=79krqeuo): Transcription produced no segments. No module named 'pkg_resources'

<img width="455" height="72" alt="Image" src="https://github.com/user-attachments/assets/1d3cfcce-f5e9-4250-9e61-f87e24ad85ee" />

sharagakidd:
I've already tried downgrading Python and setuptools, but it still didn't work.

WhiskyCoca34:
Same error.

debpalash:
Fix incoming in PR #62 — `setuptools>=75.0` is now pinned in `pyproject.toml` to ensure `pkg_resources` is available on Python 3.12+.

**Workaround until the next release:** run `pip install setuptools` in your Python environment (or `uv pip install setuptools` if using the desktop app's venv).

PR #138, opened 2026-05-29
fix(windows): gate torch.compile on Triton + ASR critical-path smoke (plan-02, closes #65)

### LINKED ISSUE #65: [Bug] torch.compile on Windows causes TTS OOM due to missing Triton
## Describe the bug

On **Windows**, `torch.compile(..., mode="reduce-overhead")` does not throw during model loading but silently generates code paths that depend on Triton at inference time. Since Triton has no official Windows support, TTS generation fails with:

```
TTS engine stopped mid-generation. This usually means it ran out of memory.
Underlying error: Cannot find a working triton installation.
```

## To Reproduce

1. Run OmniVoice-Studio on Windows with CUDA
2. The model loads successfully and logs `torch.compile applied.`
3. Attempt TTS generation
4. Generation fails immediately with OOM / Triton error

## Root Cause

- `torch.compile` with `mode="reduce-overhead"` uses Triton CUDA kernels internally
- Triton is **not available on Windows** (no official support from OpenAI/PTX team)
- The `try/except` in `_load_model_sync()` only catches **compile-time** errors, but the failure happens at **runtime** during inference
- This leads to a confusing error that looks like an OOM, masking the actual Triton dependency issue

## Expected Behavior

On platforms without Triton (Windows), `torch.compile` should be skipped gracefully so TTS inference falls back to eager mode.

## Environment

- OS: Windows
- GPU: Any CUDA-capable
- PyTorch: any recent version
- Triton: not installed / not available

## Suggested Fix

Check for Triton availability before calling `torch.compile`:

```python
import importlib
if importlib.util.find_spec("triton") is not None:
    _model.llm = torch.compile(_model.llm, mode="reduce-overhead")
```

debpalash:
**plan-02** — Closes #65. Addresses #129, #116.

Two Windows failures that only bite at **inference time**:
1. **#65** — `torch.compile(mode="reduce-overhead")` needs **Triton**, which has no Windows wheel. The old `device == "cuda"`-only guard ran it on Windows+CUDA → failed, surfaced as a confusing **"OOM"**.
2. **#116** — `ModuleNotFoundError: pkg_resources` via ctranslate2→whisperx, mid-transcription.

## Fix
- **`engine_env.should_torch_compile(device)`** — applies `torch.compile` only when device==CUDA **and** `importlib.util.find_spec("triton")` is present **and** the existing `perf.torch_compile_disabled` setting is off. Otherwise → **eager mode** with an INFO log. `model_manager.py` uses it in place of the bare CUDA check.
- **smoke-test.sh INST-02** — imports the full ASR path (`torch`, `ctranslate2`, `whisperx`) so a missing transitive dep **fails the build** instead of crashing on the user's machine. Runs in the CI `smoke-matrix` on Windows/macOS/Linux.
- `setuptools>=75.0` (fix-sequence step 1) was already pinned (#58) — confirmed.

## Cross-platform parity
The gate is Triton-availability-driven: Linux/CUDA+Triton is **unchanged** (compile still applied); Windows falls back to eager — same user-visible result (working inference), no divergent default.

## Tests (TDD, fail-before/pass-after — Constitution V)
`tests/test_torch_compile_gate.py` (4): skip on non-CUDA, skip when Triton missing, compile when present+enabled, skip when disabled. Engine/settings/model regression subset: 90 passed, 0 failed.

## Follow-up (out of scope)
A richer in-app post-bootstrap "setup incomplete" banner — deferred; the build-time smoke gate + ASR `is_available()` message cover it for now.

Spec/plan/tasks in `specs/003-windows-runtime-integrity/`.

## Summary by CodeRabbit

* **New Features**
  * Gate performance compilation so it runs only when CUDA, a native accelerator, and a runtime accelerator library are available and not disabled by settings; added a build-time smoke test validating key ML dependencies.

* **Bug Fixes**
  * Reduces risk of compilation-related crashes/out-of-memory on Windows+CUDA setups.

* **Documentation**
  * Added implementation plan, spec, and task documents for runtime integrity.

* **Tests**
  * Added tests covering the compilation eligibility gate.

[Review Change Stack](https://app.coderabbit.ai/change-stack/debpalash/OmniVoice-Studio/pull/138?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/services/engine_env.py
+++ b/backend/services/engine_env.py
@@ -6,8 +6,8 @@ CosyVoice / IndexTTS subprocess backends from Phase 2) should call
 That gives us ONE place to inject:
 
   - HF_TOKEN / YOUR_HF_TOKEN from the 3-source resolver (AUTH-04)
-  - TORCH_COMPILE_DISABLE=1 on Windows when the user enabled the
-    Performance toggle (INST-12, issue #65)
+  - TORCH_COMPILE_DISABLE=1 when the user disabled torch.compile OR Triton is
+    unavailable (INST-12, issue #65) — mirrors should_torch_compile()
 
 The function returns a fresh dict (caller may further mutate before
 passing to `subprocess.Popen(env=...)`).
@@ -17,7 +17,6 @@ from __future__ import annotations
 import importlib.util
 import logging
 import os
-import sys
 from typing import Optional
 
 logger = logging.getLogger("omnivoice.engine_env")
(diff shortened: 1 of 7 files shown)


debpalash:
Addressed both observations in dec971d: `build_engine_env()` now sets `TORCH_COMPILE_DISABLE=1` when the user disabled compile **or** Triton is unavailable (`find_spec`), cross-platform — so subprocess engines get the same gate as the in-process path. Also refreshed the stale module docstring and dropped the now-unused `import sys`. Added 3 tests for the subprocess gate (7/7 pass).

debpalash:
Update on the subprocess-gate observation: I implemented it, but it conflicts with a deliberate, tested contract — `tests/backend/test_perf_settings.py` asserts the subprocess `TORCH_COMPILE_DISABLE` var is **user-toggle-driven only** (Windows + flag-off ⇒ no injection; non-Windows ⇒ never injected). Auto-disabling on Triton-absence would override the user's explicit control and rewrite that contract. So I've reverted that part and kept the in-process `should_torch_compile()` gate, which is the actual #65 fix and is automatic + fully tested. If auto-gating subprocess engines is desired, that's a separate deliberate change to the contract + its tests — happy to do it if the maintainer wants it. Docstring + unused-import notes are resolved.

PR #149, opened 2026-05-29
feat(settings): configurable models directory (#64)

### LINKED ISSUE #64: [Feature] (Windows 1x) Saving downloaded models and weights to desired directory
## Is your feature request related to a problem? 
No.

A clear description of what the problem is. Ex. "I'm always frustrated when..."
By default in Windows OS the app saves all model and weights to %AppData% directory. That's not good practice, as user could have separate and fast SSD dedicated for AI work. So it could be nice feature to have ability to set working dir for downloaded models. That improves app's portability too, as it can be stored together with it's downloaded AI models and weights.

## Describe the solution you'd like

A clear description of what you want to happen.
Add possibility to set download for for models, weights and other AI related things.

## Describe alternatives you've considered

Any alternative solutions or features you've considered.
Don't have'em.

## Additional context

Add any other context, mockups, or screenshots about the feature request here.

debpalash:
Closes #64.

## What

Lets users choose **where model weights download** (the HuggingFace / Torch cache) from **Settings → Models**, instead of being pinned to `~/.cache/huggingface`. Useful when the system drive is small/slow and the user wants weights on a bigger or faster volume.

## How

**Backend**
- `core/user_env.py` — durable per-user env file (`~/.config/omnivoice/env`) helper with `get`/`set`/`unset`. Upsert preserves other keys (e.g. a persisted `HF_TOKEN`) and writes the file `0600`. `main.py` already loads this at startup *before* importing torch/HF, so a written value takes effect on the next launch. Path resolves at call time (honoring an `OMNIVOICE_ENV_FILE` override) so it's robust to module re-import.
- `api/routers/settings.py` — `GET`/`PUT /api/settings/storage/models-dir`:
  - `PUT` validates the directory is writable (`mkdir` + write-probe → `400` if not), persists the choice in the settings store, and writes `OMNIVOICE_CACHE_DIR` to the durable env. Empty path **clears** → reverts to default.
  - Returns `restart_required: true` (an in-use cache can't be safely relocated mid-process). Loopback-gated like the other settings endpoints.

**Frontend**
- `StoragePanel` in the Models tab — view / set / reset the directory; shows effective-vs-configured-vs-default and a restart note.

## Constraints

- **Cross-platform default parity:** default cache path is the HF default on every OS; the feature is the same on mac/Win/Linux. ✅
- **Local-first:** no network, no telemetry. ✅
- **Backward-compatible:** absent setting → existing behavior, no migration. ✅
- **No version bump.** ✅

## Tests

- `tests/test_user_env.py` (5) — upsert/unset/preserve/0600.
- `tests/test_models_dir_setting.py` (4) — persist+durable-env write, unwritable→400, clear→default, GET shape.

All 9 green. (The unrelated `test_supertonic3::test_license_gate` fails locally only because `supertonic` isn't in my local venv; it's a runtime dep present in CI.)

## Summary by CodeRabbit

* **New Features**
  * Storage settings panel and API to view/save the models download directory, showing configured, effective, and default paths; saving/clearing updates persisted per-user env and may require restart.
  * Input validation for saved paths, ensures directories are created and writable, and rejects invalid characters.

* **Tests**
  * Added tests covering storage settings behavior, input validation, default resolution, and per-user env persistence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/debpalash/OmniVoice-Studio/pull/149?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/api/routers/settings.py
+++ b/backend/api/routers/settings.py
@@ -13,6 +13,7 @@ The state endpoint duplicates `/system/hf-token/state` (which lives on
 from __future__ import annotations
 
 import logging
+import os
 from dataclasses import asdict
 from typing import Optional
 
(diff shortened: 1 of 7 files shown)


debpalash:
Addressed the review in `7ef9b11`:

**Store/env divergence (CodeRabbit `:283`/`:219`, Greptile `:259`, both P1/Major)** — the `settings_store` copy of the models dir was only ever read back by this GET endpoint, so it was a redundant cache that could diverge from the durable env file (the value `main.py` actually reads at startup). Removed it entirely. `OMNIVOICE_CACHE_DIR` in the per-user env file is now the **single source of truth**: PUT writes it, GET reads it back — divergence is structurally impossible, so there's nothing to half-persist.

**XDG-aware default (CodeRabbit `:219`)** — `_default_models_dir()` now honors `XDG_CACHE_HOME`, matching `huggingface_hub`'s real default on Linux so GET reports the true default.

**Non-atomic secret exposure (Greptile P2, security)** — `user_env` now writes via an `os.open` opener that creates the file `0600` from the start; no world-readable window before chmod for a file that can hold `HF_TOKEN`.

**`_read_lines` swallowing all `OSError` (CodeRabbit `:25`)** — now only `FileNotFoundError` → empty; other `OSError`s propagate so an upsert can't silently drop existing keys on a transient read failure.

**`makedirs("")` (CodeRabbit `:29`)**, **write-probe cleanup (Greptile P2)**, **`raise ... from` (Ruff B904)**, **a11y input label (CodeRabbit `:97`)**, **OS-neutral unwritable-dir test (CodeRabbit `:42`)** — all fixed.

**CodeQL "user-provided value in path" (3× high, `:271/:273/:275`)** — not fixing these as a vuln, dismissing as false-positive for the threat model. This is a loopback-gated, same-user desktop app where the user is *intentionally* choosing their own local cache directory; the backend runs as that same user, so there's no privilege boundary for traversal to cross — picking an arbitrary directory **is** the feature. Added a control-character/NUL guard up front as defense-in-depth (turns a would-be `ValueError`/500 into a clean 400). All 12 tests green.

PR #295, opened 2026-06-07
feat(setup): unified first-run journey — install gate, studio-console wizard, platform awareness

debpalash:
## What

A ground-up rework of the first-run experience so a fresh download reaches a working install without surprises:

- **Install gate** — nothing downloads until the user approves the install plan
- **Studio-console redesign** — single visual system across the whole journey: masthead (language + region/mirrors), journey rail, hardware-aware Compute + Update channel, storage with disk-space gate
- **Unified library** — models + engines presented as ONE list; Models/Engines wizard acts merged
- **Platform-matrix awareness** — distro + arch detection, ROCm gated to Linux, no Windows console flash
- **Re-entry** — `--setup` flag re-opens the wizard; fixed wizard being skipped after first-run install (probes /setup)
- Welcome step dropped — wizard opens directly on System check

## Why

Core value of the milestone: *a first-run that actually works*. The previous flow installed before asking, hid network options users on restricted networks needed, and couldn't be re-entered after a partial install.

## Notes

- 38 files, +4572/−1275
- `feat/local-diagnostics` is stacked on this branch (separate PR)

## Summary by CodeRabbit

* **New Features**
  * Run self-check and save diagnostic bundle from Settings → About.
  * “Report” flow from error to prefilled bug report; search issues action added.
  * Recent actions breadcrumbs captured for inclusion in reports.
  * Speaker-count (auto) option in dubbing UI.

* **Documentation**
  * Troubleshooting guide updated with self-diagnosis and bundle instructions.

* **Improvements**
  * Unified error toasts include a “Report” button and global uncaught-error handling.
  * Expanded translations across many locales; UX tweaks to setup/wizard and reduced-motion support.


debpalash:
Review findings from CodeRabbit + Greptile addressed in 3c9ebd9a25dac209f0d45b97514aec9bf3b7fdc1:

**Security**
- `setup.rs` `valid_mirror` — plaintext `http://` mirror URLs rejected (HTTPS required; explicit `http://localhost` / `127.0.0.1` / `[::1]` exceptions only). Tests cover loopback-lookalike hosts.
- `setup.rs` + `FirstRunSetup.jsx` — AMD GPUs no longer auto-map to `kind="rocm"`: a cheap ROCm userspace probe (`/opt/rocm` dir or `rocminfo` on PATH) gates it. Unverified AMD machines report `kind="amd"`, so ROCm is offered but never pre-selected and the "matches this machine" badge only shows when verified.

**Functional**
- `--setup` re-entry: `complete_setup` tears down any backend still serving on the port before `retry_bootstrap`, so changed env/mirror/layout settings actually apply instead of silently re-attaching.
- `nvidia-smi` probe runs behind a 3 s timeout — a wedged driver degrades to CPU instead of hanging the setup screen's IPC.
- `is_first_run` is now a pure read; the existing-install migration write moved to `migrate_existing_install_if_needed`, called only from the bootstrap thread.
- `complete_setup` propagates config-write failures to the UI instead of bootstrapping into a stale layout.
- Setup-complete log line records default-vs-custom flags, not raw user paths.
- `scrub.py` + `bugReport.js` redact forward-slash Windows homes (`C:/Users/<name>`, `file:///C:/...`), ordered before the macOS pattern so no `C:~` residue; tests added both sides.
- `bugReport.js` context fetches bounded with a 2.5 s AbortController timeout (partial context instead of a hung report button).
- Crash ack now stores `{size, mtime}` (legacy size-only ack still honored) and `/system/logs/clear` drops the ack — truncation can no longer permanently suppress `crash-last-session`.
- Linux Tauri-log probe honors `XDG_DATA_HOME`.
- Setup-stream phases: verified the `WizardLibrary` reducer is correct against what the backend actually emits (`install_start`/`install_done`/`install_error` lifecycle + per-file `start|progress|done` tqdm events) — the stale part was the `SetupProgressEvent` type, which now documents the full taxonomy; the reducer is annotated on why a file-level `done` must not clear a repo row.
- `SetupWizard` step rail clamps to the highest unlocked step (preflight/models gates) — forward jumps disabled.

**Polish**
- `BootstrapSplash` Waveform memoized (`useMemo([bars])`) like its siblings.
- `detectHints()` returns i18n keys rendered via `t()`; `bootstrap.hint_*` translated in all 21 locales.
- Step-rail `aria-label` localized (`setup.step_aria` / `setup.step_completed`) in all 21 locales.
- `word-break: break-word` → `overflow-wrap: anywhere`; reduced-motion override now also stops the pulsing `frs-hw-pulse` LEDs (install step rail + wizard busy LED).

**Deliberately deferred to a follow-up PR (design-level, out of scope for review fixes)**
- `--setup` re-entry cannot round-trip previously chosen custom dirs/mirrors into the form (`setup.rs`) — needs the saved plan threaded back through `get_setup_state` and the form seeded from it.
- Timed-out deep checks leave worker threads running (`diagnose.py`) — needs a cancellable check executor rather than a spot fix.

PR #311, opened 2026-06-10
fix(tts): let studio engines skip the broadcast mastering chain

tkaufmann:
## Problem

`_run_tts()` in the OpenAI-compatible `/v1/audio/speech` route applies `apply_mastering()` to **every** engine's output. That chain — `HighpassFilter(60Hz)` + `Compressor(threshold=-15, ratio=1.5)` + `Reverb(wet=0.08)` — is tuned for OmniVoice's 24 kHz clone output, where it adds broadcast polish.

Applied to **VoxCPM2**, whose native output is already studio-grade 48 kHz, the same chain is *degradation*, not polish: the compressor introduces an audible level pump and the reverb adds a tail. Cloning the same reference through VoxCPM2 directly (raw) vs. through `/v1/audio/speech` makes the difference clearly audible.

## Fix

Add an opt-out class flag on the backend base:

```python
class TTSBackend(ABC):
    applies_own_mastering: bool = False   # default: unchanged for all engines
```

Set it on the one engine that emits mastered studio output:

```python
class VoxCPM2Backend(TTSBackend):
    applies_own_mastering = True
```

`_run_tts()` skips `apply_mastering()` when the active backend declares it. **Loudness normalisation (`normalize_audio`) still runs for every engine** — that's a benign peak scale, not dynamics processing, so output levels stay consistent.

## Compatibility

Default is `False`, so behaviour is **unchanged for every existing engine** except VoxCPM2. No API/schema change. Other studio-grade engines (e.g. future 48 kHz backends) can opt in by setting the same flag.

## How to reproduce the original issue

1. Install VoxCPM2 (`pip install voxcpm`), select it as the active TTS engine.
2. Create a voice profile, then `POST /v1/audio/speech` with `voice=<profile_id>`, `response_format=wav`.
3. Compare against raw `VoxCPM.generate(...)` output of the same text/reference — the API version has a level pump + reverb tail that the raw output doesn't.

## Summary by CodeRabbit

* **New Features**
  * Studio-grade TTS engines can opt out of shared mastering to preserve native quality; streaming still applies loudness normalization for consistent volume.
* **Bug Fixes**
  * Avoids redundant mastering that could degrade audio or waste processing, improving output fidelity and efficiency.

### DIFF
--- a/backend/api/routers/batch.py
+++ b/backend/api/routers/batch.py
@@ -298,6 +298,9 @@ async def _run_batch_pipeline(job_id: str, job: dict):
                         denoise=True, postprocess_output=True,
                     )
                     audio_out = audios[0]
+                    # TODO(#312): this route runs the OmniVoice model directly (not the active
+                    # backend), so VoxCPM2 never reaches it. When these routes become
+                    # engine-aware, guard with `if not getattr(backend, "applies_own_mastering", False)`.
                     mastered = apply_mastering(
                         audio_out,
                         sample_rate=sr,
--- a/backend/api/routers/dub_generate.py
+++ b/backend/api/routers/dub_generate.py
@@ -265,6 +265,9 @@ async def dub_generate(job_id: str, req: DubRequest):
                     if seg_effect_preset == "raw":
                         return audio_out
 
+                    # TODO(#312): this route runs the OmniVoice model directly (not the active
+                    # backend), so VoxCPM2 never reaches it. When these routes become
+                    # engine-aware, guard with `if not getattr(backend, "applies_own_mastering", False)`.
                     mastered_audio = apply_mastering(audio_out, sample_rate=sr)
                     effect_chain = get_effect_chain(seg_effect_preset)
                     if effect_chain:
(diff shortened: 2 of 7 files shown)


tkaufmann:
Thanks @greptile-apps — good catch on the call-site coverage. I verified each one against the source rather than guarding blindly, and the picture splits in two:

**Routes that run the *active* backend (can be VoxCPM2) → need the guard:**
- `openai_compat.py` `_run_tts` — guarded (original commit)
- `tts_stream.py` — resolves the engine via `get_active_tts_backend(model=model)`, so it genuinely hit the regression. **Now guarded** in the follow-up commit.

**Routes that call the OmniVoice model directly (never the active backend) → VoxCPM2 cannot reach them:**
- `generation.py` (`_run_inference(model, …)`), `batch.py`, `batched_tts.py`, `dub_generate.py` all generate via `get_model()` / `_model.generate(...)`, which loads `OmniVoice.from_pretrained(...)`. The active TTS backend is never consulted there — so when VoxCPM2 is selected, its output never flows through those `apply_mastering()` calls. They master OmniVoice output, which is the intended behaviour.

That OmniVoice-only routing is itself a bit surprising (I filed #312 about `/generate` ignoring the engine selection). If/when those routes become engine-aware, the same `if not getattr(backend, "applies_own_mastering", False)` guard should move with them.

I kept the diff to the two routes that actually run the active engine to avoid adding always-False guards on the OmniVoice-only paths — but if you'd prefer a defensive guard on all six sites regardless, I'm happy to add it.

tkaufmann:
Good call — added the `TODO(#312)` pointers at the six OmniVoice-only `apply_mastering()` sites (generation.py, batch.py, batched_tts.py, dub_generate.py ×3) instead of dead guards, so the guard lands exactly when those routes become engine-aware. Pushed.

Issue #276, opened 2026-06-03
[Feature] Pause in Transcript


graywolf336:
## Is your feature request related to a problem?

I would like to add a way to tell the system, in the transcript, to pause. Or configure how long the voice pauses when it hits the end of a sentence. Or even a new line.

## Describe the solution you'd like

I think a `[pause]` or something like that would work or even a "short breathe in take" type thing.

## Describe alternatives you've considered

I considered the `[sigh]` but it didn't sound/feel right.

## Additional context

n/a

debpalash:
Thanks — this is a nice, well-scoped ask. Good news: it's implementable cleanly **without touching the model**. The synthesis path already splits long text into chunks and stitches the audio back with a configurable silence gap (`cross_fade_chunks(..., silence_duration=...)`), so an inline pause marker fits naturally there:

**Plan**
1. Parser: `[pause]` → a default gap; `[pause 500ms]` / `[pause 1s]` → an explicit duration. Split the text on the marker.
2. Synthesize each text span as today, then insert a `torch.zeros` silence of the requested length between spans before concatenation (reusing the existing silence/concat helper). Pure, deterministic, unit-testable — no model changes, and it's fully opt-in so it can't affect anyone who doesn't use the marker.

**One design question before I build it** — which syntax do you (and others reading this) prefer?
- `[pause]` + optional duration: `[pause]`, `[pause 500ms]`, `[pause 1s]`  ← my lean (matches your suggestion, simple)
- SSML-style: `<break time="500ms"/>` (more verbose, but a known standard)
- A global "pause length at sentence end / newline" setting (your alternative idea) — note this is a separate, broader knob; the inline marker gives finer control. I could do both.

I'll go with the `[pause Nms]` inline marker unless there's a strong preference otherwise. (Note on `[sigh]`/`[laugh]`: those are *acted* nonverbal sounds the model voices; a pause is just silence, so it's handled differently — at the audio-stitch layer, as above.)

It'll land in an upcoming build once the syntax is settled.

graywolf336:
`[pause Nms]` is a good option as it fits the other options in the system already.

debpalash:
[...]

PR #324, opened 2026-06-11
fix(tts): /generate honors the selected TTS engine (#312)

### LINKED ISSUE #312: /generate ignores the selected TTS engine (always uses OmniVoice) — intended?
## Observation

`POST /generate` appears to always use the **OmniVoice** model, regardless of which TTS engine is active via `POST /engines/select`. The engine selection only seems to affect `/v1/audio/speech` (and `tts_stream`), which resolve the active backend — whereas `generate_speech()` in `backend/api/routers/generation.py` calls `_model = await get_model()` (which loads `OmniVoice.from_pretrained(...)`) and never consults the active TTS backend.

## Reproduce

```bash
# Select VoxCPM2, generate with a fixed seed
curl -s -X POST :PORT/engines/select -H 'Content-Type: application/json' \
     -d '{"family":"tts","backend_id":"voxcpm2"}'
curl -s -F 'text=Kontrolltext.' -F 'language=de' -F 'seed=42' :PORT/generate -o a.wav

# Select OmniVoice, same seed
curl -s -X POST :PORT/engines/select -H 'Content-Type: application/json' \
     -d '{"family":"tts","backend_id":"omnivoice"}'
curl -s -F 'text=Kontrolltext.' -F 'language=de' -F 'seed=42' :PORT/generate -o b.wav

# a.wav and b.wav are byte-identical → /generate ignored the VoxCPM2 selection
```

In my test the two files were identical (same md5), confirming `/generate` ran OmniVoice both times.

## Question

Is this intended — i.e. is `/generate` deliberately the OmniVoice-native endpoint, and pluggable engines are meant to be reached only through `/v1/audio/speech`?

- If **yes**: it would help to document this (the engine picker silently has no effect on `/generate`), or surface a hint in the UI so users don't think their engine choice applies there.
- If **no**: `/generate` should route through the active backend like the OpenAI-compatible route does.

Happy to help with a PR once the intended behaviour is clear. (Context: I'm running VoxCPM2 for German voice cloning and hit this while comparing engines.)

debpalash:
## Summary

`POST /generate` always ran the OmniVoice model directly via `get_model()`, silently ignoring the engine selected in Settings (and answering issue #312's question: no, it wasn't intended). It now:

- resolves the active backend per request: explicit `engine` form field > `OMNIVOICE_TTS_BACKEND` env > Settings selection > OmniVoice default — the same pattern `/ws/tts` (`engine`) and `/v1/audio/speech` (`model`) already use
- keeps the OmniVoice default path **byte-identical** (native path with the full advanced parameter surface: `t_shift`, layer/position/class controls) — existing API consumers see zero change
- reuses the per-process engine instance cache shared with the health-check route, so weights load once
- keeps inline `[pause Nms]` markers (#276) working on every engine (silence stitching is model-free)
- honors `applies_own_mastering` so studio engines skip the broadcast mastering chain (closing the TODO left by 5d602c8), while loudness normalization still applies
- unknown/unavailable engines return an actionable 400 listing valid ids

Fixes #312

## Tests

`tests/test_generate_engine.py` (new, 6 tests, engine layer stubbed as in test_api.py): Settings-selected engine generates; per-request `engine` overrides; unknown → 400; unavailable → 400; default path unchanged; own-mastering engines skip the chain. Full file passes locally; CI runs the complete suite.

## Summary by CodeRabbit

* **New Features**
  * Per-request TTS engine override and engine-aware inference path.
  * Language normalization (e.g., "auto" → undetermined) and consistent sample-rate handling.
  * Shared audio effect chain with optional raw passthrough and conditional mastering when engines apply their own mastering.

* **Bug Fixes**
  * Clearer errors for unknown or unavailable engines and improved out-of-memory error handling.

* **Tests**
  * Added tests covering engine selection, overrides, availability errors, and mastering behavior.

### DIFF
--- a/backend/api/routers/generation.py
+++ b/backend/api/routers/generation.py
@@ -59,13 +59,65 @@ def _render_with_pauses(gen_span, segments, sample_rate):
             parts.append(torch.zeros(*shape, dtype=ref.dtype, device=ref.device))
     return torch.cat(parts, dim=-1)
 
+
+def _apply_effect_chain(audio_out, sample_rate, effect_preset, *, skip_mastering=False):
+    """Shared post-DSP for /generate: preset validation → mastering →
+    effect chain → loudness normalization.
+
+    ``skip_mastering`` honors a backend's ``applies_own_mastering`` flag
+    (issue #312): studio engines (e.g. VoxCPM2's native 48 kHz output)
+    opt out of the broadcast Compressor + Reverb chain that's tuned for
+    OmniVoice's 24 kHz clone output. Loudness normalization still runs —
+    it's a benign peak scale. Mirrors ``_run_tts`` in openai_compat.py.
+    """
+    from services.audio_dsp import (
+        EFFECT_PRESETS, apply_mastering, normalize_audio,
+        apply_effects_chain, get_effect_chain,
+    )
+
+    preset = effect_preset or "broadcast"
+    if preset not in EFFECT_PRESETS:
+        raise ValueError(
+            f"Unknown effect preset: {preset!r}. "
+            f"Valid: {list(EFFECT_PRESETS.keys())}"
+        )
+
+    if preset == "raw":
+        # Raw: skip all DSP — return raw model output
+        return audio_out
+
+    if not skip_mastering:
+        audio_out = apply_mastering(audio_out, sample_rate=sample_rate)
+    chain = get_effect_chain(preset)
+    if chain:
+        audio_out = apply_effects_chain(
+            audio_out, sample_rate=sample_rate, chain=chain,
+        )
+    return normalize_audio(audio_out, target_dBFS=-2.0)
+
+
+def _oom_friendly_reraise(e):
+    """Best-effort cache flush + the user-facing OOM hint shared by both
+    inference paths."""
+    import gc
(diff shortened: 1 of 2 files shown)


debpalash:
Thanks — the approach looks right (resolution order matches /ws/tts and /v1/audio/speech, and keeping the OmniVoice default path byte-identical is the right call), but CI is red and the failures are in this PR's own new tests:

- All 4 engine-stub tests in `tests/test_generate_engine.py` fail with `400 — Unknown TTS engine: 'fake-engine'`: the stubbed engine isn't registered when the full suite runs (works in isolation per the PR body, fails in CI). Looks like the engine registry the stub patches isn't the one the route reads at request time — or another test's setup/teardown resets the registry. Worth registering the stub through the same registry the route consults, via a fixture that re-applies per test.
- One teardown error: `RuntimeError: <Queue> is bound to a different event loop` in `test_generate_respects_applies_own_mastering` — suggests the test creates async state across event loops; the patterns in test_api.py (TestClient-per-test) avoid this.

CI run: https://github.com/debpalash/OmniVoice-Studio/actions/runs/27321203190

Happy to re-review once the suite is green.

PR #332, opened 2026-06-11
fix(updater): preview channel never offers updates once stable catches up

### LINKED ISSUE #326: Preview update channel never offers updates once stable catches up (semver pre-release ordering)
## What the user reported

On the **preview** update channel (Settings), checking for updates on v0.3.5 always says "you already have the latest version", even though newer preview builds of `main` exist. Updating used to work for them (e.g. 0.3.4 → 0.3.5 on stable).

> "I have a quick question about the "preview" channel in the program settings. In theory, it should download and apply the changes listed in the main GitHub repository? Because if so, it's not working for me with version 0.3.5, although it has worked before when I've tried to update, for example, from 0.3.4 to 0.3.5 in the "stable" channel." — "Basically, it says I already have the latest version"

## Environment

- App version: 0.3.5 (stable build), update channel set to **preview**

## Root cause (verified)

- `frontend/src-tauri/src/updater_channel.rs` points the preview channel at the rolling `preview` release manifest: `https://github.com/debpalash/OmniVoice-Studio/releases/download/preview/latest.json`.
- That manifest currently advertises `"version": "0.3.5-41"` (pub_date 2026-06-07).
- `tauri-plugin-updater`'s default comparator is a plain semver `update > current`. In semver, `0.3.5-41` is a **pre-release of 0.3.5** and therefore sorts *below* `0.3.5`.
- So any user on stable `0.3.5` who switches to the preview channel is told they're up to date — even though the preview build is 4 days newer. It worked before stable caught up because `0.3.5-x > 0.3.4`.

## Fix direction

For the **preview** channel only, install a custom `version_comparator` on the updater builder (the documented Tauri pattern for rolling-release channels): treat an update as available whenever the remote version differs from the current one. Stable channel keeps the default semver comparison, so default behavior is unchanged on all platforms.

---
Reported by **nanai97** on Discord: https://discord.com/channels/1497926884707799144/1497926886020354102/1514363596841287862

debpalash:
## Problem

A user on stable `0.3.5` who switches the update channel to **preview** is always told they're up to date, even though the rolling `preview` release is days newer (reported on Discord, #326).

Preview builds are versioned as pre-releases of the *current* stable base (the manifest currently advertises `0.3.5-41`), and `tauri-plugin-updater`'s default availability check is a strict semver `remote > current`. In semver, `0.3.5-41 < 0.3.5` — so the moment stable catches up to the preview base version, the preview channel goes permanently silent. It worked before only because `0.3.5-x > 0.3.4`.

## Fix

Install a channel-aware `version_comparator` on the updater builder in `check_update` and `install_update` (`frontend/src-tauri/src/updater_channel.rs`):

- **preview** — rolling channel: any manifest version *different* from the running one is an update.
- **stable** (and anything unknown) — keeps the default strict `remote > current`, so default behavior is byte-for-byte unchanged.

The predicate is factored into `update_available()` with unit tests covering the regression case (stable 0.3.5 → preview 0.3.5-41), rolling upgrades (0.3.5-40 → -41), same-build no-op, and stable's unchanged ordering. Adds `semver = "1"` as a direct dep (already in the tree via tauri itself).

Pure Rust logic — identical on macOS/Windows/Linux; no UI strings, no data-format changes, no version bumps.

## Tests

- `cargo test --lib updater_channel`: **3 passed** (new)
- `cargo check`: clean (one pre-existing warning in setup.rs)
- Backend `uv run pytest tests/ -q`: **729 passed, 21 skipped, 11 xfailed, 3 xpassed**
- `bun run test:frontend`: **36 passed**

Fixes #326

### DIFF
--- a/frontend/src-tauri/Cargo.lock
+++ b/frontend/src-tauri/Cargo.lock
@@ -2957,6 +2957,7 @@ dependencies = [
  "libc",
  "log",
  "reqwest",
+ "semver",
  "serde",
  "serde_json",
  "sysinfo",
--- a/frontend/src-tauri/Cargo.toml
+++ b/frontend/src-tauri/Cargo.toml
@@ -24,6 +24,9 @@ tauri-plugin-log = "2"
 tauri-plugin-dialog = "2"
 tauri-plugin-window-state = "2.0.0"
 tauri-plugin-updater = "2"
+# Channel-aware update comparator (preview is a rolling channel; see
+# updater_channel.rs). Same semver crate tauri-plugin-updater uses internally.
+semver = "1"
 tauri-plugin-process = "2"
 tauri-plugin-opener = "2"
 tauri-plugin-global-shortcut = "2"
--- a/frontend/src-tauri/src/updater_channel.rs
+++ b/frontend/src-tauri/src/updater_channel.rs
@@ -30,6 +30,22 @@ fn channel_endpoints(channel: &str) -> Vec<tauri::Url> {
     raw.iter().filter_map(|u| u.parse().ok()).collect()
 }
 
+/// Update-availability predicate for a channel.
+///
+/// Stable keeps the plugin's default strict-semver `remote > current`. Preview
+/// builds are versioned as pre-releases of the *current* stable base (e.g.
+/// `0.3.5-41`), and semver orders a pre-release *below* its release — so once
+/// stable catches up, the default comparator tells preview users they are
+/// already up to date forever (#326). Preview is a rolling channel: any
+/// manifest version that differs from the running one is an update.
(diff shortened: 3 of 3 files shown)


debpalash:
Superseded by #335, which fixes the same semver pre-release ordering bug and additionally fixes the endpoint fallback: tauri-plugin-updater stops at the first manifest that parses, so with this PR a reachable preview manifest would still hide a newer stable release from preview-channel users. #335's cross-channel comparator is also downgrade-safe (strict ordering instead of "any different version"). Both PRs were generated against the same issue in parallel — closing this one. (No commits are lost; the branch stays.)

PR #639, opened 2026-06-23
Fix dub generation memory spikes

trungthanh1288:
## Changes

-

## Type

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] ♻️ Refactor
- [ ] 📝 Documentation
- [ ] 🧪 Tests
- [ ] 🔧 CI / Build
- [ ] 🚀 Release prep

## Testing

-

## Checklist

- [ ] I've updated relevant documentation (if applicable)
- [ ] No local machine paths, logs, or personal env details in this PR
- [ ] Version files are in sync (if version bump): `pyproject.toml`, `package.json`, `tauri.conf.json`, `Cargo.toml`
- [ ] If this PR changes runtime behavior, the regression fixture at `tests/fixtures/omnivoice_data/` still loads green on the `smoke-matrix` CI job (macOS + Windows + Linux)

## Release cadence

OmniVoice ships **continuous-to-main** — no release candidates, no soak windows.
Every merged PR is immediately part of the rolling preview (`main`, Docker
`:latest`, the desktop Preview channel). Versioned releases are tagged from
`main` when it's ready; `main` then bumps to the next patch automatically.
Users who want stability pin a release tag / Docker `:stable` / the desktop
Stable channel.

# Fix Memory Spikes During Dub Generation

## Overview
This PR refactors the dub generation pipeline in `backend/api/routers/dub_generate.py` to prevent RAM/VRAM memory spikes by switching to a disk-first, streaming assembly model. It also corrects a latent indexing bug affecting negative `place_at` values and ensures watermarking occurs on the assembled final track (not per-segment WAVs).

### Memory Architecture Shift (disk-first + memmap)
- **Before:** accumulated full-mix audio in large in-memory tensors.
- **After:** write each segment (or silence) to disk during the TTS “hot loop”, then assemble the final output using a **temporary float32 `np.memmap`** buffer sized to the computed `total_samples`.

### New/Updated Internal Helpers (within `_stream`)
- `_release_audio_tensors()` for best-effort GPU/CPU memory cleanup (del + `gc.collect`, GPU cache clears).
- `_store_mix_wav()` to atomically persist a segment (or silence) to disk and return a lightweight manifest tuple `(start, end, path, sr)`.
- `_entry_num_samples()` to determine segment sample counts from disk-backed WAVs (prefers `torchaudio.info`, falls back to a load-based approach).
- `_load_entry_wav()` to load and resample segment WAVs on demand during assembly.
- `_write_memmap_wav_atomic()` to finalize output via chunked float32→PCM16 writing; **watermark embedding is performed inside this function on the assembled mix** (chunked to control memory).

### Hot Loop / Segment Handling Changes
- Silence/empty segments are **stored immediately** to disk and tensors are released.
- Cached segments are routed through `_store_mix_wav` using the manifest/path model (not retained tensors).
- `_pending_seg_writes` now tracks only lightweight metadata (no audio tensors).
- For newly generated segments:
  - Attempt atomic segment WAV persistence via `atomic_save_wav(...)`.
  - On failure, fall back to storing a mix copy; if errors persist, store a silent segment.

### Final Assembly Rewrite
- Removes `full_audio = torch.zeros(...)` accumulation.
- Allocates a temp float32 memmap for the full track.
- For each segment:
  - Load from disk, apply per-segment gain.
  - Apply the existing strategy-specific processing (preserving fit strategies: `stretch_video`, `smart_fit`, `concise`, `strict_slot`).
  - Handle trimming/stretching/overflow, apply short fades, and overlap-add into the memmap.
- Flush memmap, then write final WAV via `_write_memmap_wav_atomic` (including assembled-track watermark embedding and PCM16 conversion), followed by cleanup.

### Timing / Strategy Calculations
- `stretch_video` and `smart_fit` compute per-segment durations/wavelength from the **disk-backed sample counts** (`_entry_num_samples`) instead of tensor lengths.

### Latent Bug Fix
- Fixes incorrect buffer indexing when `place_at` is negative by trimming the head appropriately.

## Watermarking Behavior Change (Blocking Concern Addressed)
- Watermarking is applied at the **final assembled-track write stage** (inside `_write_memmap_wav_atomic`) so watermark detection is preserved after time-stretching, fading, gain scaling, and overlap-add.
- Per-segment WAVs are no longer expected to contain the watermark marker.

## Processing Flow
```mermaid
graph TD
  A["Dub Generation Start"] --> B["Segment Hot Loop"]
  B --> C{"Segment Type"}
  C -->|Silence/Empty| D["Store silence WAV to disk (_store_mix_wav)"]
  C -->|Cached| E["Store/reuse cached WAV path manifest (_store_mix_wav)"]
  C -->|TTS Generated| F["Generate segment -> atomic_save_wav to disk (fallback on failure)"]
  D --> G["Release tensors (_release_audio_tensors)"]
  E --> G
  F --> G
  G --> H["Record lightweight segment metadata (_pending_seg_writes)"]
  H --> I["Allocate float32 memmap for total_samples"]
  I --> J["For each segment: load WAV from disk (_load_entry_wav)"]
  J --> K["Apply gain + strategy processing + fades + overlap-add into memmap"]
  K --> L{"All segments processed?"}
  L -->|No| J
  L -->|Yes| M["Finalize: chunked watermark embed on assembled mix + PCM16 write (_write_memmap_wav_atomic)"]
  M --> N["Flush memmap + cleanup temp files"]
  N --> O["Return final dubbed_*.wav"]
```

## Tests
- Added `test_final_dub_track_is_watermarked_after_assembly` in `tests/test_smart_fit_generate.py` to assert:
  - the final `dubbed_es.wav` is detected as watermarked,
  - `embed_watermark` is called exactly once for the full assembled duration,
  - per-segment WAVs do not contain the watermark marker.

## Impact
- **Files changed:** 2
- **Lines changed:** +353 / -178
- **Code review effort:** High

### DIFF
--- a/backend/api/routers/dub_generate.py
+++ b/backend/api/routers/dub_generate.py
@@ -107,6 +107,99 @@ async def dub_generate(job_id: str, req: DubRequest):
         all_segment_wavs = []
         sync_scores = []
 
+        def _release_audio_tensors(*objs) -> None:
+            """Best-effort RAM/VRAM cleanup after a segment is safely on disk."""
+            for obj in objs:
+                try:
+                    del obj
+                except Exception:
+                    pass
+            try:
+                import gc
+                gc.collect()
+            except Exception:
+                pass
+            try:
+                if torch.cuda.is_available():
+                    torch.cuda.empty_cache()
+                elif hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
+                    torch.mps.empty_cache()
+            except Exception:
+                pass
+
+        def _store_mix_wav(start: float, end: float, wav: torch.Tensor, sr: int, seg_key: str):
+            """Write one segment to disk and keep only its path in the mix manifest."""
+            path = dub_seg_path(job_id, seg_key)
+            os.makedirs(os.path.dirname(path), exist_ok=True)
+            atomic_save_wav(path, wav.detach().cpu(), sr)
+            _release_audio_tensors(wav)
+            return (start, end, path, sr)
+
+        def _entry_num_samples(entry) -> int:
+            try:
+                info = torchaudio.info(entry[2])
+                return int(info.num_frames)
+            except Exception:
+                wav, _sr = torchaudio.load(entry[2])
+                n = int(wav.shape[-1])
+                _release_audio_tensors(wav)
+                return n
+
+        def _load_entry_wav(entry, target_sr: int) -> torch.Tensor:
+            wav, loaded_sr = torchaudio.load(entry[2])
+            if loaded_sr != target_sr:
+                import torchaudio.functional as AF
+                wav = AF.resample(wav, loaded_sr, target_sr)
+            return wav
+
+        def _write_memmap_wav_atomic(target_path: str, samples, sample_rate: int) -> None:
+            """Watermark and write a mono float32 memmap without loading it all."""
+            import tempfile
+            import wave
+            import numpy as np
+
+            target_dir = os.path.dirname(target_path) or "."
+            target_base = os.path.basename(target_path)
+            fd, tmp_path = tempfile.mkstemp(
+                prefix=f".{target_base}.",
+                suffix=".wav",
+                dir=target_dir,
+            )
+            os.close(fd)
+            chunk_samples = max(sample_rate * 30, 1)
+            try:
+                with wave.open(tmp_path, "wb") as wf:
+                    wf.setnchannels(1)
+                    wf.setsampwidth(2)
+                    wf.setframerate(sample_rate)
+                    total_len = int(samples.shape[0])
+                    for off in range(0, total_len, chunk_samples):
+                        chunk = np.array(samples[off: off + chunk_samples], dtype=np.float32, copy=True)
+                        if chunk.size == 0:
+                            continue
+                        np.nan_to_num(chunk, copy=False, nan=0.0, posinf=1.0, neginf=-1.0)
+                        chunk_tensor = torch.from_numpy(chunk).unsqueeze(0)
+                        chunk_tensor = embed_watermark(chunk_tensor, sample_rate)
+                        chunk = (
+                            chunk_tensor
+                            .detach()
+                            .cpu()
+                            .to(torch.float32)
+                            .squeeze(0)
+                            .numpy()
+                        )
+                        np.nan_to_num(chunk, copy=False, nan=0.0, posinf=1.0, neginf=-1.0)
+                        chunk = np.clip(chunk, -1.0, 1.0)
+                        pcm = (chunk * 32767.0).astype("<i2", copy=False)
+                        wf.writeframes(pcm.tobytes())
+                os.replace(tmp_path, target_path)
+            except BaseException:
+                try:
+                    os.unlink(tmp_path)
+                except OSError:
+                    pass
+                raise
+
         # Phase 4.1 — partial regen. If `regen_only` is set, we only run TTS
         # on segments whose id is in that set; the others reuse their existing
         # `seg_i.wav` on disk and slot into the final mix unchanged.
@@ -126,9 +219,9 @@ async def dub_generate(job_id: str, req: DubRequest):
         # reorder; index-keyed readers (preview/export) resolve via this manifest.
         job["seg_order"] = [seg_ids[k] if k < len(seg_ids) else f"seg_{k}" for k in range(len(req.segments))]
 
-        # Deferred disk writes: collect (index, tensor, sr, seg_id, fingerprint,
-        # num_step) tuples during the hot loop and batch-flush after all TTS
-        # completes. Eliminates ~200ms/seg of synchronous I/O from the GPU path.
+        # Per-segment metadata to persist after the hot loop. Audio itself is
+        # written immediately and only file paths are kept, so long videos don't
+        # retain every generated tensor in RAM until final assembly.
         _pending_seg_writes: list[tuple] = []
 
         # Phase 4.1 bench instrumentation: measure where incremental time goes.
@@ -151,7 +244,12 @@ async def dub_generate(job_id: str, req: DubRequest):
             if seg_duration <= 0.05 or not seg.text.strip():
                 sr = _model.sampling_rate
                 silence = torch.zeros(1, int(seg_duration * sr))
-                all_segment_wavs.append((seg.start, seg.end, silence, sr))
+                all_segment_wavs.append(_store_mix_wav(seg.start, seg.end, silence, sr, f"mix_{seg_id}"))
+                try:
+                    del silence
+                except Exception:
+                    pass
+                _release_audio_tensors()
                 sync_scores.append(1.0)
                 continue
 
@@ -182,7 +280,12 @@ async def dub_generate(job_id: str, req: DubRequest):
                                 cached_wav = torch.nn.functional.pad(cached_wav, (0, target_samples - current_samples))
                             elif current_samples > target_samples:
                                 cached_wav = cached_wav[..., :target_samples]
-                        all_segment_wavs.append((seg.start, seg.end, cached_wav, _model.sampling_rate))
+                        all_segment_wavs.append(_store_mix_wav(seg.start, seg.end, cached_wav, _model.sampling_rate, f"mix_{seg_id}"))
+                        try:
+                            del cached_wav
+                        except Exception:
+                            pass
+                        _release_audio_tensors()
                         sync_scores.append(getattr(seg, 'sync_ratio', None) or 1.0)
                         _t_cache += time.perf_counter() - _t_cache_0
                         continue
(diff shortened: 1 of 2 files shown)


debpalash:
Thanks for this — it's a genuinely well-built refactor. I reviewed the full diff and ran the suite locally (checked out the branch): **1857 passed**, and the dub-specific tests (191) are green. The disk-first/memmap streaming approach is sound, the four fit strategies (stretch_video / smart_fit / concise / strict_slot) are preserved verbatim, the atomic writes + `nan_to_num`+clip are correct, and you even fixed a latent bug (negative `place_at` now trims the head instead of mis-indexing the buffer). The memory goal is real. 🙌

**One blocker before merge — the AI watermark regresses.**

This drops the final whole-track watermark pass:
```python
- full_audio = embed_watermark(full_audio, sr)   # removed
+ _write_memmap_wav_atomic(track_path, mix_audio[:mix_samples], sr)   # no watermark
```
The final dubbed track is now assembled from per-segment WAVs that were watermarked **before** they're time-stretched (atempo/interp), faded, gain-scaled and overlap-added. Two problems:
1. **AudioSeal isn't robust to time-stretching** — watermarking *before* the stretch/mix (instead of once over the final mix, as `main` does) likely degrades or destroys detectability on `smart_fit`/`strict_slot`/`stretch_video` outputs.
2. **Inter-segment gaps are silent zeros** that never pass through any watermark, so large stretches of the track carry no mark.

This is silent — no test detect-watermarks the final mix, which is why the suite stays green.

**Suggested fix:** watermark the assembled mix, not the segments. Easiest is to keep the per-segment writes clean and add the watermark in the final write — e.g. inside `_write_memmap_wav_atomic` run `embed_watermark` over the float32 mix (chunked to preserve the memory win, or in one pass if size allows) before the PCM16 encode. A regression test that asserts `detect_watermark` succeeds on `dubbed_<lang>.wav` would lock it in.

**Minor (non-blocking):** on Windows, `np.memmap` can hold the temp `.f32` handle open past `del`, so `os.unlink(mix_path)` may raise `PermissionError` — it's caught, so it just leaks a temp file. A `mix_audio._mmap.close()` (or `del` + `gc.collect()` which you already do) before unlink tightens that up.

Really nice work — fix the watermark ordering + add the detection test and I'll merge. Happy to help if you want to pair on the watermark placement.

debpalash (commented):
Thanks for this — the core idea is right and the memory problem is real. I reviewed the diff and ran the suite locally (`tests/test_smart_fit_generate.py` → **6 passed**), so the assembly refactor is behavior-preserving for the existing strategies. Notes below, grouped by severity.

### 👍 What's good
- **Disk-backed `np.memmap` mix** replaces `full_audio = torch.zeros(1, total_samples)` — that full-track RAM allocation was the spike, so this is the right fix.
- **Path-based `all_segment_wavs`** (write each seg, keep only the path) avoids retaining every tensor until assembly.
- Genuine **bug fix**: the new `s < 0` handling (trim the head when `place_at` is negative) — the old `full_audio[:, s:e]` with negative `s` indexed from the end. Nice catch.
- Atomic temp-file write + `finally` cleanup of the memmap and `.f32` scratch.

### ❌ Blocking: the watermark test is insufficient and partly incorrect
`test_final_dub_track_is_watermarked_after_assembly` **mocks `embed_watermark`** with a trivial head-marker, so it validates nothing about how real **AudioSeal** behaves when applied per-chunk. More importantly it only exercises a **single sub-30s chunk** (`duration=3.0`), and asserts:
```python
assert watermark_calls == [int(job["duration"] * SR)]
```
But the new `_write_memmap_wav_atomic` calls `embed_watermark` **once per 30s chunk** (`chunk_samples = sr*30`). For any track **>30s** — the exact long-video case this PR targets — there are multiple calls, so this assertion would **fail**. The test passes only because it never crosses a chunk boundary. Please add a **>30s, multi-chunk** case and assert correctness there (and ideally a real-AudioSeal embed→`detect_watermark` round-trip, gated on `audioseal` availability).

### ⚠️ Chunked AudioSeal watermarking — needs validation
The old code watermarked the whole track in one `embed_watermark(full_audio, sr)` call; this PR watermarks **independent 30s windows**. The fixed `OMNI_MESSAGE` means each chunk carries the same message (good — no per-chunk message drift), but AudioSeal is a neural generator and independent windows can introduce **boundary effects at each 30s seam**. Please confirm on a **real >30s dub** that (a) `detect_watermark` still passes across the whole track and (b) there are no audible 30s artifacts. If memory is the only driver, note AudioSeal runs at 16 kHz internally, so a whole-track embed is far cheaper than the 44.1 kHz mix — chunking may not even be necessary for the watermark step.

### ⚠️ Performance: this reverses a deliberate optimization
The old deferred-write phase explicitly kept I/O off the GPU-hot path (the removed `~200ms × N_segments` comment). Now each segment is written immediately **and** reloaded at assembly, plus `gc.collect()` + `torch.cuda.empty_cache()` run **per segment**. `empty_cache()` forces a CUDA sync; on a 50-segment dub that's 50 syncs and could measurably slow generation. Suggest throttling the cache clears (every N segments, or only when free VRAM is low) and/or sharing a quick before/after timing on a long video.

### ℹ️ Minor
- Per-segment `seg_i.wav` is **no longer watermarked** (only the final track is). Probably fine since the final export is what carries provenance, but please confirm it's intentional — preview/partial-regen consumers now read unwatermarked seg files.
- `_release_audio_tensors(*objs)` only `del`s its local parameter binding, which doesn't free the caller's reference — the real release is the caller's own `del` + `gc.collect()`. Harmless, just slightly misleading.
- CI: only CodeRabbit has run here. Could a maintainer kick the full **Tests + smoke-matrix** (macOS/Windows/Linux)? The assembly path is central enough that cross-platform green is worth confirming before merge.

Happy to help with the >30s watermark test or the cache-clear throttling if useful.

debpalash (commented):
Thanks for this — the core approach is right and it targets a real pain (RAM/VRAM blowups on long videos and 50-clip batches). Streaming each segment to disk and assembling the final mix into a disk-backed `np.memmap` genuinely bounds peak memory instead of holding every segment tensor until assembly. A few things to confirm/tighten before merge.

### Correctness — please verify
1. **Watermark scope changed, not just relocated.** Previously each per-segment WAV *and* the final track were watermarked; now segment WAVs are written un-watermarked and only the final track is marked (chunk-by-chunk in `_write_memmap_wav_atomic`). The new `test_final_dub_track_is_watermarked_after_assembly` asserts exactly this. That's arguably cleaner (no double-embed), but it's a behavior change — please confirm per-segment `seg_i.wav` files are never surfaced to users as deliverables (they feed preview + partial-regen only, which then re-mix into the watermarked final). If any export path hands a raw `seg_i.wav` to the user, it would now be unwatermarked.
2. **Watermarking is now applied per 30-second chunk** rather than over the whole track in one call. Please validate with the **real AudioSeal** model (the test uses a fake `embed_watermark`) that `detect_watermark` still passes on the assembled track, including across chunk boundaries.
3. **The final track is now written by a second, hand-rolled WAV writer** (`_write_memmap_wav_atomic`, hard-coded 16-bit PCM mono via the `wave` module) instead of `atomic_save_wav` → `_safe_torchaudio_save`. Confirm bit-depth/format parity with the central `audio_io` path, since this bypasses its audited normalization. The mix was always mono so mono is fine; it's the encoding parity I'd nail down.

### Performance
4. `_release_audio_tensors()` runs `gc.collect()` **and** `torch.cuda/mps.empty_cache()` after *every* segment. `empty_cache()` forces a device sync and returns blocks to the OS, so on a 50-segment job this can measurably slow generation and fight the resident TTS model for the allocator. Suggest freeing the tensor every segment but gating `empty_cache()` to every N segments (or dropping it from the per-seg hot path — the tensors are already freed). This keeps the memory win without the per-seg sync cost.

### Nice catches
- The negative-`place_at` guard (`if s < 0: …`) fixes a latent out-of-range write the old `full_audio[:, s:e]` path had.
- Stereo down-mix on disk-loaded segments and `_entry_num_samples` via `torchaudio.info` (no full load) are both good.

### Before merge (project rules)
- Please fill in the **Summary** and **Testing** sections of the PR description.
- Add a `## [Unreleased]` **CHANGELOG.md** entry — this is a user-facing behavior/perf change and the repo treats stale docs as bugs (docs-sync rule).

Happy to re-review once the watermark-on-real-AudioSeal check and the `empty_cache` cadence are addressed.

debpalash:
Friendly nudge, @trungthanh1288 — this is the one PR genuinely waiting on you for the v0.3.8 line. The disk-backed memmap approach is sound and targets a real pain (the 8 GB-card OOM class). To get it mergeable, the open items from my review are: (1) confirm `detect_watermark` still passes on the assembled track with the **real AudioSeal** (the per-30s-chunk watermarking is the risk — the test uses a fake), (2) gate `empty_cache()` off the per-segment hot path (it syncs the device every segment), and (3) fill the PR Summary/Testing + add a `## [Unreleased]`/`[0.3.8]` CHANGELOG entry. Happy to re-review the moment those land — or if you're out of bandwidth, say so and I can carry it over.

debpalash:
Thanks for this, @trungthanh1288 — the streaming-to-disk architecture is the right fix and I've carried it forward in #709 with you credited via `Co-Authored-By`. An independent review pass surfaced a few things to make it merge-ready, which #709 addresses on top of your work:
- A crash on zero/negative-duration segments (real subtitle data has zero-length cues) — `torch.zeros(1, <=0)` → `atomic_save_wav` raised and killed the dub.
- Watermark provenance: since `seg_<id>.wav` is both the download file and the assembly input, #709 marks each fresh segment once at synthesis and drops the per-chunk embed (no double-mark), restoring the mark on downloadable seg WAVs. Verified at 0.9999 detect confidence with the real AudioSeal model.
- Perf (throttled `empty_cache()`, dropped per-seg `gc.collect()`), `mix_*` temp cleanup, a multi-chunk watermark test, and a CHANGELOG entry.

Closing this in favor of #709 — really appreciate the contribution.

Issue #976, opened 2026-07-06
[Install] SSL Handshake Failure when downloading required models (v0.3.11)


vodanhntmk:
### Where did it fail?

Engine install (CosyVoice / IndexTTS / MLX / etc.)

### The error

```text
Install failed: Got: ConnectError: [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] ssl/tls alert handshake failure (_ssl.c:1016) An error happened while trying to locate the files on the Hub and we cannot find the appropriate snapshot folder for the specified revision on the local disk. Please check your internet connection and try again.
```

### Operating system

Windows (x64)

### How are you installing it?

Desktop app (installer / AppImage)

### Version

v0.3.11

### Network conditions (model/dependency downloads)

Normal / unrestricted

### Logs / diagnostic bundle

```text
2026-07-06 14:31:20,878 INFO [omnivoice.setup.models] Loaded 32 models from C:\Users\ADMIN\AppData\Local\com.debpalash.omnivoice-studio\project\backend\config\models.yaml

2026-07-06 14:31:21,466 INFO [omnivoice.hf_progress] hf_progress: installed tqdm patch on huggingface_hub.utils.tqdm

2026-07-06 14:31:21,524 INFO [omnivoice.model] downloads: Xet disabled → legacy LFS (hf_xet 1.4.2 installed=True), high_perf=False

2026-07-06 14:31:24,468 INFO [omnivoice.api] MCP app mounted at /mcp

2026-07-06 14:31:24,471 INFO [omnivoice.api] Startup watchdog armed: thread dump if startup exceeds 300s (#632).

2026-07-06 14:31:27,833 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.schemas

2026-07-06 14:31:27,833 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.tables

2026-07-06 14:31:27,833 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.types

2026-07-06 14:31:27,833 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.constraints

2026-07-06 14:31:27,834 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.defaults

2026-07-06 14:31:27,834 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.comments

[...]

PR #992, opened 2026-07-07
fix(install): classify SSL handshake failures + trust the OS cert store

debpalash:
## The bug

Issue #976: Windows users behind a corporate/antivirus TLS-inspecting proxy get a raw, unhelpful error on every engine model install:
```
Install failed: Got: ConnectError: [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] ssl/tls alert handshake failure (_ssl.c:1016) ...
```
The TCP connection reaches the server fine — the handshake fails because the proxy re-signs traffic with a root CA the OS trusts (Windows CryptoAPI/SChannel) but Python's bundled `certifi` CA list doesn't. **This is a genuinely different failure mode from #984** — that was TCP-level unreachability to a fully blocked host, before any TLS negotiation begins; this is a TLS handshake failing after the TCP connection already succeeded.

## The fix

- **`backend/core/failure.py`**: new `SSL_HANDSHAKE_FAILURE` classification (handshake/cert-verify-failed/sslv3_alert/sslcertverificationerror substring markers) with an actionable hint, added to `_CONTEXT_FREE_HINT_CLASSES` so `append_hint()` — already called by the install worker in `setup/download.py` — surfaces it with zero further plumbing.
- **`backend/main.py`**: `truststore.inject_into_ssl()` at module level, before any `huggingface_hub`/`requests`/`httpx` network I/O. This is the actual root-cause fix, not just a nicer error: it patches `ssl.SSLContext` process-wide to verify against the OS trust store instead of only `certifi`'s bundled Mozilla CA list — so a corporate root CA the OS already trusts is now trusted by Python too. Not platform-gated (a correctness improvement everywhere); wrapped in try/except so it can never block startup.
- **`pyproject.toml`/`uv.lock`**: `truststore>=0.9` — pure Python, MIT, PyPA-maintained, zero transitive deps, same class of fix as the existing `socksio` addition.

## Verification

- `uv lock --check` + `uv sync --frozen`: clean (lockfile diff is exactly the one new package).
- `import main` succeeds cleanly with the new module-level call.
- No `backend.spec` `hiddenimports` entry needed — `backend/main.py` is PyInstaller's direct entry script, so a top-level `import truststore` traces normally (unlike `socksio`'s case, which was `httpx`'s *internal* lazy import, invisible to static analysis).
- `tests/test_failure_classify.py`: 10 passed, including the exact reported error string, a `CERTIFICATE_VERIFY_FAILED`-style variant, and a false-positive guard (a plain `ConnectionError` must not be misclassified).
- Full backend suite run clean prior to commit.

## Summary
- Added `SSL_HANDSHAKE_FAILURE` classification with TLS handshake / certificate verification markers and a `truststore`-oriented hint.
- Wired the new classification into context-free hinting so install-time failures can surface the guidance automatically.
- Injected `truststore` into SSL at `backend/main.py` import time, best-effort, so TLS validation uses the OS trust store before network clients initialize.
- Added `truststore>=0.9` as a runtime dependency.
- Expanded failure-classification tests to cover handshake alerts, certificate verification errors, and a non-match control case.

## Behavior
```mermaid
flowchart TD
  A[App starts] --> B[backend/main.py imports]
  B --> C[try truststore.inject_into_ssl()]
  C -->|success| D[SSL uses OS trust store]
  C -->|failure ignored| E[Continue startup]
  D --> F[Network request during model install]
  E --> F
  F --> G[Corporate proxy re-signs TLS]
  G --> H[Handshake/cert error classified]
  H --> I[SSL_HANDSHAKE_FAILURE hint shown]
```

### DIFF
--- a/backend/core/failure.py
+++ b/backend/core/failure.py
@@ -42,6 +42,7 @@ _HINTS: dict[str, str] = {
     "TRANSFORMERS_IMPORT": "Your transformers install is incomplete. Reinstall it (`uv pip install --reinstall transformers`) or switch ASR to faster-whisper (Settings → Models).",
     "OS_INVALID_ARGUMENT": "The OS rejected a file operation (Errno 22 / invalid argument) — in the transcribe path this is the temporary WAV write before ASR. It's almost always the temp directory: missing, read-only, on a full or removed drive, or blocked by antivirus. Check that your system TEMP/TMP folder exists and is writable and the drive has free space (add an OmniVoice antivirus exclusion if you use one), then retry.",
     "SOCKS_PROXY_SUPPORT_MISSING": "A SOCKS proxy is configured in your environment (ALL_PROXY/HTTPS_PROXY=socks5://…) and the backend's HTTP client is missing SOCKS support. Newer OmniVoice builds ship SOCKS support (the socksio package) — update the app. If you still see this, unset ALL_PROXY/HTTPS_PROXY for OmniVoice, or run `uv pip install 'httpx[socks]'` in the backend venv, then restart.",
+    "SSL_HANDSHAKE_FAILURE": "A corporate or antivirus proxy is intercepting HTTPS traffic and re-signing certificates with its own CA — your OS trusts that CA, but Python's bundled certifi CA list doesn't, so the TLS handshake fails even though the connection reached the server. Newer OmniVoice builds trust the OS certificate store at startup (the truststore package), which should already fix this — update the app and retry. If you still see this, add an HTTPS-scanning exclusion for OmniVoice/Python in your antivirus, or ask IT for the proxy's CA bundle and set SSL_CERT_FILE to it, then restart.",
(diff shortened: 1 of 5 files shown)


debpalash:
**Owner decision on CodeRabbit's Cross-Platform Default Parity warning: the unconditional `truststore.inject_into_ssl()` stays.**

Rationale: the suggested remedy (gate the injection to Windows or an opt-in) would reintroduce the exact bug this PR fixes for macOS and Linux users — corporate/antivirus TLS-inspecting proxies exist on all three platforms, and the reporter class this addressed isn't Windows-specific. The "parity" concern is real but inherent: `truststore` delegates to each OS's native trust store, which by definition differ per machine — the *behavior* ("trust what the OS trusts") is identical everywhere, which is the parity that matters for the project's cross-platform rule. Python's bundled certifi list was already a divergence from what the user's OS/browser trusts; this change makes TLS behavior *more* consistent with user expectations, not less.

Decision documented here so the finding doesn't resurface as an unresolved thread.

Issue #977, opened 2026-07-06
[Bug] 500 Internal Server Error: Couldn't synthesize audio. See Settings → Logs → Back


scprdytj2s-beep:
## Error

```
500 Internal Server Error: Couldn't synthesize audio. See Settings → Logs → Backend for the full trace. Underlying error: TTS engine stopped mid-generation with an error OmniVoice doesn't recognize. Retry once; if it keeps failing, please report it with the full trace. Underlying error: ('du', {'a': 'American English', 'b': 'British English', 'e': 'es', 'f': 'fr-fr', 'h': 'hi', 'i': 'it', 'p': 'pt-br', 'j': 'Japanese', 'z': 'Mandarin Chinese'})
v@tauri://localhost/assets/client-DU5MrUgu.js:2:3494
```

## Environment

**Version:** `0.3.11`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.3.1-arm64-arm-64bit`
**Python:** `3.11.15`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M2 Max`
**RAM:** `32 GB`
**Disk free:** `142.3 GB`
**Active TTS engine:** `mlx-audio`

## Recent actions

```
11:59:24 generate:start (audio)
12:17:08 view:projects
12:17:09 view:transcriptions
12:17:09 view:gallery
12:17:10 view:audiobook
12:17:11 view:stories
12:17:11 view:dub
12:17:26 view:settings
12:18:40 view:launchpad
12:18:41 view:studio
12:18:48 generate:start (audio)
12:20:38 generate:start (audio)
12:25:31 generate:start (audio)
12:26:13 generate:start (audio)
12:26:58 generate:start (audio)
12:29:19 view:settings
12:35:45 engine:tts=mlx-audio
12:35:52 view:studio
12:35:55 generate:start (audio)
12:46:58 generate:start (audio)
```

debpalash:
[...]

Issue #877, opened 2026-07-02
feat(asr): adopt Qwen3-ASR-0.6B (30 langs + 22 zh dialects, ~2-3× realtime on CPU) once transformers ≥5.13 is stable


debpalash:
## Summary

Adopt **Qwen/Qwen3-ASR-0.6B-hf** as an opt-in ASR engine once a stable `transformers` release (≥ 5.13) ships the `qwen3_asr` model type. Evaluated 2026-07-02 on an Apple Silicon M2 (CPU, one model at a time) against the current default:

| Model | RTF (CPU) | Accuracy on test clips | Languages |
|---|---|---|---|
| faster-whisper large-v3 int8 (default) | 2.0–4.8 | good | ~100 |
| parakeet-tdt-0.6b-v3 (now un-gated, #871) | 0.08–0.23 | perfect | 25 EU |
| **Qwen3-ASR-0.6B** | **0.31–0.5** | **perfect** | **30 + 22 zh dialects** |
| Qwen3-ASR-1.7B | 1.5–3.1 | perfect | 30 + 22 zh dialects |
| nemotron-3.5-asr-streaming-0.6b | 0.89–8.6 (erratic) | good | 40 |

Verdict: 0.6B is the sweet spot — near-Parakeet speed with much broader language coverage (esp. Chinese dialects); complements rather than replaces whisper/parakeet. 1.7B and nemotron: declined (details in the eval; 1.7B is 3–6× slower for identical transcripts, nemotron targets GPU-server streaming).

## Blocker

`qwen3_asr` exists only on transformers git main (5.13.0.dev0); the repo pins `>=5.3` and stable is 5.12.1. **Do not pin a git ref** — it breaks lockfile reproducibility and the Docker frozen-lockfile gate. The backend below capability-gates at runtime (probes `CONFIG_MAPPING_NAMES`), so it can merge before or after the 5.13 bump and simply reports "transformers>=5.13 required" until then.

## Ready-to-land scaffold (validated against the eval)

Notes: load with explicit dtype (fp16 CUDA / fp32 CPU — bf16 weights + fp32 features crash CPU conv2d); decode strips the `language X<asr_text>` prefix; word timestamps need the separate ForcedAligner model (out of scope, `words: []` like other non-aligning backends).

<details><summary><b>backend/services/… Qwen3ASRBackend</b> (paste above the Registry section of asr_backend.py)</summary>

```python
[...]

PR #1003, opened 2026-07-08
feat(asr): generic OpenAI-compatible transcription backend

debpalash:
## First slice of #877's community proposal

The full ask (adopt Qwen3-ASR-0.6B directly) is blocked on `transformers>=5.13` stabilizing upstream — still true, still tracked separately, not attempted here. But a community member (LauraGPT) proposed splitting this into two tracks so OmniVoice doesn't have to wait: **add a generic OpenAI-compatible transcription backend now**, so anyone who wants Qwen3-ASR today can run it behind a compatible server (or use FunASR/SenseVoice self-hosted, or OpenAI's own API). This PR is that first track.

## What's new

- **`OpenAICompatASRBackend`** — a pure network client (no local model, no install) that calls `POST /v1/audio/transcriptions` on whatever server you configure. Prefers `response_format=verbose_json` for real per-segment timestamps; degrades to plain text (matching `MoonshineASRBackend`'s shape) for minimal servers that reject it. Never leaks a raw SDK/httpx exception to the user — wraps failures in a clean, actionable message (the #977 convention from earlier this session).
- **Settings**, encrypted the same way `llm_providers.py` already does it — the API key goes through `settings_store.set_secret` (Fernet-encrypted, per-install key, never a `.env` row, never echoed back); base_url/model are plain settings rows. New loopback-gated `GET`/`PUT /api/settings/asr-openai-compat`.
- **A small settings panel** (Settings → Models), structurally identical to `HFMirrorPanel`. Honest caveat documented plainly: there's no ASR engine *picker* UI yet for any ASR backend (only TTS has one) — activating this still needs `OMNIVOICE_ASR_BACKEND=openai-compat-asr`.
- README's ASR Engines table (9 → 10) and `docs/features.yaml`'s drift-checker inventory updated — corrected the "all fully local" claim, since this one genuinely sends audio to wherever you point it.
- `docs/engines/openai-compatible-asr.md` — setup steps + an explicit privacy note.

## Tests

New `tests/test_asr_openai_compat_877.py` (12 tests): availability gating, both response-adaptation shapes, network-failure error hygiene, SDK retry disabling (`max_retries=0`, matching `llm_skills.py`'s convention), and the settings endpoints' persist/mask/clear-vs-unchanged semantics.

## A real lesson from verification (not glossed over)

Testing this in isolation looked perfect (12/12), but the **full** suite caught two real problems I fixed properly rather than dismissed:
1. The API route-inventory snapshot (`tests/fixtures/api_routes.txt`) needed regenerating for the two new routes — expected maintenance, done.
2. My own test file had a genuine **module-staleness bug**: a collection-time `from services import settings_store` import went stale relative to a test-time-fresh fixture, because some other test elsewhere in the ~2400-test suite pops/reimports `services.settings_store` mid-session. Same root cause (and same fix — make the module a fixture resolved fresh at test-run time) as a bug I fixed earlier this session in `tests/test_mm2_lifecycle.py`. Worth flagging as a recurring hazard in this codebase's test suite for future test authors.

Full backend suite after the fix: **2402 passed, 0 failed**. Full frontend suite: **918 passed, 0 failed**. `docs-drift`: clean. CJK guard: clean. `bun run format:check`: clean.

Added a new OpenAI-compatible remote ASR backend (`openai-compat-asr`) that transcribes via `POST /v1/audio/transcriptions`, prefers `verbose_json` for segment timestamps, falls back to plain text/json when needed, and wraps SDK/network failures in user-facing errors with retries disabled.

Also added encrypted settings for the backend’s base URL, model, and API key, exposed through loopback-gated `GET`/`PUT` routes at `/api/settings/asr-openai-compat`, and wired a new Settings → Models panel for configuring it.

Updated docs and feature metadata to reflect the extra ASR engine, the remote nature of this option, and its setup/privacy implications.

Behavior flow:
```mermaid
flowchart TD
  A[Settings / backend config] --> B[Resolve base_url, model, api_key]
  B --> C{OpenAI package + base_url present?}
  C -- no --> D[Backend unavailable]
  C -- yes --> E[Create OpenAI client with max_retries=0]
  E --> F[POST /v1/audio/transcriptions]
  F --> G{verbose_json accepted?}
  G -- yes --> H[Normalize segments + chunks]
  G -- no --> I[Fallback to plain text/json]
  I --> H
  F --> J[Wrap failures in actionable RuntimeError]
```

UI sketch:
```text
Before: Settings → Models
+----------------------------+
| Storage                    |
| HF Mirror                  |
+----------------------------+

After: Settings → Models
+--------------------------------------+
| Storage                              |
| OpenAI-compatible ASR (remote)      |
|   Base URL [......................]  |
|   Model    [......................]  |
|   API key  [......................]  |
|   [Save]                            |
| HF Mirror                            |
+--------------------------------------+
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,10 +322,10 @@ Professional-grade voice AI, minus the subscription and the cloud.
 
 ### 🎧 ASR Engines
 
-**9 engines, all fully local** — they power dictation, video dubbing, and subtitles. **WhisperX** is the cross-platform default (~100 languages, word-level timing); the rest are opt-in and auto-detected. Switch in **Settings → ASR Engine** or via the `OMNIVOICE_ASR_BACKEND` env var.
+**10 engines** — they power dictation, video dubbing, and subtitles. **WhisperX** is the cross-platform default (~100 languages, word-level timing); the rest are opt-in and auto-detected. Switch in **Settings → ASR Engine** or via the `OMNIVOICE_ASR_BACKEND` env var. Nine run fully on-device; one (OpenAI-compatible) is an optional remote client for pointing at Qwen3-ASR or another compatible server — see below.
 
 <details>
-<summary><b>📊 The full lineup</b> — 9 engines, what each is best at, and compute-type notes</summary>
+<summary><b>📊 The full lineup</b> — 10 engines, what each is best at, and compute-type notes</summary>
 
 <br/>
 
@@ -340,6 +340,7 @@ Professional-grade voice AI, minus the subscription and the cloud.
 | **Moonshine** | `moonshine` | English | Edge / low-latency, ONNX |
 | **FunASR** | `funasr` | 50+ | All-in-one multilingual — built-in VAD + inline speaker diarization (SenseVoice) |
 | **sherpa-onnx** (live dictation) | `sherpa-onnx-asr` | 25 EU + 90+ | Live, faster-than-real-time dictation — small streaming/offline ONNX models (Parakeet TDT v3/v2, streaming Zipformer & Paraformer, Whisper Tiny), CPU, identical on macOS / Windows / Linux. Picked per-model in **Settings → Voice**. |
+| **OpenAI-compatible** ⚠️ remote | `openai-compat-asr` | Server-dependent | A path to **Qwen3-ASR** today (self-hosted server, no transformers wait), any OpenAI-compatible transcription endpoint, or OpenAI's own API — no install, configure in **Settings → Models**. Audio leaves your machine to whatever server you point it at; see [docs/engines/openai-compatible-asr.md](docs/engines/openai-compatible-asr.md). |
 
 > Whisper-family engines cover ~100 languages; **FunASR / SenseVoice** adds an all-in-one multilingual path with built-in voice-activity detection and inline speaker diarization. **sherpa-onnx** powers the live dictation model picker — you talk and text appears as you speak. Every engine runs on-device — no API keys, no cloud.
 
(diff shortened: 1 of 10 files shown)


LauraGPT:
This is a good first slice of #877. Keeping Qwen3-ASR / FunASR / SenseVoice behind a generic `POST /v1/audio/transcriptions` client lets users test those engines now without blocking this repo on a direct `transformers` release.

A few small hardening points would make this much easier to merge and run safely:

- Add an explicit request timeout to the OpenAI client or transcription call. A LAN/self-hosted ASR server can hang during model load or GPU queueing, and dictation/video flows should fail with an actionable message instead of waiting on the SDK default.
- Narrow the `verbose_json` fallback. Retrying as `response_format="json"` is right when the server rejects the format, but auth failures, connection errors, timeouts, and 5xx responses should surface directly rather than making a second request and obscuring the real failure.
- The backend supports `api_key=""` to clear the saved key, but the current settings panel only omits `api_key` when the field is blank, which means users cannot clear a stored key from the UI. A small “clear saved key” action that sends `{api_key: ""}` would close that loop.
- The `RuntimeError` wrapper should avoid echoing the full raw SDK exception if a compatible server includes request/body details in its error text. Status class plus a short hint is usually enough for users to debug base URL/auth/model issues.

For FunASR/SenseVoice specifically, this contract is the useful part: base URL should point at the server’s `/v1` root, the request stays standard multipart audio, and `verbose_json`/plain-json fallback keeps timestamp-capable and minimal compatible servers on the same path.

Issue #730, opened 2026-06-28
[Tracking] Windows: backend goes unreachable after a whisperx/ASR transcribe hangs


debpalash:
**Tracking issue** consolidating the recent wave of *"Can't reach the local OmniVoice backend — it may still be starting up, or it stopped"* reports.

### Symptom
The desktop UI shows "Can't reach the local backend" — usually right after a `generate:start (audio)` or a dub transcribe. The backend process is up but unresponsive (health checks fail), so every request 500s / times out.

### Root cause (from #723's logs, on 0.3.8-55 / RTX 4060 Ti 8 GB Windows)
A `whisperx large-v3 (float16)` transcribe starts and then the backend goes **silent for hours** until it's restarted — i.e. the transcribe call **hangs hard** (inside CTranslate2 / faster-whisper or the VAD step) and never returns. 0.3.8's `#656` bound (`transcribe_with_timeout` → `asyncio.wait_for`) surfaces a timeout message but **cannot kill the stuck worker thread**, so the GPU worker pool starves and the backend stays unreachable.

### What 0.3.8 already fixes (helps the 0.3.7 cohort)
- `#656` — bounds whole-file transcription so a *stall* is reported instead of a silent "can't reach backend"
- `#692` — ASR falls back instead of crashing when CTranslate2's `.so` won't load
- `#630/#655` — speechbrain lazy-import guard unblocks WhisperX on Windows

### Still open (the residual, this tracker)
Make a hung whisperx/ASR transcribe **recoverable** on Windows so it can't brick the backend (kill the worker / route through the killable ASR sidecar / prevent the hang). Tracked here.

### Reports folded in
- Pre-fix (0.3.7/0.3.5, will largely clear on the 0.3.8 build): #724, #725, #727, #728, #729, #726
- On `main` (0.3.8-55, the live evidence): #720 (transcribe stream dropped), #721, #723 (has logs)

If you're on **0.3.7 or older**: please update to the 0.3.8 build when it's out and retest. If you can still reproduce on 0.3.8, drop your `Settings → Logs → Backend` tail here.

debpalash:
[...]

Issue #255, opened 2026-06-02
[Bug] Transcribe stream dropped before emitting any segments


camilocarrai1997:
## Describe the bug

The transcription is still failing

## To reproduce

I just followed these steps to fix the pkg_resources error:
https://github.com/debpalash/OmniVoice-Studio/pull/253

The problem is that now I get the following error when the transcription is progressing:
Upload failed: Transcribe stream dropped before emitting any segments. Likely ASR backend failed to load — check backend log + Settings → Models.

Im using Whisper large-v3 (faster-whisper — default, cross-platform)

## Expected behavior

Transcribe video

## Screenshots / Logs

<img width="753" height="143" alt="Image" src="https://github.com/user-attachments/assets/febb7b0c-b162-4e31-b582-64d6440baad3" />

<img width="1901" height="65" alt="Image" src="https://github.com/user-attachments/assets/4f8cedc9-220e-4ec8-9827-64bb74e33d11" />

## Environment

- **OS:** Windows 11
- **Install method:** Desktop app
- **Version:** 0.3.0
- **GPU:** NVIDIA RTX 4070 ti Super
- **RAM:** 32GB

## Additional context

This error also keeps appearing:

Upload failed: No ASR backend is ready. Install WhisperX/faster-whisper/MLX Whisper or set OMNIVOICE_PRELOAD_TTS_ASR=1 before launching to use the PyTorch fallback.
 
This error appears when i use PyTorch Whisper (CUDA / CPU via transformers pipeline)

debpalash:
Thanks for the detailed report, and for confirming you'd already applied #253.

The message you saw — *"Transcribe stream dropped before emitting any segments…"* — is unfortunately a **generic fallback** the UI shows whenever the transcribe stream closes without sending a specific error. It was hiding the real cause. I've pushed a fix (#259) that makes the backend always emit the **actual** failure (sanitized) instead of dropping the stream, so the next run will tell us exactly what's wrong.

On Windows + NVIDIA, the most common root cause is **faster-whisper / CTranslate2 not finding cuDNN 8**. PyTorch 2.8 ships cuDNN 9, but CTranslate2 needs cuDNN 8, so the app side-loads a cuDNN-8 copy from `.venv/Lib/site-packages/cudnn8_compat/`. If that folder is missing (some upgrade/repair paths don't run the cuDNN-8 install step), the model fails to load natively.

Could you help confirm, once you're on a build that includes #259:
[...]

PR #1005, opened 2026-07-08
fix(clone): re-transcribe written reference clips so (ref_audio, ref_text) match (#1004)

### LINKED ISSUE #1004: Cross-language dub can speak the source-language reference text verbatim — auto-clone (ref_audio, ref_text) pairs can mismatch
## Symptom

When dubbing across languages (Korean source → Spanish target, auto speaker clones), some generated segments come out **entirely in the source language** — the TTS speaks the clone reference sentence verbatim instead of the translated text. Other takes of the same segment render the target text but with source-language phonetics heavy enough that Whisper language-detects them as `ko`/`ja`.

## Root cause

`services/speaker_clone.py` builds clone references by slicing the Demucs vocals at ASR segment timestamps and pairing that audio with the ASR segment `text` (`extract_speaker_clones`, and `extract_segment_refs` line ~215). This assumes the ASR text exactly matches the audio inside `[start, end]` — but Whisper's segment text and its timestamps routinely drift (trailing words missing from the text while present in the audio, words cut at boundaries, etc.).

When the `(ref_audio, ref_text)` pair disagrees, the OmniVoice TTS prompt priming breaks down and the model can emit the **reference text** (source language) instead of the target text.

## Evidence (A/B on the same audio clip)

Reference clip: one speaker's line sliced from vocals (6.5 s). The ASR segment text was missing the trailing word that IS audible in the clip.

| ref_text | result of 6 generations of a Spanish test sentence |
|---|---|
| ASR segment text (mismatched: trailing word missing) | 6/6 wrong language — Whisper detects `ko`/`ja`/`zh`, output speaks the Korean reference sentence |
| Whisper re-transcription of the exact same clip (matched) | clean Spanish on the 1st try |

Reproduced consistently across two different speakers in the same video; every mismatched-pair reference showed the behaviour, every matched-pair reference was clean.

## Proposed fix

After writing each reference wav, re-transcribe the actual clip via the active ASR backend (the registry path introduced in #321) and use that transcript as `ref_text`, falling back to the segment text when ASR fails. This guarantees the pair matches by construction. The ASR model is already warm at extraction time (it just ran the transcription pass).

Happy to submit a PR.

## Environment

- Image: `palashdeb/omnivoice-studio:stable` (Docker, CUDA / RTX 3090)
- Pipeline: ko → es dub, pyannote diarization enabled, per-speaker + per-segment auto clones

stronghamjji:
Closes #1004

## Problem

`extract_speaker_clones` / `extract_segment_refs` pair the sliced vocals audio with the ASR segment text. Whisper's segment text routinely drifts from what is actually audible inside `[start, end]` (boundary jitter, dropped trailing words). When the `(ref_audio, ref_text)` pair disagrees, TTS prompt priming breaks down and cross-language dubs can speak the **reference text verbatim in the source language** — see the A/B evidence in #1004 (mismatched pair: 6/6 generations came out in the source language; matched pair: clean target-language output on the first try).

The module docstring already states the invariant this PR enforces: *"The accompanying transcript text must align with the audio slice or the TTS cloner will mis-align its phoneme lookups."*

## Fix

After writing each reference wav (per-speaker and per-segment), re-transcribe the exact written clip via the active ASR backend and use that transcript as `ref_text`. The pair now matches by construction.

- `speaker_clone.py`: new `_retranscribe_ref()` helper + optional `transcriber` kwarg on both extract functions. `transcriber=None` (default) keeps the previous behaviour, so existing callers/tests are unaffected.
- `dub_core.py`: the transcribe flow passes a small wrapper around the live `_asr_backend` (already warm — the transcription pass just used it). Reads `text` with a `chunks`-join fallback so every ASR backend shape works.
- Any ASR failure logs a warning and falls back to the segment text — never worse than before.

## Cost

One extra short ASR call per written reference on a warm backend. Per-speaker refs are few; per-segment refs add roughly one call per ≥3 s segment.

## Verification (real ko→es dub, Docker/CUDA)

- Before: the CEO speaker's auto-clone `ref_text` was missing the trailing word audible in the clip; every generation using it spoke Korean instead of Spanish.
- After: stored `ref_text` matches an independent transcription of the written clips at 0.97–0.99 similarity for all speaker clones, and the verbatim source-language failure no longer reproduces.

Fixes cross-language dubbing failures caused by `(ref_audio, ref_text)` mismatches by re-transcribing the exact WAV reference clip immediately after it’s written, using the already-warmed ASR backend; the re-transcript replaces `ref_text` for both per-speaker and per-segment clone references (falling back to the original segment text if re-transcription fails or times out).

```mermaid
flowchart LR
  A[Slice vocal audio window(s)] --> B[Write reference WAV(s)]
  B --> C{transcriber provided?}
  C -- no --> Z[Use computed segment text as ref_text]
  C -- yes --> D{one-way wedged latch?}
  D -- yes --> Y[Return \"\" immediately]
  D -- no --> E[Start daemon thread:\n_asr_backend.transcribe(ref)\njoin <= ASR_TRANSCRIBE_TIMEOUT_S]
  E --> F{thread finished?}
  F -- no --> G[Set _ref_tx_wedged flag;\nlog warning (sanitized path);\nreturn \"\"]
  F -- yes --> H[Extract transcript:\nprefer r.text else concat chunk.text;\ntrim]
  H --> Y[Return transcript (possibly empty)]
  G --> Y
  Y --> I[_retranscribe_ref:\nif empty/exception -> fallback segment text\nelse use re-transcript as ref_text]
  I --> J[Clone/TTS prompt priming uses matching (ref_audio, ref_text)]
```

- `backend/services/speaker_clone.py`
  - Added `transcriber: Callable[[str], str] | None` support to both:
    - `extract_speaker_clones(...)` (overwrites `ref_text` from the written per-speaker WAV)
    - `extract_segment_refs(...)` (overwrites `ref_text` from the written per-segment WAV)
  - New `_retranscribe_ref(...)` helper:
    - if no transcriber: returns `fallback`
    - if transcriber throws or returns empty/whitespace: logs warning (by label) and falls back to `fallback`
- `backend/api/routers/dub_core.py`
  - In the `/dub/transcribe-stream/{job_id}` SSE auto-speaker-clone flow, added `_ref_transcriber(path)`:
    - logs reference paths via `core.failure.sanitize`
    - runs ` _asr_backend.transcribe(path, word_timestamps=False)` in a joined daemon thread bounded by `ASR_TRANSCRIBE_TIMEOUT_S`
    - on timeout: sets a one-way latch (`_ref_tx_wedged`) so all later reference re-transcriptions short-circuit and return `""`, preventing backend wedging/races
    - falls back to segment text when the helper returns `""` (handled by `_retranscribe_ref`)
- Net effect: ensures the audible slice’s transcript matches the exact written reference clip, preventing prompts from priming with source-language reference text in cross-language dubs.

### DIFF
--- a/backend/api/routers/dub_core.py
+++ b/backend/api/routers/dub_core.py
@@ -977,6 +977,9 @@ async def dub_transcribe_stream(
             from services.speaker_clone import extract_speaker_clones, auto_profile_id
             vocals_for_clone = job.get("vocals_path") or asr_audio_target
             clones = {}
+            # One-way latch: set when a reference re-transcription wedges past
+            # its timeout, so later references skip ASR entirely (see below).
+            _ref_tx_wedged = {"flag": False}
 
             def _ref_transcriber(path: str) -> str:
                 # Re-transcribe the exact written reference clip so its
@@ -991,8 +994,14 @@ async def dub_transcribe_stream(
                 # with a joined side thread instead. On timeout we return ""
                 # (caller falls back to the segment text) rather than letting
                 # a wedged transcribe (#730) keep the SSE ping loop alive
-                # forever. The abandoned daemon thread mirrors the guarded
-                # path's known tradeoff.
+                # forever. Once one call wedges, all later calls short-circuit:
+                # the abandoned thread may still hold the backend, and a second
+                # concurrent transcribe on a GPU backend risks corrupting its
+                # CUDA context.
+                from core.failure import sanitize
+
+                if _ref_tx_wedged["flag"]:
+                    return ""
                 done: list[str] = []
 
                 def _run():
@@ -1005,15 +1014,21 @@ async def dub_transcribe_stream(
                             ).strip()
                         done.append(text)
                     except Exception as e:
-                        logger.warning("ref re-transcription failed for %s: %s", path, e)
+                        logger.warning(
+                            "ref re-transcription failed for %s: %s", sanitize(path), e,
+                        )
+                        done.append("")
 
                 from services.asr_backend import ASR_TRANSCRIBE_TIMEOUT_S
                 t = threading.Thread(target=_run, daemon=True)
                 t.start()
                 t.join(ASR_TRANSCRIBE_TIMEOUT_S)
                 if not done:
+                    _ref_tx_wedged["flag"] = True
                     logger.warning(
-                        "ref re-transcription timed out for %s; keeping segment text", path,
+                        "ref re-transcription timed out for %s; falling back to "
+                        "segment text for this and all remaining references",
+                        sanitize(path),
                     )
                     return ""
                 return done[0]
--- a/backend/api/routers/dub_core.py
+++ b/backend/api/routers/dub_core.py
@@ -976,6 +976,22 @@ async def dub_transcribe_stream(
             from services.speaker_clone import extract_speaker_clones, auto_profile_id
             vocals_for_clone = job.get("vocals_path") or asr_audio_target
             clones = {}
+
+            def _ref_transcriber(path: str) -> str:
+                # Re-transcribe the exact written reference clip so its
+                # (ref_audio, ref_text) pair matches by construction — the
+                # ASR segment text can drift from what is audible inside the
+                # sliced window, and a mismatched pair makes cross-language
+                # dubs speak the reference text verbatim (#1004). The ASR
+                # backend is still warm from the transcription pass above.
+                r = _asr_backend.transcribe(path, word_timestamps=False)
+                text = (r.get("text") or "").strip()
+                if not text:
+                    text = " ".join(
+                        (c.get("text") or "").strip() for c in (r.get("chunks") or [])
+                    ).strip()
+                return text
+
             if labels_source == "heuristic":
                 # Clone-purity guard: heuristic labels are silence-gap
                 # estimates, not voice identity — a per-speaker reference cut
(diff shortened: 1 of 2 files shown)


stronghamjji:
Addressed the Greptile P1 in the follow-up commit: the reference re-transcription now runs on a joined daemon side thread bounded by `ASR_TRANSCRIBE_TIMEOUT_S` (the async `run_transcribe_guarded` can't wrap a call made inside the clone-extraction executor thread). On timeout it returns an empty string and `_retranscribe_ref` falls back to the segment text, so a wedged backend (#730) can no longer keep the transcribe SSE ping loop alive.

stronghamjji:
Addressed both follow-up findings: (1) a one-way latch now short-circuits all remaining reference re-transcriptions after the first timeout, so an abandoned wedged thread can never race a second `transcribe()` call or the later `unload()`; (2) reference clip paths are logged through `core.failure.sanitize` per the #255 convention. ASR exceptions are distinguished from timeouts (exception → fall back for that clip only; timeout → latch).

debpalash:
First — thank you. Your #1004 report was one of the clearest this repo has received (the A/B table made the root cause undeniable), and this PR is a clean implementation of exactly the right fix.

Second — an honest apology: you said you'd submit a PR, you did, and I then implemented the same fix hours later (#1008, merged) without checking the open-PR queue first. That's my process failure, not any shortcoming of this PR. The merged version is functionally equivalent to yours (re-transcribe each written clip via the already-warm active ASR backend, fall back to the original text on failure), with regression tests covering the mismatch-correction path, ASR-failure fallback, and batch behavior.

Closing this as superseded by #1008 rather than asking you to rebase against your own already-shipped idea. The fix that shipped is your diagnosis and your proposed design — the CHANGELOG entry credits the report's root-cause work directly. If you spot anything in the merged version that your implementation handled better, please open an issue or PR and I'll take it seriously (and check the queue first this time).

Issue #1013, opened 2026-07-08
New issue — microphone:


ssummone:
### Before filing

- [x] I searched [existing issues](https://github.com/debpalash/OmniVoice-Studio/issues?q=is%3Aissue) and this isn't a duplicate.
- [x] I'm on the latest release (or `main`) — older builds may already be fixed.

### What happened?

New issue — microphone:

## Describe the bug
Live recording is impossible on macOS: clicking record shows "Microphone
access denied. macOS: open System Settings → Privacy & Security →
Microphone and enable OmniVoice" — but OmniVoice never appears in that
list, so there is nothing to enable.

macOS only lists apps that have made a mic request. The Tauri webview's
getUserMedia permission request is never forwarded to the OS (no
permission-request handler on the WKWebView), so it auto-denies inside
the app and macOS never sees a request. `NSMicrophoneUsageDescription`
IS present in Info.plist, and `tccutil reset Microphone
com.debpalash.omnivoice-studio` + relaunch changes nothing — no prompt
ever appears.

## Steps to reproduce
1. Fresh install of the 0.3.12 DMG on macOS (Apple Silicon)
2. Studio → try to record a voice sample
3. "Microphone access denied" instantly, no macOS permission prompt
4. System Settings → Privacy & Security → Microphone: OmniVoice absent

## Environment
Version 0.3.12 · macOS 26.5.1 arm64 · Apple M4 Pro · install: DMG

## Workaround
Record in another app and upload the file — upload-based cloning works.

Comment to add on your existing IndexError issue:

Root cause found. In `backend/services/tts_backend.py`, MLXAudioBackend's
`generate()` reads `voice`, `ref_audio`, `language`, `speed` from kwargs
but drops `ref_text` — it's never added to the kwargs passed to
`self._model.generate()`. The sesame/CSM model only builds its cloning
context when BOTH `ref_audio` and `ref_text` are present
(`sesame.py` ~line 693); with `ref_text` missing, `context` stays empty
and `context[0].text` raises `IndexError: list index out of range`
(sesame.py:715). So voice cloning on the CSM engine can never work as
shipped — and the ref transcript the backend auto-generates upstream
(generation.py ~780) is thrown away right before it's needed.

Fix that worked for me — pass it through:

    ref_text = kw.get("ref_text")
    ...
    if ref_audio and ref_text: kwargs["ref_text"] = ref_text

With that two-line change, CSM cloning works end-to-end (MPS, 0.3.12).

[...]

PR #1015, opened 2026-07-08
fix(tts): forward ref_text in MLXAudioBackend so CSM cloning doesn't crash

MahdiHedhli:
## Summary
- `MLXAudioBackend.generate()` read `voice`/`ref_audio`/`language`/`speed` from `kwargs` but never read `ref_text`, so it was silently dropped before reaching `self._model.generate()`.
- CSM (sesame) only builds its voice-clone context when **both** `ref_audio` and `ref_text` are present. Without `ref_text`, the context stays empty and indexing it later raises an unrelated-looking `IndexError: list index out of range` — so voice cloning on the CSM engine could never work as shipped, even though `generation.py` already has a `ref_text` by the time it calls the backend (user-supplied, or auto-transcribed at `generation.py:~780` when absent).
- Every sibling backend in this file (`omnivoice_gguf`, `moss_tts_v15`, `dots_tts`) already forwards `ref_text` — `MLXAudioBackend` was the outlier.

## Fix
Forward `ref_text` whenever `ref_audio` is also present, matching the existing pattern used by the other `ref_text`-accepting backends:

```python
if ref_audio and ref_text: kwargs["ref_text"] = ref_text
```

## Test plan
- Added `tests/backend/services/test_mlx_audio_ref_text.py` (3 tests, mocked model — no real `mlx_audio`/CSM weights needed): `ref_text` is forwarded when `ref_audio` is present, omitted when `ref_audio` is absent, and the no-ref design/instruct path is unaffected.
- `uv run pytest backend/ -x -q` — 132 passed.

Related: surfaced while triaging #1013 (a separate mic-permission issue, whose body had this diagnosis pasted into it by mistake — not related to this fix).

## Contribution license
Submitting under AGPL-3.0 per CONTRIBUTING.md, including the commercial-dual-license grant to the maintainer.

## Summary
- Fixed `MLXAudioBackend.generate()` to forward `ref_text` only when `ref_audio` is present, matching the behavior of the other TTS backends.
- Added regression tests covering:
  - `ref_text` is forwarded with `ref_audio`
  - `ref_text` is omitted without `ref_audio`
  - the no-reference path remains unchanged

```mermaid
flowchart TD
  A[generate() called] --> B{ref_audio provided?}
  B -- no --> C[Do not forward ref_text]
  B -- yes --> D{ref_text provided?}
  D -- yes --> E[Forward ref_audio + ref_text to MLX model]
  D -- no --> F[Forward ref_audio only]
```

### DIFF
--- a/backend/services/tts_backend.py
+++ b/backend/services/tts_backend.py
@@ -849,6 +849,7 @@ class MLXAudioBackend(TTSBackend):
 
         voice     = kw.get("voice")
         ref_audio = kw.get("ref_audio")
+        ref_text  = kw.get("ref_text")
         language  = kw.get("language")
         speed     = float(kw.get("speed", 1.0))
 
@@ -858,6 +859,12 @@ class MLXAudioBackend(TTSBackend):
         # — we pass them all and let the engine ignore what it doesn't use.
         kwargs = {"text": text, "speed": speed}
         if voice:     kwargs["voice"] = voice
+        # CSM (sesame) only builds its voice-clone context when BOTH ref_audio
+        # and ref_text are present — with ref_text missing it silently builds
+        # an empty context and later crashes with an unrelated-looking
+        # `IndexError: list index out of range`. Forward it whenever we have
+        # both, same as every other ref_text-accepting backend in this file.
+        if ref_audio and ref_text: kwargs["ref_text"] = ref_text
         if ref_audio: kwargs["ref_audio"] = ref_audio
         if language and language != "Auto":
             if self._model_id == self.CURATED_MODELS.get("kokoro"):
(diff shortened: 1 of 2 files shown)


debpalash:
Thank you — both for the precise diagnosis in #1013 (down to the exact line in sesame.py where the empty context blows up) and for turning it into a clean PR with better test coverage than the fix that ended up merging.

Honest timeline, because you deserve it: your PR was opened at 17:42 UTC; I implemented the same two-line fix a couple of hours later (#1017, merged) working from your issue text — without checking the PR queue first, where this was already sitting. That's my mistake, and I'm sorry the duplication cost your PR the merge.

Two things so your work isn't wasted:
1. The merged fix IS your fix — same two lines, same guard, and the commit/CHANGELOG credit the diagnosis.
2. Your test file covers one case mine missed (the no-ref_audio, no-ref_text design path staying unaffected) — I'm absorbing that test into the suite with a `Co-authored-by` credit to you.

Closing as superseded by #1017. Your other PR (#1016) is under review right now — the entitlements diagnosis looks correct against the tauri-utils/wry sources, and it's likely to merge as-is.

Issue #567, opened 2026-06-20
[Bug] Can't reach the local OmniVoice backend — it may still be starting up, or it sto


picoin42520-hub:
## Error

```
Can't reach the local OmniVoice backend — it may still be starting up, or it stopped. Wait a few seconds and try again; if it persists, restart the app (or check Settings → Logs → Backend).
ApiError: Can't reach the local OmniVoice backend — it may still be starting up, or it stopped. Wait a few seconds and try again; if it persists, restart the app (or check Settings → Logs → Backend).
    at u (http://tauri.localhost/assets/client-CJSns_MO.js:1:1616)
    at async d (http://tauri.localhost/assets/client-CJSns_MO.js:1:2085)
    at async http://tauri.localhost/assets/main-app-BzqFQ7GD.js:16:23121
```

## Environment

**Version:** `0.3.7`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36 Edg/149.0.0.0`

## Recent actions

```
16:55:46 view:settings
17:02:37 view:donate
17:02:43 view:settings
17:02:44 view:donate
17:04:56 view:settings
17:07:37 view:projects
17:07:54 view:gallery
17:07:55 view:audiobook
17:07:59 view:stories
17:08:22 view:dub
17:08:42 dub:ingest-url
```

debpalash:
Thanks for the report — and you're already on the latest (**v0.3.7**), so this isn't a "please update" case.

One useful detail from the internals: `/dub/ingest-url` returns **immediately** (HTTP 202) and does the actual download in a background task (progress then streams over `/tasks/stream/…`). So a "Can't reach the local backend" at the moment you hit **Ingest** isn't a slow/stuck download — it means the backend process itself was unreachable right then (either it had stopped, or a transient WebView2 connection blip on Windows). Your breadcrumb fits that: the app navigated fine for ~13 min, and the failure landed on the first call that actually pokes the backend.

To tell a real crash apart from a one-off blip, the **backend log** is decisive:

1. **Settings → Logs → Backend** (or open `%LOCALAPPDATA%\OmniVoice\logs\backend.log`).
2. Copy the **last ~50 lines**, especially anything around the time you clicked Ingest (a traceback, an "exit", or a restart line).
[...]

PR #1031, opened 2026-07-09
fix(clone): bound the ref-text re-transcribe like every other ASR dispatch (#730)

stronghamjji:
Follow-up to #1008 (the merged fix for #1004), on your invitation to send anything my #1005 handled that the merged version doesn't.

#1008 correctly fixes the (ref_audio, ref_text) mismatch, and its exception/empty-transcript fallbacks are clean. The one case it doesn't cover is a re-transcribe that **wedges** (hangs) rather than raises — the #730 whisperx/CTranslate2 class.

### The gap

`refine_ref_texts` is dispatched with a bare `run_in_executor(_gpu_pool, ...)` at both call sites in `dub_core.py`, and `refine_ref_text` calls `asr_backend.transcribe()` directly. Its `try/except Exception` catches a *raised* error, but a hang isn't an exception, so that dispatch has no wall-clock bound. Every other transcribe in the same file already routes through `run_transcribe_guarded` — the chunk loop and the whole-file `what="Dub"` call — which bounds the call and `reset()`s the pool on timeout.

On a ≤10 GB card the GPU pool is 1 worker (the #567 sizing), so a wedged refine holds that worker and starves later GPU work into the misleading "Can't reach the local backend". There's also no `ping` on that `await`, so the EventSource drops.

### The fix

Route both refine dispatches (per-speaker and per-segment) through the existing `run_transcribe_guarded` — `ASRTimeoutError` and the guard are already imported in this file. On timeout it resets the pool and raises; keep the original clones, matching `refine_ref_text`'s own "failure is a strict no-op" fallback. No new machinery, just the convention the rest of the file already uses.

### Test

Adds a repro to `tests/test_speaker_clone_purity.py`:
- `refine_ref_texts` dispatched raw is unbounded on a hang (the gap), and
- through `run_transcribe_guarded` it times out and falls back to the original `ref_text` (the fix).

If `_gpu_pool` already bounds this somewhere I've missed, feel free to close — just wanted the wedge path covered the same way as the rest.

## Summary

This change bounds speaker-clone reference text refinement in the dub transcription stream so it uses the same guarded ASR path as other dispatches.

### What changed
- Routed both speaker-clone refinement paths in `backend/api/routers/dub_core.py` through `run_transcribe_guarded` instead of calling `asr_backend.transcribe()` indirectly via an unbounded executor path.
- On `ASRTimeoutError`, the flow now logs a warning and falls back to the original `ref_text` values, preserving the existing “failure is a strict no-op” behavior.
- Added a regression test in `tests/test_speaker_clone_purity.py` that reproduces the wedge scenario and verifies timeout handling plus fallback behavior.

### New behavior
```mermaid
flowchart TD
  A[Start ref-text refinement] --> B[Dispatch via run_transcribe_guarded]
  B --> C{ASR completes before timeout?}
  C -->|Yes| D[Use refined ref_texts]
  C -->|No| E[Raise ASRTimeoutError]
  E --> F[Log warning]
  F --> G[Keep original clone ref_texts]
```

### DIFF
--- a/backend/api/routers/dub_core.py
+++ b/backend/api/routers/dub_core.py
@@ -1008,9 +1008,22 @@ async def dub_transcribe_stream(
                     yield _sse_event("ping", {})
                 if clones:
                     from services.speaker_clone import refine_ref_texts
-                    clones = await loop.run_in_executor(
-                        _gpu_pool, lambda: refine_ref_texts(clones, _asr_backend),
-                    )
+                    # Bound the re-transcribe like every other ASR dispatch in
+                    # this file (#730): a wedged transcribe would otherwise hold
+                    # the GPU-pool worker forever and starve later work into a
+                    # "can't reach backend". On timeout the guard resets the pool
+                    # and raises — keep the original (unrefined) clones, matching
+                    # refine_ref_text's own "failure is a strict no-op" fallback.
+                    try:
+                        clones = await run_transcribe_guarded(
+                            _gpu_pool,
+                            lambda: refine_ref_texts(clones, _asr_backend),
+                            what="Dub clone ref-text refine",
+                        )
+                    except ASRTimeoutError as e:
+                        logger.warning(
+                            "clone ref-text refine timed out; keeping original ref_text: %s", e
+                        )
             # Wave 3.2: per-segment clone refs. Cut each long-enough segment's
             # own reference from the vocals so the dub of each line matches the
             # prosody of its source line. Short lines fall back to the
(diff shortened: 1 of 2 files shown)


debpalash:
Merged as-is — this is a model follow-up. You found a real gap in the merged #1008 (I bounded nothing on that dispatch; a #730-class wedge would have held the single GPU-pool worker forever and surfaced as yet another "can't reach backend"), fixed it in exactly the file's own convention, and the fail-before/pass-after wedge-test pair is textbook. Verified independently: patch applies clean on main, 19/19 in the purity suite, dub integration tests green.

Thank you for taking the invitation seriously — between #1004's diagnosis, #1005, and this, you've materially raised the dub pipeline's reliability. Ships in the next release.

PR #1088, opened 2026-07-11
feat(tts): streaming playback preview — audio starts on the first chunk while the rest renders

debpalash:
## What

Long scripts made users stare at a spinner until the **entire** audio was synthesized. This adds **streaming TTS playback**: audio starts playing as soon as the first chunk is ready, while the rest generates — as a preview channel layered on top of the existing generate path, not a rewrite of it.

## Backend — `POST /generate` + `stream=true`

- New optional `stream` form field on the existing route (default `false` → classic WAV response, byte-unchanged). With `stream=true` the response is `application/x-ndjson`: `start` → N × `chunk` (base64 PCM16 preview per text chunk, with the same effect chain the final take gets) → `done` (saved take metadata: id, path, duration, gen_time, seed) or in-band `error`.
- **Reuses the existing Wave 1.2 sentence-boundary chunking** (`split_text_into_chunks`) and per-chunk seeding — each chunk is one GPU-pool job whose raw output is yielded immediately, then concatenated with the same crossfade. **Engine-agnostic**: text-level chunking only, works with any `TTSBackend` (and the native OmniVoice path); no per-engine token streaming.
- **The saved file is untouched**: the finalize tail (watermark → save → history row → retention prune → event emit) was extracted verbatim into `_finalize_generation`, now shared by both paths — so the on-disk take is **byte-identical** to a non-streamed render (regression-tested, plus verified over real HTTP).
- `[pause]`-marker inputs and short single-chunk texts keep their unchanged single-shot pipeline and stream as one chunk.
- **Timeout budgets follow #1039**: engine warm-up runs under the model-load budget *before* the stream starts; each chunk then gets its own generate budget, so a long script can't time out merely for being long.
- All prep (engine resolution, routing gate #21, profile/seed/normalization/pronunciation) is 100% shared — a prep failure raises the same HTTP errors as today. `X-Seed` and routing-notice headers ride the streaming response too.

## Frontend

- When the preview would auto-play anyway (`autoPlayPreview` on) **and** Web Audio exists, `useTTS` drives the stream and the **global mini-player (#1042)** starts playback from the first received chunk: scheduled `AudioBufferSource` nodes with the backend's linear crossfade, live seek within the buffered region, progressive waveform peaks, growing duration, and a "Streaming preview…" label that flips to "Generated audio" on completion (i18n'd).
- **Platform parity (strict rule)**: plain `AudioContext` — present in WKWebView, WebView2, and WebKitGTK alike; no MediaSource (unreliable for audio/wav in WKWebView). If Web Audio is ever absent, the code takes the classic path — identical to today on every OS, so no opt-in toggle is needed.
- **Underruns** (synthesis slower than playback): the playhead waits at the buffered edge and resumes when the next chunk lands — no clipped audio.
- **Fallback**: any **mid-stream** failure (in-band `error`, transport drop, Web Audio failure) tears the preview down silently and re-runs the classic whole-file flow — the user sees nothing beyond the old wait. **Pre-stream** HTTP errors keep their `ApiError` identity, so real 400/503s surface exactly as before without a wasted second render.
- No new dependencies; no `package.json` change (root `bun.lock` verified with `bun install --frozen-lockfile`).

## Tests

- **Backend** (`tests/test_generate_streaming.py`): incremental delivery proven at the ASGI-message boundary with a per-chunk-delay mock engine (TestClient/httpx buffer whole bodies and can't observe it); final saved file **byte-identical** to the classic output; mid-stream engine failure → `error` event, no `done`, no history row, no file; single-chunk short text; classic default untouched.
- **Frontend** (`frontend/src/test/streamingTts.test.js`, 11 tests): NDJSON client → progressive scheduling, label flip, metadata resolution; `stream=true` appended without mutating the caller's FormData; error-event / no-done / mid-body-drop → `StreamingPreviewError` (fallback signal) vs pre-stream `ApiError` passthrough; chunk-player seek + stop transport.
- **Verified end-to-end over real HTTP** (uvicorn + fake engine): chunk arrival spread 1.0s across 3 chunks (truly incremental on the wire), saved file byte-identical to the classic take.
- Full frontend suite (1160 tests), touched-area backend suites (78 tests), `test_app_version.py` + `test_no_hardcoded_cjk.py` — all green; oxlint/tsc clean (no new warnings); vite build OK.

## Summary

Adds an optional streaming TTS preview path to `POST /generate` via `stream=true`, delivering NDJSON (`start` → base64 PCM16 `chunk`×N → `done`, with in-band `error`). Classic (non-streaming) generation behavior remains unchanged, including byte-identical saved WAV outputs. Mid-stream failures are prevented from creating incomplete history entries/files, and mid-stream streaming failures fall back silently to classic generation on the frontend.

### Backend
- Refactors the shared “successful generation tail” into `_finalize_generation`, which performs:
  - AudioSeal watermark embedding (executor)
  - WAV saving to `OUTPUTS_DIR`
  - Generation-history insertion with best-effort SQLite schema self-healing on `OperationalError`
  - Best-effort history retention pruning
  - `event_bus` emission
- Adds `stream: bool = Form(False)` to `@router.post("/generate")`.
- When `stream=true`:
  - Uses the existing chunking vs single-shot decision (including pause-marker parsing).
  - Streams NDJSON events:
    - `start`
    - per-text-chunk preview PCM16 audio as base64 via `chunk`
    - `done` once the full audio is assembled and finalized
    - `error` for validation/network/timeout/other failures that occur after streaming begins
  - Assembles the final audio (concat + effect chain) and then runs the exact same `_finalize_generation` path to persist the final WAV + history.
  - Client disconnects propagate without saving; in-band mid-stream `error` events avoid writing incomplete outputs/history.
- When `stream=false`:
  - Uses `_finalize_generation` and returns the same classic `audio/wav` semantics as before (including unchanged saved-file identity).

### Frontend
- Introduces `frontend/src/utils/streamingTts.js`:
  - `supportsStreamingPreview()` capability detection.
  - `streamGenerateSpeech(formData, ...)` that requests `stream=true` without mutating the caller’s `FormData`.
  - NDJSON parsing and progressive Web Audio playback (mini-player “output” track), including:
    - seeking/rescheduling within the buffered region
    - waveform/peaks updates
    - duration/progress updates
    - pause/resume via `AudioContext.suspend/resume`
    - teardown/cleanup on failure or stop
  - In-band streaming `error` (and other mid-body transport drops) become `StreamingPreviewError` to trigger fallback; pre-stream HTTP failures keep the original error identity.
- Updates `frontend/src/hooks/useTTS.js`:
  - Uses streaming preview only when `autoPlayPreview` is enabled and `supportsStreamingPreview()` is true.
  - Centralizes shared response-header handling via `applyResponseHeaders`:
    - updates UI seed from `X-Seed`
    - shows a one-time de-duped routing notice based on `X-OmniVoice-Routing`
  - Falls back to classic generation on `StreamingPreviewError`; pre-stream HTTP errors retain existing behavior.
- Adds `player.streaming_preview` i18n string (“Streaming preview…”).
- Clarifies mini-player retitling behavior in `frontend/src/utils/playback.js` so `update({ label })` retitles the active entry without merging `label` into the track payload.

#### Player flow

```mermaid
flowchart LR
  A[Generate with autoplay preview enabled] --> B{supportsStreamingPreview?}
  B -- No --> C[Classic whole-file generation]
  B -- Yes --> D[POST /generate with stream=true]
  D --> E{Outcome}
  E -- start/chunks/done --> F[Progressive mini-player playback + seek/waveform/progress updates]
  E -- Mid-stream in-band error/transport drop --> G[Throw StreamingPreviewError → fallback silently]
  E -- Pre-stream HTTP/validation error --> H[Preserve existing error behavior]
  F --> I[Finalize: persist WAV + history on backend; update status/label on completion]
  G --> C
```

### Validation & test reliability
- Backend/end-to-end tests for streaming preview validate:
  - NDJSON event ordering (`start` → incremental `chunk`s → `done`)
  - incremental delivery (first chunk arrives before later chunks finish rendering)
  - short inputs producing a single chunk while following the same persisted-history/file flow
  - mid-stream engine exception semantics: in-band `error`, no `done`, and no new history entry
  - native OmniVoice path parity: streamed saved WAV matches classic saved WAV
  - classic generation remains `audio/wav` with the expected `x-audio-id`
- Frontend unit tests cover:
  - PCM16 base64 decoding, waveform peak extraction, and streaming helpers
  - playback scheduling/teardown/seek behavior in Web Audio fakes
  - fallback wrapping rules (`StreamingPreviewError` only for mid-stream issues; pre-stream errors unchanged)
  - `streamGenerateSpeech` appends `stream=true` without mutating caller `FormData`
- Test stability fix: an autouse fixture pins both `OUTPUTS_DIR` bindings and the history database to per-test temporary paths, preventing saved-file lookups from diverging during full-suite runs and resolving the previously observed deterministic failures.
- Includes native OmniVoice coverage updates and an `oxfmt` line-length adjustment.

### DIFF
--- a/backend/api/routers/generation.py
+++ b/backend/api/routers/generation.py
@@ -625,6 +625,94 @@ def _persist_profile_ref_text(profile_id: str, ref_text: str) -> None:
         )
 
 
+async def _finalize_generation(
+    audio_tensor, sample_rate, *, text, history_mode, ref_audio_path,
+    language, instruct, resolved_profile_id, used_seed, start_time,
+):
+    """Shared tail of a successful generation: watermark → save WAV →
+    history row (self-healing) → retention prune → event emit.
+
+    Used verbatim by both the classic whole-file response path and the
+    streaming-preview path (``stream=true``), so the on-disk artifact —
+    watermark, filename, history row, retention behavior — is identical
+    regardless of how the audio was delivered to the client.
+
+    Returns ``(watermarked_tensor, meta)`` where ``meta`` carries
+    ``id`` / ``filename`` / ``duration`` / ``gen_time``.
+    """
+    loop = asyncio.get_running_loop()
+    # Invisible AudioSeal provenance watermark on the final audio. Embedding
+    # was previously only wired into the dub pipeline (dub_generate.py), so
+    # plain TTS came out unmarked despite the setting being on. embed_watermark
+    # self-gates on the user's watermark setting + AudioSeal availability and
+    # passes the audio through unchanged on any failure, so it never breaks
+    # generation.
+    from services.watermark import embed_watermark
+    audio_tensor = await loop.run_in_executor(
+        _gpu_pool, embed_watermark, audio_tensor, sample_rate
+    )
+    gen_time = round(time.time() - start_time, 2)
+
+    audio_id = str(uuid.uuid4())[:8]
+    audio_filename = f"{audio_id}.wav"
+    audio_path = os.path.join(OUTPUTS_DIR, audio_filename)
+    _safe_torchaudio_save(audio_path, audio_tensor, sample_rate)
+
+    audio_dur = round(audio_tensor.shape[-1] / sample_rate, 2)
+
+    # #710: the clip is already generated and saved above. A history-write
+    # failure — e.g. "no such table: generation_history" on a DB that missed
+    # schema init — must NOT 500 the user's generation. Self-heal the schema
+    # once and retry; if it still fails, log and return the audio anyway.
+    def _write_history():
+        with db_conn() as conn:
+            conn.execute(
+                "INSERT INTO generation_history (id, text, mode, language, instruct, profile_id, audio_path, duration_seconds, generation_time, seed, created_at) VALUES (?,?,?,?,?,?,?,?,?,?,?)",
+                (audio_id, text[:200], history_mode or ("clone" if ref_audio_path else "design"),
+                 language or "Auto", instruct or "", resolved_profile_id,
+                 audio_filename, audio_dur, gen_time, used_seed, time.time())
+            )
+    try:
+        _write_history()
+    except sqlite3.OperationalError as e:
+        logger.warning("generation history write failed (%s); healing schema + retrying", e)
+        try:
+            ensure_schema()
+            _write_history()
+        except Exception as e2:
+            logger.warning("history write still failed after schema heal; returning audio anyway: %s", e2)
+    except Exception as e:
+        logger.warning("generation history write failed; returning audio anyway: %s", e)
+    # Retention cap: without it, takes (rows + WAVs in OUTPUTS_DIR) grow
+    # unbounded forever. Best-effort — a prune failure must never affect
+    # the generation that just succeeded.
+    try:
+        _prune_history_over_cap()
+    except Exception as e:  # noqa: BLE001
+        logger.warning("history retention prune failed (non-fatal): %s", e)
+    event_bus.emit("generation_history", {"action": "created", "id": audio_id})
+
+    return audio_tensor, {
+        "id": audio_id,
+        "filename": audio_filename,
+        "duration": audio_dur,
+        "gen_time": gen_time,
+    }
+
+
+def _pcm16_b64(wav_tensor) -> str:
+    """Mono 16-bit little-endian PCM, base64-encoded — the streaming-preview
+    wire format (same conversion as /ws/tts). N-D tensors take channel 0."""
+    import base64
+
+    import torch
+
+    pcm = (wav_tensor * 32767).clamp(-32768, 32767).to(torch.int16)
+    while pcm.ndim > 1:
+        pcm = pcm[0]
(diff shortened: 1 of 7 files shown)


debpalash:
Rebased on current `main` (picks up #1094 + #1095) and fixed the red CI. Root cause: the three streaming tests read the saved take back through `core.config.OUTPUTS_DIR` while the router writes it through `api.routers.generation.OUTPUTS_DIR` — two separate module bindings that a full-suite run can split apart (an earlier test reloading `core.config`/`main` under a tmp data dir, e.g. `test_dub_transcribe`'s `app_client`, moves one and not the other), so the save landed in one dir and the read-back looked in another → `FileNotFoundError`. Added an autouse fixture pinning **both** `OUTPUTS_DIR` bindings + the history DB at per-test throwaway paths (same `ensure_schema.__globals__` seam the takes suite uses for the #909/#932 leak). Deterministic fail-before (divergent bindings pre-seeded) reproduced the exact 3 failures; pass-after green, including a 38-test interaction slice with the dub/takes suites.

PR #1154, opened 2026-07-14
fix(remote-auth): show an API-key gate (not PIN) for API-key 401s in remote-backend mode

paoloantinori:
## Problem

When `OMNIVOICE_API_KEY` is set (remote-backend mode — the `docs/remote-gpu.md` setup), a non-loopback browser gets `401 {"detail":"API key required"}` from `BearerKeyMiddleware`. But `client.ts` fired `ov:pin-required` on **every** 401, so `RemoteAuthGate` showed its **PIN** form — whose payload (`sessionStorage` `ov_pin` → `X-OmniVoice-Pin` / `?pin=`) can never satisfy the API-key middleware. A remote user was stuck on a PIN form they could not pass.

## Fix (frontend-only — backend unchanged)

- `client.ts` reads the 401 `detail` and dispatches a single `ov:auth-required` `CustomEvent` carrying `{ mode: 'apikey' | 'pin' }`. The two backend 401s are distinguishable only by their `detail` body (`"API key required"` vs `"PIN required"`); only two 401 sites exist backend-side.
- `RemoteAuthGate` renders the matching PIN or API-key form, storing `ov_pin` (session) or `ov_api_key` (durable localStorage).
- `?api_key=<key>` deep-link bootstrap (mirrors the existing `?pin=` QR flow), **scrubbed from the URL after capture** so a reload can't re-clobber a corrected key (and the secret doesn't linger in history).
- Guarded `saveApiKey` helper shared by the gate and the bootstrap.

## Docs

`docs/remote-gpu.md` already promised *"drive it from the desktop app or a browser"* but only documented the desktop-app Settings flow; added a **"From a browser"** subsection covering the `?api_key=` deep link.

## Testing

- `RemoteAuthGate.test.jsx`: API-key-mode submit writes `ov_api_key`.
- `client.test.ts`: 401 routing dispatches `ov:auth-required` with the correct `mode` for both `"API key required"` and `"PIN required"`.
- All frontend gates green: `bun test` (162 files / 1256 tests), `typecheck:ci`, `lint`, `format:check`, and the i18n probe (new `remote_apikey_gate` keys added to all 21 locales). Backend untouched.

## Notes / follow-up

The routing discriminates on the `detail` prose, which the backend owns. A more robust follow-up would have the backend emit a structured `code` on both 401s (`{"detail":"…","code":"api_key_required"}`) and have the frontend switch on `code` rather than substring-matching a sentence — intentionally out of scope for this frontend-only fix.

## Summary

- Updated remote authentication so `401` routing chooses the correct credential form:
  - `"API key required"` → opens the API-key gate
  - `"PIN required"` → opens the PIN gate
- `frontend/src/api/client.ts` now dispatches a single `ov:auth-required` event on `401` with `detail.mode`:
  - `"apikey"` only when the parsed 401 `detail` is a **string** whose lowercase content includes `"api key"`
  - otherwise `"pin"`
- Added durable API-key persistence:
  - New exported `saveApiKey(v: string): boolean` trims and stores in `localStorage` under `ov_api_key` (returns `false` on empty/trimmed value or storage failure)
  - PIN persistence remains in `sessionStorage` under `ov_pin`
- Browser deep-link bootstrap refactor:
  - PIN still comes from query `?pin=<pin>`
  - API key now comes from URL fragment `#api_key=<key>` (so it isn’t sent to server/proxy logs)
  - Consumed credentials are scrubbed from the address bar via `history.replaceState` using a computed `cleanUrl`
  - Legacy query-string `?api_key=` is scrubbed from `cleanUrl` but is **not** read/used
  - Implemented via pure helper `_parseDeepLinkCredentials(href)` (unit-tested parsing + scrubbing)
- Expanded `frontend/src/components/RemoteAuthGate.jsx` from PIN-only to dual-mode:
  - Listens for `ov:auth-required` and switches between `remote_apikey_gate` and `remote_gate` based on `e.detail.mode` (defaulting to PIN)
  - Persists based on mode (`localStorage` for API key, `sessionStorage` for PIN) and avoids infinite reload when storage fails
- Documented browser-based remote API-key authentication in `docs/remote-gpu.md`.
- Added/updated frontend tests for `apiFetch` 401 routing/event dispatch and deep-link credential parsing/scrubbing.
- Added new i18n strings under `remote_apikey_gate` (English content pending community translation).

## UI

```text
Before:                         After:
[ PIN required ]                [ PIN required ]  /  [ API key required ]
[ PIN input   ]                [ PIN input   ]      [ API-key input ]
[ Connect     ]                [ Connect      ]     [ Connect      ]
```

## Authentication flow

```mermaid
flowchart TD
  A[Remote request] --> B{HTTP status 401?}
  B -- No --> C[Normal error handling]
  B -- Yes --> D[Read 401 body (detail/error text)]
  D --> E{detail is string && contains "api key" (case-insensitive)?}
  E -- Yes --> F[dispatch ov:auth-required {mode:"apikey"}]
  E -- No --> G[dispatch ov:auth-required {mode:"pin"}]
  F --> H[RemoteAuthGate shows API-key form]
  G --> I[RemoteAuthGate shows PIN form]
  H --> J[Persist API key: saveApiKey() → localStorage (ov_api_key)]
  I --> K[Persist PIN: sessionStorage (ov_pin)]
  J --> L[Retry gated request]
  K --> L

  subgraph Browser bootstrap (on load)
    M[Load URL] --> N[_parseDeepLinkCredentials(window.location.href)]
    N --> O[Extract ?pin= (query) and `#api_key`= (fragment)]
    O --> P[Scrub consumed creds from address bar (replaceState → cleanUrl)]
    P --> Q[Persist: pin → sessionStorage; apiKey → saveApiKey/localStorage]
  end
```

## Validation

Frontend tests, type checking, linting, formatting, and i18n checks pass; backend behavior remains unchanged.

### DIFF
--- a/docs/remote-gpu.md
+++ b/docs/remote-gpu.md
@@ -83,6 +83,22 @@ Settings → Sharing → **Remote backend**:
 
 Leave the URL empty to go back to the local backend.
 
+### From a browser (no desktop app)
+
+You can also drive the remote from a plain browser — open the URL with the key
+in the query string once:
+
+```
+https://gpu-box.your-tailnet.ts.net/?api_key=<key>
+```
+
+The key is stored for that browser and the `?api_key=` param is scrubbed from
+the address bar (so it doesn't linger in history or get re-applied on a reload).
+Thereafter the UI loads normally with the key attached to every request. If a
+request ever 401s again (wrong/rotated key), you're prompted to re-enter it. The
+same gate shows a LAN-share **PIN** prompt instead when network sharing — not a
+remote key — is what's gating access.
+
 ## Security notes
 
 - **Plain HTTP is sniffable.** A bearer key over `http://` on a hostile
--- a/docs/remote-gpu.md
+++ b/docs/remote-gpu.md
@@ -86,14 +86,18 @@ Leave the URL empty to go back to the local backend.
 ### From a browser (no desktop app)
 
 You can also drive the remote from a plain browser — open the URL with the key
-in the query string once:
+in the **fragment** once:
 
 ```
-https://gpu-box.your-tailnet.ts.net/?api_key=<key>
+https://gpu-box.your-tailnet.ts.net/#api_key=<key>
 ```
 
-The key is stored for that browser and the `?api_key=` param is scrubbed from
-the address bar (so it doesn't linger in history or get re-applied on a reload).
+Use the fragment (`#`, not `?`) deliberately: fragments are never sent to the
+server, so the key stays out of the GPU box's and any reverse proxy's request
+logs. The key is stored for that browser and the fragment is scrubbed from the
+address bar (so it doesn't linger in history or get re-applied on a reload). If
+your key contains `+`, `&`, `#`, or `=`, URL-encode it (e.g. `#api_key=a%2Bb`);
+keys from `secrets.token_urlsafe` (above) need no encoding.
 Thereafter the UI loads normally with the key attached to every request. If a
 request ever 401s again (wrong/rotated key), you're prompted to re-enter it. The
 same gate shows a LAN-share **PIN** prompt instead when network sharing — not a
--- a/frontend/src/api/client.test.ts
+++ b/frontend/src/api/client.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { _parseDeepLinkCredentials } from './client';
 
 describe('apiFetch PIN header', () => {
   let realFetch: typeof globalThis.fetch;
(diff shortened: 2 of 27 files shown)


paoloantinori:
Thanks for the automated reviews — triaged the findings. Pushed one fix (7f4b24a); reasoning on the rest below.

## Fixed
- **URL fragment dropped when scrubbing `?api_key=`/`?pin=`** *(greptile P2, coderabbit)*: the `replaceState` rebuilt the URL without `url.hash`, nuking any `#fragment`. Now `pathname + (?query) + hash`.

## Noted — not changing here, with reasons

- **`?api_key=` reaches request logs before the JS scrub** *(greptile P1)*: `?api_key=` is this app's established browser-auth convention — `BearerKeyMiddleware` accepts it, `wsUrl()` appends it (browser WebSockets can't set headers), and `docs/remote-gpu.md` documents it. The `ov_key` cookie takes over after the first authenticated request, so the key is in a URL only on that one bootstrap hit. A `#api_key=` fragment bootstrap would remove the server-log surface for the page load — happy to add it (or support both) if you want that hardening; I didn't switch unilaterally since it diverges from the documented `?api_key=` UX and the WS path still needs the query param.

- **Plaintext `localStorage` for the key** *(CodeQL, coderabbit, greptile)*: this is the pre-existing storage model — `RemoteBackendPanel.jsx` already writes `ov_api_key` to localStorage, and `_apiKey()`/`wsUrl()` read it. This PR reuses that exact mechanism via one shared `saveApiKey` writer. Migrating the remote-backend credential to an encrypted vault / OS keystore / httponly cookie is a separate project-wide change to the remote-auth model — out of scope for this fix, and yours to decide. (The CodeQL alert fires on the new line but is flagging the established pattern.)

- **Translate the 21 locales** *(coderabbit)*: the `remote_apikey_gate` strings ship as English in non-EN locales by design — the project's practice for new keys pending community translation (i18n probe passes, coverage 100%). I didn't auto-generate 20 translations (error-prone without native review). Happy to add them or leave for the community.

- **Visible error message on persistence failure** *(coderabbit)*: the gate already avoids the loop/freeze — it skips the reload when `saveApiKey` returns false or `sessionStorage.setItem` throws, so the form stays put. A toast would add a new i18n key across 21 locales for a rare privacy-mode / storage-failure edge; noted as a follow-up, can add if wanted.

- **PR title needs an issue ref** *(coderabbit pre-merge check)*: I don't have a tracking issue for this bug — happy to file one and reference it, or leave it to you.

No backend changes; all frontend gates green (1256 tests, `typecheck:ci`, `lint`, `format:check`, i18n probe).

paoloantinori:
Follow-up on the 17:03 re-review: addressed the one new finding.

- **`String(detail)` can throw on a malformed 401 detail** *(coderabbit, Stability)*: switched the mode discriminator to a `typeof detail === 'string'` guard so a non-string / broken-`toString` detail falls back to PIN instead of throwing before the event dispatches. (commit on the way)

The earlier findings stand as triaged in my previous comment — notably the **CodeQL failure is the established `localStorage` API-key pattern** (`RemoteBackendPanel` already stores `ov_api_key` there; this PR only unifies the write under `saveApiKey`). Clearing it needs a maintainer dismissal of the code-scanning alert, which I can't do from a fork.

paoloantinori:
Addressed greptile's P1 (API key reaching request logs): the remote-backend deep link now uses the **fragment** — `https://host/#api_key=<key>` — instead of `?api_key=`. Fragments are never sent to the server, so the durable key no longer reaches the GPU box's / reverse-proxy's request logs on the page load. `?pin=` stays on the query (QR flow, session PIN). (b06baca)

Implementation notes:
- Extracted the bootstrap into a pure, unit-tested `_parseDeepLinkCredentials` helper (pin from query, api_key from fragment, one-shot scrub of both).
- A stray legacy `?api_key=` in the query is scrubbed from the URL *without* being read (reading it would resend the secret on reload — the leak this avoids).
- Docs (`remote-gpu.md`) document `#api_key=` with encoding guidance for keys containing `+`/`&`/`#`/`=`.

One residual, honestly disclosed in a code comment: the dictation **WebSocket** still appends `?api_key=` (`wsUrl` — browsers can't set WS headers), so the key still reaches logs on the WS handshake. Fully closing that needs a backend protocol change (first-frame auth or a short-lived ticket) and is tracked as a separate, deeper follow-up — out of scope for this frontend-only PR.

debpalash:
Merged — thanks for a genuinely excellent PR: the fragment-based deep link, the legacy-`?api_key=` scrub-without-read, and the triage notes on each review finding made this easy to take. The CodeQL `js/clear-text-storage-of-sensitive-data` alert was dismissed as the established `RemoteBackendPanel` storage pattern, exactly as you argued; moving the remote credential to an OS keystore (and first-frame auth for the dictation WebSocket) stay tracked as separate follow-ups. The structured-`code`-on-401 backend follow-up you suggested is noted as well.

PR #1161, opened 2026-07-16
fix: return empty text instead of partial garbage when EPUB HTML parse fails

bultodepapas:
`_html_to_title_body` used `except Exception: pass` when the HTMLParser raised.

### Problem
The parser accumulates state incrementally across `handle_starttag`, `handle_endtag`, and `handle_data` callbacks. A mid-parse failure returns whatever partial or corrupted `title` and `_parts` were accumulated before the error — with zero indication anything went wrong.

### Impact
An EPUB with a single corrupt XHTML entry produces silently wrong chapter text. The user sees incorrect audiobook chapter narration with no error feedback.

### Fix
- Returns `("", "")` on parse failure instead of partial state
- Logs the failure with `logger.warning(exc_info=True)` so the traceback is preserved

## Summary

- Preserves partially extracted EPUB XHTML text when `_html_to_title_body` encounters a mid-chapter `HTMLParser` exception, instead of discarding the whole chapter by returning empty title/body.
- Emits a module-scoped warning with traceback (`logger.warning(..., exc_info=True)`) on HTML parsing failure, improving diagnosability while still returning `p.title` and `p.text()` accumulated before the failure.
- Adds regression tests ensuring:
  - A clean EPUB imports all chapters.
  - A simulated mid-chapter parse failure keeps Chapter 2 with pre-failure text and logs an “HTML parsing failed” warning.
  - If parsing fails before any text is collected, that chapter remains skippable while later chapters still import.

```mermaid
flowchart TD
    A[Parse EPUB XHTML] --> B{HTML parsing succeeds?}
    B -->|Yes| C[Return collected title and body]
    B -->|No| D[Log warning with traceback]
    D --> E[Return title/body collected before failure]
    E --> F{body has text?}
    F -->|Yes| G[Include chapter in audiobook]
    F -->|No| H[Caller skips chapter]
```

### DIFF
--- a/backend/services/longform_import.py
+++ b/backend/services/longform_import.py
@@ -12,6 +12,7 @@ just a front door onto the existing pipeline.
 from __future__ import annotations
 
 import io
+import logging
 import posixpath
 import re
 import zipfile
@@ -29,6 +30,8 @@ _CHAPTER_TITLE_MAX = 60
 # *uncompressed* bytes read from the archive.
 _EPUB_MAX_ENTRY_BYTES = 25 * 1024 * 1024
 _EPUB_MAX_TOTAL_BYTES = 300 * 1024 * 1024
+
+logger = logging.getLogger("omnivoice.longform_import")
 
 
 def chapterize_plaintext(text: str) -> str:
@@ -108,7 +111,8 @@ def _html_to_title_body(xhtml: str) -> tuple[str, str]:
     try:
         p.feed(xhtml)
     except Exception:
-        pass
+        logger.warning("HTML parsing failed for EPUB entry", exc_info=True)
+        return "", ""
     return p.title, p.text()
 
 
--- a/backend/services/longform_import.py
+++ b/backend/services/longform_import.py
@@ -111,8 +111,11 @@ def _html_to_title_body(xhtml: str) -> tuple[str, str]:
     try:
         p.feed(xhtml)
     except Exception:
-        logger.warning("HTML parsing failed for EPUB entry", exc_info=True)
-        return "", ""
+        # Keep whatever the extractor collected before the failure: an empty
+        # return would make the caller's `if not body.strip(): continue` drop
+        # the whole chapter from the audiobook silently — a partial chapter
+        # plus this log line is strictly more recoverable than a missing one.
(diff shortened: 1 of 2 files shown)


backend/services/longform_import.py
```diff
@@ -108,7 +111,8 @@ def _html_to_title_body(xhtml: str) -> tuple[str, str]:
     try:
         p.feed(xhtml)
     except Exception:
-        pass
+        logger.warning("HTML parsing failed for EPUB entry", exc_info=True)
+        return "", ""
```
debpalash:
```suggestion
        logger.warning("HTML parsing failed for EPUB entry; using partial text", exc_info=True)
```

debpalash (changes requested):
The `logger.warning(..., exc_info=True)` addition is genuinely good — that silent `pass` deserved to go. But the `return "", ""` swaps one silent failure for another: an empty body hits `if not body.strip(): continue` at line 185, so the chapter is silently dropped from the audiobook — which is exactly the "silently wrong narration" this PR sets out to fix, just in a different shape. The partial text the extractor accumulated before the parse error is usually most of the chapter, and it's strictly better for the user than a missing one.

Suggestion below keeps your logging and the partial text. (The alternative precedent in this file is the PDF path: raise so the route returns an actionable 400 — also acceptable if you'd rather fail loud. Either way, not silent.)

debpalash:
Ran this through a real 3-chapter EPUB with a mid-chapter parse failure injected into chapter 2 (`html.parser` itself refused to raise on 10/10 hostile inputs — unclosed scripts, bad charrefs, null bytes, 10k-deep nesting, truncated entities — so the except branch only fires in genuinely exotic cases):

| | chapters in output | chapter 2 content |
|---|---|---|
| **current main** (`except: pass`) | **3** | partial text kept (everything parsed before the failure) |
| **this PR** (`return "", ""`) | **2** | **silently gone** — empty body hits the `continue` at line 185 |

So the `logger.warning(..., exc_info=True)` is a clear win — keep it. But the empty return converts "possibly-truncated chapter" into "silently missing chapter", which is the harder failure to notice in a finished audiobook. Apply the one suggestion (log + keep partial text) and this is mergeable.

PR #1162, opened 2026-07-16
fix: prevent probeAudioDuration from hanging on unresolved Promise

bultodepapas:
The original Promise constructor only accepted `resolve` — no `reject` callback. The `error` event handler called `resolve(null)` instead of rejecting.

### Problem
If the `Audio` element never emits `loadedmetadata` or `error` (rare browser conditions, garbage collection races, invalid blob URLs), the Promise hangs forever with no settlement path. Any caller `await`-ing this function stalls indefinitely.

### Fix
- Added 10-second timeout that rejects if neither event fires
- Proper `reject()` on error event instead of `resolve(null)`
- `{ once: true }` on event listeners to prevent double-invocation
- Dedicated cleanup function for `URL.revokeObjectURL`

## Summary

- Fixes `probeAudioDuration` hanging indefinitely by adding a 10-second timeout.
- Ensures the function **never rejects**: it resolves with either a finite duration or `null` on both media `error` (e.g., undecodable codec) and on timeout.
- Uses one-time event listeners (`{ once: true }`) plus shared cleanup to `URL.revokeObjectURL`, so settling happens exactly once and late events after timeout don’t cause double-cleanup or errors.
- Adds regression tests with a `FakeAudio` stub to cover: successful `loadedmetadata`, `error` (never rejects, resolves `null`), silent media (resolves `null` after timeout), and a late `loadedmetadata` arriving after timeout.

```mermaid
flowchart TD
  A[Create Audio and object URL] --> B[Register one-time listeners + start 10s timeout]
  B --> C{Event or timeout occurs first?}
  C -->|loadedmetadata| D[Clear timeout + revoke URL + resolve duration (finite)]
  C -->|error| E[Clear timeout + revoke URL + resolve null]
  C -->|10-second timeout| F[Revoke URL + resolve null]
  F --> G[Late loadedmetadata (after timeout) is ignored / no double-revoke]
```

### DIFF
--- a/frontend/src/utils/format.js
+++ b/frontend/src/utils/format.js
@@ -4,14 +4,19 @@ export function formatTime(s) {
   return `${m}:${sec.padStart(4, '0')}`;
 }
 
+// Contract: settles with a number or null, NEVER rejects. null means
+// "duration unknown — keep the file": callers (ingestRefAudio) have no
+// try/catch and must still accept clips this webview can't decode, because
+// the backend decodes them with ffmpeg (Tauri WebKit lacks several codecs).
+// The timeout guards against media elements that never fire any event.
 export async function probeAudioDuration(file) {
-  return new Promise((resolve, reject) => {
+  return new Promise((resolve) => {
     const url = URL.createObjectURL(file);
     const a = new Audio();
     const cleanup = () => URL.revokeObjectURL(url);
     const timeout = setTimeout(() => {
       cleanup();
-      reject(new Error('probeAudioDuration timeout'));
+      resolve(null);
     }, 10000);
     a.addEventListener('loadedmetadata', () => {
       clearTimeout(timeout);
@@ -21,7 +26,7 @@ export async function probeAudioDuration(file) {
     a.addEventListener('error', () => {
       clearTimeout(timeout);
       cleanup();
-      reject(new Error('probeAudioDuration failed to load'));
+      resolve(null);
     }, { once: true });
     a.src = url;
   });
--- a/frontend/src/utils/format.js
+++ b/frontend/src/utils/format.js
@@ -18,16 +18,24 @@ export async function probeAudioDuration(file) {
       cleanup();
       resolve(null);
     }, 10000);
-    a.addEventListener('loadedmetadata', () => {
-      clearTimeout(timeout);
-      cleanup();
-      resolve(isFinite(a.duration) ? a.duration : null);
-    }, { once: true });
-    a.addEventListener('error', () => {
-      clearTimeout(timeout);
-      cleanup();
-      resolve(null);
-    }, { once: true });
+    a.addEventListener(
+      'loadedmetadata',
+      () => {
+        clearTimeout(timeout);
+        cleanup();
+        resolve(isFinite(a.duration) ? a.duration : null);
+      },
+      { once: true },
+    );
+    a.addEventListener(
+      'error',
+      () => {
+        clearTimeout(timeout);
+        cleanup();
+        resolve(null);
+      },
+      { once: true },
+    );
     a.src = url;
   });
 }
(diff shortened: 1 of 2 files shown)


frontend/src/utils/format.js
```diff
@@ -5,15 +5,24 @@ export function formatTime(s) {
 }
 
 export async function probeAudioDuration(file) {
-  return new Promise((resolve) => {
+  return new Promise((resolve, reject) => {
```
debpalash: `reject` is no longer used once the failure paths resolve:
```suggestion
  return new Promise((resolve) => {
```

frontend/src/utils/format.js
```diff
@@ -5,15 +5,24 @@ export function formatTime(s) {
...
-      URL.revokeObjectURL(url);
-      resolve(v);
-    };
-    a.addEventListener('loadedmetadata', () => done(isFinite(a.duration) ? a.duration : null));
-    a.addEventListener('error', () => done(null));
+    const cleanup = () => URL.revokeObjectURL(url);
+    const timeout = setTimeout(() => {
+      cleanup();
+      reject(new Error('probeAudioDuration timeout'));
+    }, 10000);
```
debpalash, line 20: Timeout should degrade the same way a decode failure always has — `null` means "duration unknown", and the caller treats it as such:
```suggestion
    const timeout = setTimeout(() => {
      cleanup();
      resolve(null);
    }, 10000);
```

frontend/src/utils/format.js
```diff
@@ -5,15 +5,24 @@ export function formatTime(s) {
...
+    a.addEventListener('loadedmetadata', () => {
+      clearTimeout(timeout);
+      cleanup();
+      resolve(isFinite(a.duration) ? a.duration : null);
+    }, { once: true });
+    a.addEventListener('error', () => {
+      clearTimeout(timeout);
+      cleanup();
+      reject(new Error('probeAudioDuration failed to load'));
+    }, { once: true });
```
debpalash: Same here — keep the original `resolve(null)` contract so `ingestRefAudio` keeps accepting files the media element can't decode:
```suggestion
    a.addEventListener('error', () => {
      clearTimeout(timeout);
      cleanup();
      resolve(null);
    }, { once: true });
```

debpalash (changes requested):
The timeout + `{once: true}` hardening is welcome — but switching the error path from `resolve(null)` to `reject()` breaks the function's only caller. `ingestRefAudio` (`frontend/src/hooks/useTTS.js:70`) awaits this with no try/catch: today a file whose metadata the media element can't decode probes to `null` and is still accepted (the backend decodes it with ffmpeg — this matters under Tauri WebKit, where several codecs won't decode in `<audio>` but dub/clone fine). After this change the same file becomes an unhandled rejection and the reference audio silently never gets set — a cross-platform regression.

The suggestions below keep your timeout and listener cleanup but settle with `resolve(null)` on both failure paths, preserving the caller contract. With those applied (and ideally a small test asserting the timeout resolves `null`), happy to merge.

debpalash:
Ran a live behavioral test of all three variants (current main, this PR, and the suggested fix) through a faithful replica of the real caller (`ingestRefAudio`, which has no try/catch), under both failure modes:

| Scenario | current main | this PR | with suggestions |
|---|---|---|---|
| **Undecodable codec** (media element fires `error` — what Tauri WebKit does for several codecs) | file accepted ✅ | `ingestRefAudio` **throws**, reference **never set** ❌ | file accepted ✅ |
| **Metadata never arrives** (the hang this PR targets) | **hangs forever** ❌ | unblocks after timeout, but throws into the uncatching caller ❌ | resolves `null`, file accepted ✅ |

Two takeaways: **the hang you found is real** — current main genuinely never settles when no event fires, so the timeout is a good fix and we want it. And the `reject()` contract change is also really a regression — the error path breaks a file-ingest flow that works today. Apply the three suggestions (timeout stays, both failure paths `resolve(null)`) and this is mergeable.

PR #1163, opened 2026-07-16
fix: prevent healthy WebSocket queues from being killed by race condition

bultodepapas:
When `put_nowait` raised `QueueFull`, the recovery handler called `get_nowait` to free a slot and retried the put.

### The race
Consumers drain queues asynchronously **without holding `_lock`**. If a consumer drains ALL items between the `QueueFull` from `put_nowait` and the `get_nowait` call, `get_nowait()` raises `asyncio.QueueEmpty`. The bare `except Exception` treated this as a fatal queue error and removed the queue from `_listeners` — permanently disconnecting that WebSocket client.

### Impact
Under high event throughput, a WebSocket client whose reader happens to drain the queue in that narrow window is silently disconnected from the event bus. The frontend sidebar freezes with no recovery mechanism.

### Fix
Explicitly catches `asyncio.QueueEmpty` and retries `put_nowait` (the queue now has room), only marking the queue dead on a subsequent `QueueFull` or other unexpected exception.

## Summary

Fixes a race condition that could incorrectly remove healthy WebSocket queues from `_listeners`.

When recovering from a full queue, `_broadcast` now handles `asyncio.QueueEmpty` if another coroutine drains the queue between `put_nowait` and `get_nowait`. It retries inserting the event and only treats subsequent `QueueFull` or unexpected exceptions as listener failures.

```mermaid
flowchart TD
    A[Queue is full] --> B[Drop oldest item]
    B -->|QueueEmpty race| C[Retry put_nowait]
    B -->|Success| C
    C -->|Success| D[Keep listener]
    C -->|QueueFull or unexpected error| E[Mark listener failed]
```

### DIFF
--- a/backend/core/event_bus.py
+++ b/backend/core/event_bus.py
@@ -76,6 +76,13 @@ async def _broadcast(event_str: str) -> None:
                 try:
                     q.get_nowait()
                     q.put_nowait(event_str)
+                except asyncio.QueueEmpty:
+                    # Consumer drained the queue between our put_nowait
+                    # and get_nowait — queue has room; retry the put.
+                    try:
+                        q.put_nowait(event_str)
+                    except asyncio.QueueFull:
+                        dead.append(q)
                 except Exception:
                     dead.append(q)
         for q in dead:


debpalash:
Thanks for the PR — declining this one after deep review (three independent concurrency analyses, all reaching the same conclusion): the race this fixes cannot occur.

The claimed window — a consumer draining the queue between `put_nowait` raising `QueueFull` and the recovery `get_nowait` — requires a task switch inside that span. But the span contains **no `await`**: `put_nowait`, the `except`, and `get_nowait` are straight-line synchronous code, and asyncio is cooperatively scheduled — a coroutine only yields at an await point. The consumer's `await q.get()` (`events.py:41`) cannot run a single instruction in between. `QueueFull` proves the queue holds exactly 64 items; one synchronous line later it still does, so `get_nowait()` cannot raise `QueueEmpty` — the new handler is unreachable.

The cross-thread escape hatch doesn't exist either: if `emit()` is ever called off the loop thread, `asyncio.get_running_loop()` raises and the event is dropped *before touching any queue* (`event_bus.py:62-64`), so all queue ops are confined to one loop thread (plus `_broadcast` holds `_lock`).

One subtle downside of the patch as written: an unexpected exception inside the new nested branch would escape `_broadcast` entirely, where today the outer `except Exception` contains it — so it's a small exception-safety regression guarding an impossible case.

We're adding a comment to `event_bus.py` documenting this invariant so the pattern doesn't trip up the next reader. If you're seeing *real* WebSocket disconnects in the wild, that's worth a fresh issue with the actual exception hitting `dead.append(q)` — the mechanism just isn't this one.

(Your #1159 and #1160 were solid and are merged — thanks for those.)

PR #1208, opened 2026-07-20
feat(audiobook): expressive maturity — overrides, emotion, cache opt-out (#1208)

debpalash:
Audiobook renders were locked to the model's most deterministic preset (32 steps / 2.0 guidance / model-default temps) with no way to change it — which is why books sounded flatter than the same voice on the Voice page. This opens expressive control without changing any default byte-for-byte.

### Added
- **Production Overrides in the Audiobook tab** — `position_temperature`, `class_temperature`, `num_step`, `guidance_scale`, `postprocess_output` (+ seed), reusing the Voice page's panel. Unset reproduces today exactly.
- **IndexTTS2 graded emotion** (`emo_vector` / `emo_text` / `emo_alpha`) reaches the longform path via a typed engine-options object; engines that don't understand an option ignore it (verified across the ~14 backends — no crash).
- **Cache opt-out** ("vary repeated lines") so identical lines can get distinct takes; default off keeps the content-addressed replay.

### Fixed
- Reaction tags (`[laughter]`, `[sigh]`, …) that already work in audiobooks are now listed in the tab's Markup reference.
- `AudiobookGenerateBody` dropped `language`, so audiobook language selection never reached the backend — now wired.

### Docs
- `docs/expressive-speech.md` corrected so no recipe it names is unreachable in the Audiobook surface.

**Correctness:** every new param is folded into **both** cache layers (chapter + segment) and the per-chapter preview, so changing a knob re-renders instead of replaying stale audio — and an all-default request keeps its old cache key, so **existing books don't re-render**. Regression tests: `tests/test_audiobook_expressive.py` (backward-compat, cache-signature loop, preview/render parity, engine-ignores-unknown, cache opt-out, emotion-reaches-engine) + `audiobookOverrides.test.jsx`.

Adds audiobook production overrides, language forwarding, IndexTTS2 emotion controls, repeat-line variation, and expressive-aware preview/render/resume caching while preserving default behavior and cache keys. Updates the Audiobook UI, persistence, API types, documentation, translations, and regression coverage. Human review should focus on cache invalidation/resume compatibility and backend-specific option propagation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 - Info/warn system notifications are dismissible and stay dismissed across restarts; error-level notices can't be dismissed, and the unclean-shutdown notice is now acknowledged server-side — thanks @agudmund! (#1192)
 - `clone_voice` MCP tool — AI agents can clone a new voice from a base64 reference audio sample; returns a `profile_id` immediately usable with `generate_speech` — thanks @paoloantinori! (#1194)
 - Dub tab: **Paste Translation** — paste a translation made elsewhere (ChatGPT, DeepL, a human) as subtitles, numbered lines, or plain lines; it maps onto the existing segments with a before→after preview, keeping timings and the source transcript intact (#1203)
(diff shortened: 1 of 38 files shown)


debpalash:
Merged into main via the v0.3.23 integration batch (commit d7608b1b is on main as of cbbddfce). Closing — the audiobook expressive-maturity work + review fixes are shipped.

PR #1210, opened 2026-07-20
fix(clone): AudioTrimmer preview plays the selected region on VBR clips

debpalash:
The trimmer's waveform, selection, and exported clip all live on the decoded-AudioBuffer timeline, but **preview** played the original file through an `<audio>` element seeked with *buffer-seconds* as if they were *container-seconds*. On any clip whose container duration diverges from the decoded length (VBR MP3, estimated/mis-reported durations), the preview drifted off the region the user selected — and off the region that actually gets exported for cloning.

### Fixed
- Preview now plays the **same decoded `AudioBuffer`** the waveform is drawn from and the export slices, over the exact `[start, end]` window (Web Audio `BufferSource`, native `loopStart`/`loopEnd`). One timeline — preview, waveform, and exported clip can't diverge. Also removes per-container media-decoding differences across the three platform WebViews.

Regression tests (fail-before/pass-after): `AudioTrimmerPreview.test.jsx` + `audioTrimSelectionPlayhead.test.js`.

AudioTrimmer preview playback was reworked to use the decoded `AudioBuffer` with a Web Audio `BufferSource` (via `loopWindow`) and to drive the playhead from an `AudioContext` clock, so the preview loops and stop behavior match the waveform/export timeline for the exact selected `[start, end]` region. This fixes prior drift caused by VBR and misreported-duration clips by ensuring preview, waveform, and exported audio all read from the same decoded buffer. Risk to human review: `AudioContext`/source lifecycle and correct loop/playhead retargeting when the user edits the selection during active playback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 
 ### Fixed
 
+- Voice-clone trimmer: the preview now plays exactly the selected region on variable-bitrate clips (it had drifted off on VBR/mis-reported-duration files by playing the original file on a different timeline) (#1210)
 - A backend that fails to start now says why — exit code and error output, with actionable hints and a one-click report — instead of the evidence-free "Can't reach the local OmniVoice backend" (#1177)
 - Generation no longer crawls on CPU after a cancelled or failed dub: the TTS model is moved back to the GPU on every exit path, and each generation now verifies its own placement (#1191)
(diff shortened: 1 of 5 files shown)


debpalash:
Fixed in 442125da. Extracted a pure `loopWindow(start, end, duration)` helper that floors the segment at `MIN_LOOP_SEC` (so `loopStart < loopEnd` always) and clamps into the buffer — an empty/inverted selection can no longer collapse the range into the whole-buffer fallback. Unit-tested for empty/inverted/near-end cases in `audioTrimSelectionPlayhead.test.js`. Good catch.

PR #1216, opened 2026-07-20
feat(audiobook): stop/cancel generation + live per-chapter progress (#1216)

debpalash:
The **Create audiobook** button had no way to stop, and even breaking the read loop left the fetch open — so the backend kept rendering the whole book. Fixed end-to-end.

### Added
- **Stop button** while generating (danger variant). It cancels for real: frontend `AbortController` → `fetch` signal → `reader.cancel()` → the backend SSE renderer polls `request.is_disconnected()` at each chapter boundary, stops early, **keeps the resume manifest** (finished chapters are content-addressed cached), and emits a terminal `stopped` event. The UI shows "Stopped — finished chapters are cached; Create again to resume," not an error.
- **Live per-chapter progress**: a progress bar (completed/total), elapsed (mm:ss) and an ETA from average chapter time, and a per-chapter list marking pending / rendering / done / cached / failed — driven by the existing stream events.
- `Cmd/Ctrl+Enter` in the editor triggers Create; empty-script hint points to Load sample.

### Fixed
- The audiobook SSE renderer now cleans up on client disconnect instead of rendering to completion after the client is gone.

Backward-compatible; an untouched flow behaves exactly as before. Extracted `GenerationProgress`/`PlanList`/`AudiobookResult` (AudiobookTab back under the max-lines lint). Tests: `AudiobookStopCancel.test.jsx`, `longformStream` abort/cancel cases, `tests/test_audiobook_cancel.py`. Verify: `bun run test` 1476 passed; `test:prod-bundle` 2 passed; audiobook/longform backend 270 passed; parity 104.

Adds true audiobook generation cancellation by propagating `AbortController` and client disconnect detection (`request.is_disconnected`) through the SSE longform renderer, which stops after the current chapter boundary, emits a terminal `stopped` event, records the job as cancelled, and preserves the durable resume manifest plus cached/finished chapters for resuming. Improves UX with a dedicated Stop button (including unmount cleanup), live per-chapter progress with elapsed time and ETA, and clearer “stopped” messaging; also adds `Cmd/Ctrl+Enter` to create and an empty-script hint. Risk to review: correctness of partial/interrupt behavior (only intended chapters render, stream cleanup prevents late events) and that cancellation is consistently treated as cancelled (not failed) while resume metadata remains usable.


debpalash:
Review findings fixed on-branch:
- **Abort on unmount (Major):** added a `useEffect` cleanup that aborts the controller + sets the abort flag when the tab unmounts, so a mid-render tab switch stops the stream/backend job and a late `done`/`error` can't clobber a newer generation's output.
- **Cancelled vs failed:** a client-disconnect Stop now calls `job_store.mark_cancelled` (which exists) instead of `mark_failed` — a user stop is a cancellation.
- **Test race:** the mocked `read()` now rejects immediately if `signal.aborted` is already true, instead of hanging to timeout.
All verified: frontend stop-cancel + longformStream 7 passed, backend cancel 2 passed.

PR #1217, opened 2026-07-20
feat(audiobook): cast/voice mapping (fixes multi-voice) + markup toolbar + stats + validation (#1217)

debpalash:
Builds on #1216. Turns the Audiobook tab into a real authoring surface — and fixes multi-voice, which was silently broken.

### The multi-voice fix
`[voice:NAME]` never worked: the parser stored the raw NAME as the voice id, and `_resolve_voice` looked it up as a **profile UUID** — always missed — so every `[voice:…]` fell to the engine default (all narrators sounded identical). Now resolution happens in the synth resolver (`_map_span_voice`), not by mutating the parser:
- an optional `voice_map` (NAME → profile id) flows through the request → `_prepare_synth` → `_build_synth`'s `resolve` closure;
- a mapped NAME → its profile; an unmapped token that's a real profile id → itself (exact-id back-compat); an unmapped NAME → the book's **default voice** (fixes the silent-default bug); no `[voice:]` → default.
- `voice_map` is folded into **every** cache signature (chapter + segment + preview) like `ExpressiveOptions`; an absent/empty map yields today's exact keys, so **existing books don't re-render**.

### Added (UI)
- **Cast** panel: parses the distinct `[voice:NAME]`s and maps each to a profile (persisted, backward-compatible store field).
- **Markup insert toolbar**: `[pause]`, `[voice:]`, `[slow]`/`[fast]`/`[emphasis]`/`[spell]` (wraps selection), reaction tags — inserted at the cursor.
- **Live stats**: chapters · words · est. runtime.
- **Pre-flight validation**: flags unknown voices (clears once mapped) and empty chapters.

New pure utils (`audiobookScript.js`), extracted components, 19 i18n keys ×21 locales. Also freezes the #1216 progress clock on terminal state (last review note).

Behavioral nuance: a multi-voice book that today renders in one voice will re-render once (its names now resolve correctly) — that's the fix. Single-voice books are byte-identical.

Verify: `bun run test` 1494 passed; `test:prod-bundle` 2 passed; audiobook/longform backend 280 passed; cast+expressive 130 passed; parity/cjk/changelog green.

The Audiobook tab is now an authoring surface with a Cast panel for mapping `[voice:NAME]` tokens to voice profiles, a markup insertion toolbar, live chapter/word/runtime stats, and pre-flight validation warnings for unknown voices, empty chapters, and unknown tags, with voice assignments persisted client-side. Backend rendering now remaps span cast tokens to concrete profile IDs via a per-request `voice_map` across preview, generation, and resume, and adds a cache-signature so multi-voice changes correctly trigger re-rendering while preserving backward-compatible cache behavior for single-voice books. The main risk for human review is around voice-map fallback and token handling (unmapped names → default voice, exact profile-id vs token), plus cache-key invalidation and validation edge cases (especially empty chapters and tag recognition).


debpalash:
Bot findings triaged:
- **Stats `~` markers (Minor):** fixed — removed `~` before the exact word count and before the already-'est.' runtime across all 21 locales.
- **Changelog style (Minor):** fixed — neutral one-liner, no bold-lead/colloquial wording.
- **Module-level test imports (Major):** intentionally kept. `.coderabbit.yaml`'s instruction targets tests that pollute `sys.modules` (app reloads / TestClient lifespan). This suite imports only pure dataclasses/functions (`Chapter`/`Span`/`ExpressiveOptions`/`voice_map_signature`) and never reloads or monkeypatches `services.audiobook`, so there's no stale-reference risk — and it matches the established sibling `tests/test_audiobook_expressive.py` (same module-level import, green on main). Moving these to function scope in isolation would only add inconsistency.

PR #1218, opened 2026-07-20
fix(export): audiobook/story download hijacked the desktop app; blank-guard fallback broken on macOS (#1218)

debpalash:
**P0, user-reported.** Clicking Download on a finished audiobook blanked the whole app.

### Root cause
`AudiobookResult` (and Stories) used `<a href={mediaUrl} download>`. In the desktop WebView (WKWebView especially) the `download` attribute isn't honored for a media URL it can play — WebKit **navigates the whole window to the m4b and plays it fullscreen**, replacing the app. `#root` goes empty, the blank-window guard reloads 3× and paints its failure page, and its probe then throws `report_render_state not allowed` (it's running on the media page). One download click cascades into the whole thing.

### Fix
- New shared `downloadMedia(url, name, { sourceFilename })` (`utils/mediaDownload.js`) — the mechanism `App.jsx` already used correctly: **desktop** → native Save dialog + a server-side copy via the pre-existing `/export` endpoint (path-safe `_safe_source`/`_safe_destination`); **browser/Docker** → the existing blob download. No `<a href={media} download>` anywhere.
- Routed the **audiobook** and **Stories** downloads through it; `App.jsx`'s `triggerDownload` DRY'd onto the same util (its callers unchanged). Whole-class grep: those were the only two raw-anchor sites.
- **Blank-guard hardening** (defense-in-depth for the same incident): the probe now `.catch()`es its invoke (kills the unhandled rejection); the fallback page is shown by **injection** instead of a `data:` URL the WebViews refuse to top-frame-navigate to (the `Not allowed to navigate top frame to data URL` error); and Reload invokes a new `recover_main_window` Rust command that re-navigates to the app's entry URL — so it returns to the app even if the window had wandered off to a media file.

Cross-platform (macOS/Windows/Linux) — the Save-dialog + server-copy path is the parity-safe route. No new deps. Verify: `bun run test` 1498; new `mediaDownload.test.js` 4; `test:prod-bundle` 2; `cargo test --lib blank_guard` 8.

Introduced a shared `downloadMedia` utility to route audiobook, Story, and app downloads through the native Tauri Save dialog plus server-side export/copy on desktop—preserving blob-based downloads in browsers/Docker—and removed raw `download` anchors that could hijack the WebView/app window; it also added `res.ok`/status guards for subtitle and dynamic-endpoint paths and moved success/history recording so cancelled downloads don’t show “export done.” Hardened the blank-window guard by swallowing probe/invoke errors, injecting the fallback page into the current document instead of using a `data:` navigation, and adding a `recover_main_window` IPC command so Reload restores the main webview (falling back to `location.reload()` when needed). Human review is most worth a look for correctness and edge cases in the Tauri save-path/export/copy flows and the blank-guard injection/recovery behavior across navigation/registration failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,8 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 
 ### Fixed
 
+- Downloading a finished audiobook (or story mix) no longer hijacks the app: in the desktop WebView a plain download link to the media file made WebKit navigate the whole window to it and play it fullscreen (then the blank-window guard misfired) — downloads now go through the native Save dialog + a server-side copy instead (#1218)
+- The blank-window guard's fallback page is shown by injection rather than a `data:` URL the desktop WebViews refuse to navigate to, and its Reload button now returns to the app even if the window had navigated away (#1218)
 - Security (server mode): the admin routes (`/system/*`, `/api/settings/*` — RCE-class) now require the API key or genuine loopback — with an API key set and `OMNIVOICE_TRUSTED_NETWORKS` configured, a trusted-network client could previously reach them with no credential; the short share PIN no longer gates admin either (#1213)
 - A render error no longer blanks the whole window — a recoverable error card (Reload / Report) appears instead, and CI now builds the real production bundle so a pre-mount crash can't ship (#1209)
 - Voice-clone trimmer: the preview now plays exactly the selected region on variable-bitrate clips (it had drifted off on VBR/mis-reported-duration files by playing the original file on a different timeline) (#1210)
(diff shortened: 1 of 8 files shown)


debpalash:
Both findings fixed on-branch:
- **res.ok guard (Major):** `mediaDownload` now checks `res.ok` before reading the body on the subtitle and dynamic-endpoint save paths — a 4xx/5xx no longer gets written to disk or reported as a successful save (matches `browserDownload`). Test mock updated.
- **False Stories toast (greptile):** removed the unconditional `toast.success` + direct `recordValueMoment`; `downloadMedia` owns the toasts and fires `onValueMoment` only on a real save, so a cancelled save dialog no longer says 'export done'.

PR #1219, opened 2026-07-20
feat(voices): pick voice-gallery (archetype) voices anywhere via materialize-on-select (#1219)

debpalash:
Foundation for "use the voice gallery everywhere a voice is picked." A gallery **archetype** is a design *recipe* (instruct + seed, no reference audio), so it can't be a voice id directly — picking one **materialises it into a real profile**, which the whole synthesis pipeline already understands. Zero resolver changes.

### Added
- **Gallery group in `VoiceSelector`** (default on) — so the audiobook default voice and each Cast row can now pick a designed-voice archetype. Favourites first; searchable against the whole catalog (`GET /archetypes?q=`, debounced), fetched only when a picker is actually opened (many mounted pickers make zero requests until used).
- **Materialize-on-select:** picking a gallery row calls `POST /archetypes/{id}/use` and emits the returned **profile id** — the `archetype:<id>` sentinel never reaches the parent/backend, so the value contract and every downstream resolver are unchanged. Spinner + "Adding voice…" while it renders; error keeps the previous value.
- **`/archetypes/{id}/use` is now idempotent** — dedups by the `personality` column so re-picking the same gallery voice reuses its one materialised profile instead of spawning duplicates (the render + insert only run on first use).
- Free-text `q` search on the archetype catalog so any of the hundreds of voices is reachable by typing.

Stories/Dub pickers (native `<select>`) migrate onto this in the follow-up (#1220). New i18n keys ×21 locales. Verify: `bun run test` 1502; `test:prod-bundle` 2; voice/archetype backend 151 + 33; parity/route-inventory green.

Adds a searchable, debounced Voice Gallery/archetype group to `VoiceSelector` (favorites prioritized, lazy-loaded on open) that materializes selections into real voice profiles via `POST /archetypes/{id}/use`, with localized loading/error messaging and updated i18n across supported locales. Updates the backend to support free-text archetype search and to make materialization idempotent by deduplicating on `personality`, including a write-time re-check to prevent concurrent duplicates. Main human review risk: picker state transitions and edge cases where materialization fails (ensuring the previously selected value is preserved and no duplicate profiles are created), though tests and parity checks reported passing.


debpalash:
Findings triaged:
- **TOCTOU dedup (P2):** fixed — re-check the `personality` row inside the write connection immediately before INSERT, reusing a concurrently-created row and dropping our just-rendered sample. (A UNIQUE index isn't viable: marketplace/persona imports reuse the `personality` column. This closes the realistic window for a single-user desktop app.)
- **Raw error toast (Minor):** fixed — `console.error` the detail, show a generic localized message; dropped the `{{message}}` placeholder from `addVoiceFailed` across all 21 locales.
- **Materialized voice not in recents (Minor):** intentional. A just-materialized voice is already the selected value and shows in the *Designed* group (more discoverable than recents for a fresh add); the recents helpers are component-local to SearchableSelect. Low value vs. the coupling it'd add.

PR #1220, opened 2026-07-21
feat(voices): gallery voices in the Stories + Dub pickers; drop the redundant dub presets (#1220)

debpalash:
Completes "use the voice gallery everywhere" (follow-up to #1219). Migrates the last two pickers off native `<select>` onto the shared, gallery-enabled `VoiceSelector`.

### Added
- **Stories editor** — the cast picker and the per-line voice override now use `VoiceSelector` (Gallery on), so a designed-voice archetype is pickable there. Preserves the store's "no voice = null" shape and the `↳ inherited` label; a gallery pick materialises to a real profile so preview/export are unchanged.
- **Dubbing** — the per-segment picker now uses `VoiceSelector` with the from-video clone (`auto:<slug>`) options preserved; writes `seg.profile_id` exactly as before.

### Changed
- The dub picker **drops its hardcoded design-presets group** — it's superseded by the richer Gallery. Backward-compatible: stored `preset:<id>` segments still expand to their instruct and generate identically (the expansion lives in `utils/segments.js`, untouched). Investigation note: `preset:` was a *frontend* sentinel expanded before the request — never a backend bug.

### Under the hood
- New opt-in `menuPortal` on `SearchableSelect`/`VoiceSelector` (body portal + fixed positioning) so the dropdown isn't clipped inside the dub table's **virtualized/overflow** list. Default off — audiobook and other pickers keep the inline dropdown, unchanged.

No backend change, no new i18n keys. Verify: `bun run test` 1509; `test:prod-bundle` 2; dub/stories/voice backend 483; parity 104.

Stories and Dubbing voice pickers now use the shared gallery-enabled `VoiceSelector` (replacing native `<select>`), preserving existing “no voice = null”, inherited cast labels, and `auto:<slug>` cloning, while keeping legacy `preset:<id>` expansion compatible even as the Dubbing UI removes its hardcoded presets group. For the virtualized Dub per-segment list, `SearchableSelect`/`VoiceSelector` gained an opt-in `menuPortal` that renders the dropdown into `document.body` with viewport-aware fixed positioning that flips above and clamps height, updating on scroll/resize and handling click-outside correctly. Main risk to review: portal positioning/z-index and click-outside behavior plus any edge-case compatibility in stored preset/voice selection interactions and tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 ### Added
 
 - Voice picker: the designed-voice **Gallery** is now selectable anywhere a voice is chosen — the audiobook default voice and each Cast row can pick a gallery archetype (searchable, favourites first), and it's materialised into a real profile on pick so it just works everywhere (#1219)
+- The Stories editor and the Dubbing per-segment voice pickers now use the same gallery-enabled picker, so designed-voice archetypes are selectable there too; the dub picker drops its redundant hardcoded presets group in favour of the richer Gallery (existing picks unchanged) (#1220)
 - Audiobook tab: a Cast panel maps each `[voice:NAME]` in the script to a profile so multi-voice renders correctly (it previously fell back to a single voice), plus a markup insert toolbar, live stats (chapters · words · est. runtime), and pre-flight validation for unknown voices and empty chapters (#1217)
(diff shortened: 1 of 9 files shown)


debpalash:
Fixed: the portaled dropdown now **flips above the trigger** when there isn't enough room below (recomputed on scroll/resize), and the options list is capped to the available space so the chosen side always fits within the viewport — so a picker on the last dub segment row is no longer clipped off-screen.

Issue #1257, opened 2026-07-24
[Bug] 400 Bad Request: Invalid language code. Supported languages: ar (Arabic), da (Da


pulananave:
## Error

```
400 Bad Request: Invalid language code. Supported languages: ar (Arabic), da (Danish), de (German), el (Greek), en (English), es (Spanish), fi (Finnish), fr (French), he (Hebrew), hi (Hindi), it (Italian), ja (Japanese), ko (Korean), ms (Malay), nl (Dutch), no (Norwegian), pl (Polish), pt (Portuguese), ru (Russian), sv (Swedish), sw (Swahili), tr (Turkish), zh (Chinese)
Z@tauri://localhost/assets/client-C88Evhtg.js:3:10709
```

## Environment

**Version:** `0.4.0`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.5.2-arm64-arm-64bit`
**Python:** `3.11.15`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M1 Pro`
**RAM:** `16 GB`
**Disk free:** `58.8 GB`
**Active TTS engine:** `mlx-audio`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 0 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-07-24T10:31:24.000Z (11 h ago)
**Exit:** `exit code 1`
**Uptime before crash:** 4 s
**Backend version:** `0.4.0`

```
… (truncated)
24 07:31:23,407 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.types
2026-07-24 07:31:23,408 INFO [alembic.runtime.plugins] setup plugin alembic.aut2026-07-24 07:31:23,439 INFO [core.onboarding] 🎉 Seeded demo voice profile 'OmniVoice Demo Voice'
2026-07-24 07:31:23,566 INFO [omnivoice.model] Preload skipped: k2-fsa/OmniVoice is not installed locally — the mod2026-07-24 07:31:23,567 INFO [omnivoice.model] Preload skipped: k2-fsa/OmniVoice is not installed locally — the model will load (and download if requested) on first use.
2026-07-24 07:31:23,567 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager started
[...]

Issue #1254, opened 2026-07-24
[Bug] download: ERROR: [youtube] DZpjig7qxM8: This video is DRM protected


gysahlgreene:
## Describe the bug

Importing a YouTube URL intermittently failed with a DRM protection error, even though the same URL had previously worked and succeeded when retried shortly afterwards.

The failure appears to be transient rather than consistently reproducible.

## Error

```
download: ERROR: [youtube] DZpjig7qxM8: This video is DRM protected
@tauri://localhost/assets/main-app-BMQKI9Yk.js:17:9685
```

## Environment

**Version:** `0.4.0`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.5.2-arm64-arm-64bit`
**Python:** `3.11.15`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M4 Pro`
**RAM:** `48 GB`
**Disk free:** `459.2 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 4 s before this report

## Recent actions

```
09:24:53 view:settings
09:26:03 view:dub
09:26:12 dub:ingest-url
```

## What I was doing

1 - Opened the Dub view.
2 - Submitted a YouTube URL that had worked successfully before.
3 - The import failed with a message stating that the video was DRM protected.
4 - Retried the same URL without changing it.
5 - The second attempt completed successfully.

debpalash:
Fixed and merged in #1264, shipping in **v0.4.1**.

Your observation that it was transient was the key to it — a genuinely DRM-protected video doesn't become downloadable thirty seconds later. What varies is the **player client**: YouTube serves a DRM-only format set to some clients for videos that aren't actually protected.

OmniVoice already had machinery for exactly that shape — it escalates through alternate player clients on a 403, because "extraction worked but this client can't have the media" is the same situation. DRM simply wasn't routed into it, so your only recovery was retrying by hand until a different client came up. It now escalates automatically, and if every client still refuses, you get a message explaining that rather than a raw yt-dlp line.

Thanks for taking the time to note the retry behaviour — that detail is what made it diagnosable.

Issue #1253, opened 2026-07-24
[Bug] ingest: 'zfz8994p'


dustmaker124-ui:
## Error

```
ingest: 'zfz8994p'
Error: ingest: 'zfz8994p'
    at a.onmessage (http://tauri.localhost/assets/main-app-BMQKI9Yk.js:17:9680)
```

## Environment

**Version:** `0.4.0`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36 Edg/150.0.0.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 3050 Laptop GPU (4 GB VRAM)`
**CPU:** `AMD64 Family 25 Model 68 Stepping 1, AuthenticAMD`
**RAM:** `15.3 GB`
**Disk free:** `49.4 GB`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 3 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-07-23T17:29:46.000Z (15 h ago)
**Exit:** `exit code 1073807364`
**Uptime before crash:** 3456 s
**Backend version:** `0.4.0`

```
… (truncated)
oded NOTHING from 5.3s of speech-level audio — falling back to the capture ASR engine for this session
2026-07-23 22:57:44,431 ERROR [omnivoice.capture_ws] dictation model sherpa-parakeet-tdt-v3 demoted on this machine — it will no longer be auto-selected. Pick it again in Settings to give it another chance.
2026-07-23 22:57:58,899 INFO [omnivoice.onset_align] onset-align: snapped 1/7 segment start(s) to speech onset
INFO:     connection closed
2026-07-23 22:58:28,441 INFO [omnivoice.asr] whisperx transcribing ~\AppData\Local\Temp\tmpfqnja9b9.wav (word_timestamps=True)
2026-07-23 22:59:13,427 INFO [omnivoice.asr] faster-whisper loading Systran/faster-whisper-large-v3 on cuda (float16)
2026-07-23 22:59:13,500 INFO [omnivoice.onset_align] onset-align: snapped 1/5 segment start(s) to speech onset
[...]

Issue #1247, opened 2026-07-23
[Bug] 400 Bad Request: Unknown model id: engine:kittentts


JavaxmI:
## Error

```
400 Bad Request: Unknown model id: engine:kittentts
ApiError: 400 Bad Request: Unknown model id: engine:kittentts
    at apiFetch (http://localhost:3901/src/api/client.ts:375:10)
    at async apiJson (http://localhost:3901/src/api/client.ts:384:14)
    at async http://localhost:3901/src/components/EngineCompatibilityMatrix.jsx:276:4
```

## Environment

**Version:** `0.4.0`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36 Edg/150.0.0.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cpu`
**GPU:** `NVIDIA GeForce RTX 4050 Laptop GPU (6 GB VRAM)`
**CPU:** `Intel64 Family 6 Model 186 Stepping 2, GenuineIntel`
**RAM:** `15.7 GB`
**Disk free:** `51 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 4 s before this report

## Recent actions

```
20:53:38 view:settings
```

debpalash:
Fixed and merged in #1264, shipping in **v0.4.1**.

`list_loaded()` advertised in-process engines as `engine:<id>` with `"unloadable": true`, but `unload()` only ever handled `tts`, `diarization` and sidecars — so the panel rendered an Unload button for ids the backend rejected. The engines had implemented `unload()` all along; only the routing was missing.

The contract test written for it immediately found a second instance: the warm dictation model (`capture-asr`) was listed the same way with no branch either. Both fixed, and the test now enumerates the listing rather than hard-coding ids, so a future model kind can't reintroduce it.

Thanks for the report.

PR #1264, opened 2026-07-26
fix(errors): six failures that reached users as raw text (#1262, #1256, #1251, #1247, #1257, #1254)

### LINKED ISSUE #1251: [Bug] 500 Internal Server Error: The paging file is too small for this operation to co
## Error

```
500 Internal Server Error: The paging file is too small for this operation to complete. (os error 1455)
ApiError: 500 Internal Server Error: The paging file is too small for this operation to complete. (os error 1455)
    at Z (http://tauri.localhost/assets/client-DFma2LtE.js:3:10539)
    at async Ci (http://tauri.localhost/assets/main-app-BMQKI9Yk.js:16:27868)
    at async http://tauri.localhost/assets/main-app-BMQKI9Yk.js:17:3638
```

## Environment

**Version:** `0.4.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`

## Backend reachability

**Last backend response:** 2 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-07-22T22:39:06.000Z (25 h ago)
**Exit:** `exit code -1073741819`
**Uptime before crash:** 1608 s
**Backend version:** `0.4.0`

```
… (truncated)
13:55,524 INFO [omnivoice.watermark] AudioSeal generator loaded (16-bit message mode)
2026-07-23 05:13:55,524 WARNING [Audioseal] Deprecated Warning: `sample_rate` is specified but it will be ignored. 
Consider removing `sample_rate` in the model call as this is a no-op
Starting from AudioSeal 0.2+, audio is not resampled internally to 16kHz or some predefined sample rates. The user is responsible for providing the correct sample rate to the model.

2026-07-23 05:25:12,317 WARNING [omnivoice.watermark] Watermark embedding failed (passing through original): [enforce fail at alloc_cpu.cpp:121] data. DefaultCPUAllocator: not enough memory: you tried to allocate 2212442816 bytes.
2026-07-23 05:28:13,673 INFO [omnivoice.asr] Idle timeout reached. Unloading capture ASR (FasterWhisperBackend) to free memory.
2026-07-23 05:28:43,790 INFO [omnivoice.model] Idle timeout reached. Unloading OmniVoice model to free VRAM.
2026-07-23 05:38:53,343 INFO [omnivoice.model] Importing PyTorch & OmniVoice runtime…
2026-07-23 05:38:53,343 INFO [omnivoice.model] Loading OmniVoice model on device: cuda
2026-07-23 05:38:53,343 INFO [omnivoice.model] Skipping PyTorch Whisper preload; ASR will load on demand.
```

## Recent actions

```
05:49:52 view:studio
05:51:02 generate:start (audio)
06:06:07 export
06:20:36 generate:start (audio)
```

### LINKED ISSUE #1256: [Bug] 500 Internal Server Error: Couldn't synthesize audio. See Settings → Logs → Back
## Error

```
500 Internal Server Error: Couldn't synthesize audio. See Settings → Logs → Backend for the full trace. Underlying error: RuntimeError: TTS engine stopped mid-generation with an error OmniVoice doesn't recognize. Retry once; if it keeps failing, please report it with the full trace. Underlying error: FileNotFoundError: [Errno 2] No such file or directory: 'ffprobe'
Z@tauri://localhost/assets/client-DFma2LtE.js:3:10544
```

## Environment

**Version:** `0.4.0`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.5.2-arm64-arm-64bit`
**Python:** `3.11.15`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M1 Max`
**RAM:** `32 GB`
**Disk free:** `174 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Last backend response:** 2 s before this report

## Recent actions

```
13:52:01 view:studio
13:52:46 generate:start (audio)
13:52:46 generate:stream-fallback
```

### LINKED ISSUE #1262: [Bug] 500 Internal Server Error: 'latin-1' codec can't encode characters in position 2
## Error

```
500 Internal Server Error: 'latin-1' codec can't encode characters in position 22-25: ordinal not in range(256)
ApiError: 500 Internal Server Error: 'latin-1' codec can't encode characters in position 22-25: ordinal not in range(256)
    at Z (http://tauri.localhost/assets/client-DFma2LtE.js:3:10539)
    at async p (http://tauri.localhost/assets/profiles-D9z3g2r4.js:1:908)
    at async se (http://tauri.localhost/assets/VoiceProfile-CEfHkrpi.js:1:13011)
```

## Environment

**Version:** `0.4.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 4060 (8 GB VRAM)`
**CPU:** `AMD64 Family 25 Model 33 Stepping 0, AuthenticAMD`
**RAM:** `31.9 GB`
**Disk free:** `262.2 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Last backend response:** 2 s before this report

## Recent actions

```
22:19:19 view:settings
22:45:53 engine:tts=indextts2
22:46:17 view:launchpad
22:46:17 view:studio
22:46:46 view:dub
22:46:47 view:stories
22:46:48 view:dub
22:47:20 view:projects
22:47:25 view:settings
22:47:27 engine:tts=omnivoice
22:47:33 view:launchpad
22:47:37 view:studio
22:47:42 view:dub
22:53:29 view:settings
22:55:06 view:studio
22:55:12 view:voice
```

debpalash:
Closes #1262. Closes #1256. Closes #1251.

Six separate reports, one shared shape: something failed, and the user was handed the operating system's own words with no indication of what to do. Grouped because they all land in the same classification layer.

---

### #1262 — a non-Latin voice name 500'd every download

```
500: 'latin-1' codec can't encode characters in position 22-25
```

`attachment; filename="` is **exactly 22 characters**, so positions 22-25 were the first four characters of the reporter's own profile name. HTTP header values are latin-1 by definition and every download endpoint interpolated the filename straight in.

The sanitisers looked like they covered it — they all filtered with `str.isalnum()`, which is `True` for **every alphabetic script**. `"我的声音".isalnum()` is `True`. They removed punctuation and passed through precisely the characters that break the header.

This was never one endpoint: the same idiom was copy-pasted across persona export, marketplace, stories, the OpenAI-compatible speech route, and **eight** dub-export routes — so any CJK/Cyrillic/Greek/Hebrew/emoji video title broke dub downloads too. All now go through one RFC 6266 builder that emits an ASCII `filename=` **and** an RFC 5987 `filename*=UTF-8''…`, so modern browsers get the user's real name back. A test fails CI if the header is ever hand-built again.

### #1256 — `FileNotFoundError: 'ffprobe'` reported as unrecognised

Two defects. **It happened at all:** every OmniVoice call site resolves ffmpeg/ffprobe explicitly, which is why a bundled sidecar that was never on `PATH` works for us — a *dependency* shelling out by bare name gets nothing. The reporter hit this mid-synthesis on a Mac where the app's own ffprobe was resolvable the entire time. The resolved directories are now published on `PATH` at startup, which fixes every such dependency at once. **It was illegible:** `classify()` had no match, so the generic wrapper said the engine "stopped with an error OmniVoice doesn't recognize". It now names the media engine and points at Settings → Audio tools. Deliberately narrow — ffmpeg reporting a missing *input* file is a different problem and must not get this remedy.

### #1251 — Windows paging file

`The paging file is too small for this operation to complete. (os error 1455)` reached the user as a bare 500. The generate path already counted the phrase as OOM, but that advice — close other apps, pick a lighter engine — is the wrong remedy on a 32 GB machine. Windows isn't short of RAM; it's refusing to back a large mapping. The hint now gives the actual Virtual-memory steps. Note `os error 1455`, not `[WinError 1455]` — it arrives via the Rust safetensors mmap, so matching only Python's spelling missed it. `_is_oom_failure` still returns True, so retry/queue behaviour is unchanged.

---

Each fix has fail-before/pass-after coverage. Full backend suite: **3599 passed**. One existing assertion in `test_dub_subtitles_309.py` asserted the old single-parameter header shape (`endswith('.srt"')`) and was updated to assert content rather than position.

---

### #1247 — `400 Unknown model id: engine:kittentts`

`list_loaded()` advertises in-process engines as `engine:<id>` with `"unloadable": true`. `unload()` handled `tts`, `diarization`, `sidecars`, `sidecar:<id>` — and nothing else. The panel was rendering an Unload button for ids the dispatcher rejected. The engines have implemented `unload()` the whole time; only the routing was missing.

The guard test enumerates the listing rather than hard-coding ids, and **immediately found a second instance**: `capture-asr` (the warm dictation model) was listed the same way with no branch either. Both fixed. An explicit unload releases the dictation model now rather than after the idle timeout, but still declines while a live session holds the lease.

### #1257 — a language the engine can't speak

The cause is written down in the code:

```python
# Per-model; Kokoro supports 8, Qwen3 ~4, Kugel 24. Return "multi"
# so the language picker doesn't gate by engine — each engine
# silently ignores languages it doesn't know.
return ["multi"]
```

The last clause is false — the library raises. So the picker offers all 646 languages and the reporter got a bare recitation of 23 codes that named neither the engine refusing nor the fact that switching engines was the fix. Enumerating each model's real language set would be a brittle map that goes stale on every engine update; naming the engine and the way out is accurate and stays accurate. Rewritten as a `ValueError` so the route keeps returning 400, and inert for every other failure (an OOM rewritten as a language problem would be worse than the bug).

### #1254 — "DRM protected", intermittently

The reporter's own note: the same URL failed, then succeeded on a retry shortly after. Real DRM does not lapse in thirty seconds — the **player client** varies, and YouTube serves a DRM-only format set to some clients for videos that aren't protected. That is exactly the shape `_is_forbidden_download_error` already handles for a 403 by escalating through `_YT_PLAYER_CLIENTS`; DRM simply wasn't routed into it, so the user's only recovery was retrying by hand until a different client came up. Now escalated automatically, and kept out of the *transient* retry budget (blind-retrying the client that just refused is what the reporter was doing manually). If every client still refuses, the failure is classified instead of arriving as a raw yt-dlp line.

Full backend suite after both commits: **3626 passed**.

---

## Split note

The dub delete-resurrection fix (#1252, #1253) **moved to #1270**. Everything here is an independent error-message change that needed no corrections; that one is a concurrency change that needed five rounds, the last of which found the fix didn't close the reported case at all. It ships on its own rather than gating a release.

Backend suite after the split: **3628 passed**.


debpalash:
@coderabbitai review

Issue #1269, opened 2026-07-26
[Test] Four backend tests fail when both suites run in one pytest session — CI's split hides it


debpalash:
## What

Four tests in `backend/tests/` fail when the two suites run in **one** pytest session, and pass when `backend/tests/` runs alone:

```
backend/tests/test_audiobook_resume_api.py::test_jobs_lists_interrupted_with_progress
backend/tests/test_model_load_shutdown.py::test_lifespan_shutdown_mid_load_is_clean_and_clears_sentinel
backend/tests/test_personas_api.py::test_import_roundtrip_creates_profile
backend/tests/test_personas_api.py::test_import_preview_only
```

Reproduce:

```bash
uv run pytest tests/ backend/tests/ -q     # 4 failed, 3842 passed
uv run pytest backend/tests/ -q            # all pass
```

Confirmed on an unmodified `main` — not caused by any current branch.

## Why CI doesn't catch it

`.github/workflows/ci.yml` runs the two suites as **separate steps**:

```yaml
- name: Run pytest
  run: uv run pytest tests/ -q --tb=short
- name: Run pytest (backend/tests, isolated)
  run: uv run pytest backend/tests/ -q --tb=short
```

so the interfering ordering never occurs there. The comment on that second step says the split is kept "for cheaper, clearer CI output" and notes the suites are safe to mix — that last part is no longer true.

## Why it matters

Green CI is currently telling us less than it appears to. These four tests only pass because of how CI happens to invoke them; a contributor running `pytest` at the repo root — the obvious thing to do — sees four failures and has no way to know they're expected. That erodes trust in the suite, which is the thing that catches everything else.

It's also a real signal: cross-suite interference means shared process state (module-level singletons, `sys.modules`, the run sentinel, or a data dir) is leaking between tests. Whatever leaks between two test files can leak between two requests.

## Likely direction

`backend/tests/conftest.py` sets up a hermetic data dir and no longer stubs `sys.modules`. The candidates are module-level singletons imported by `tests/` and then reused with different state by `backend/tests/` — `model_manager`, the run sentinel (`core/run_sentinel.py`), and the personas/profile store are the ones named in the failures. Each failure should be reduced to its minimal interfering pair (`pytest tests/<one> backend/tests/<other>`) before fixing.

## Suggested fix

Fix the leakage, then **run both suites in one session in CI** so the guarantee is enforced rather than assumed. Keeping them split is fine for output readability, but there should be at least one job that proves they compose.

---

[...]

PR #1272, opened 2026-07-27
feat(update): announce updates as a toast with actions, not a wall of text

debpalash:
A version's release notes **are** the whole changelog section — v0.4.1's was 42 bullets. Any surface that renders them inline becomes unusable.

Older builds put them in a blocking OS dialog that filled the screen and had to be dismissed before the app could be touched:

> Update available · A new version (0.4.1) of OmniVoice Studio is available. · What's new: `**Highlights**` … 42 bullets … · Download and install now? · [No] [Yes]

That dialog is already gone from current builds. But removing it left the **opposite** failure: the only remaining signal is a 6-pixel dot beside the version number in the footer, which is easy to never notice. Users either got shouted at or told nothing.

## The toast

Names the version, offers three actions, and leaves:

- **Install and restart** — declines while a generation is running, since the relaunch would lose it
- **What's new** — routes to Settings → Updates, where the notes have room
- **Later** — dismisses

Keyed by version, so the 6-hourly re-check replaces rather than stacks. Never auto-dismisses: an update the user hasn't answered is still true. Lazy-imported, so it never loads in a browser or dev build that can't update.

## Tests

Pin the shape that actually caused the bug: the toast takes **no notes prop at all**, renders under 200 characters, routes to the notes rather than inlining them, and can't stack duplicates.

## i18n

Strings added to all 21 locales — translated, not English-filled.

Frontend suite: **1543 passed**. Typecheck clean.

Replaced blocking update release notes with a persistent, version-keyed toast offering install/restart, What’s new, and Later, with localized strings and broad busy-state protection. Improved update reliability by tracking overlapping TTS activity, preserving cache self-healing, making purge markers deterministic, and handling shutdown errors as retryable 503s with CORS support. Human review should focus on the expanded busy-state contract, update-install behavior, and translated toast wording across all locales.


debpalash:
## Review triage — round 2

**Fixed:**

| Finding | Verdict |
|---|---|
| Greptile P1 — restart guard only knew about dub synthesis | Real. Replaced with a shared `isAppBusy` covering the dub state machine, the status pill and synth. |
| Greptile P1 — preview synthesis bypasses the guard | Real, same root as below. |
| Greptile P1 — overlapping synths clear each other | Real. Both fixed by counting at `api/generate.ts`, the one `/generate` call all seven synth paths share, instead of tracking in `useTTS`. Fail-before verified: the overlap test fails against the boolean. |
| CodeRabbit Major — `busy` is a render-time snapshot | Real. `UpdatesPanel` re-reads at click time. |
| CodeRabbit Major — 503 skips the CORS headers | Real, and it would have defeated the fix: the browser would report a bare CORS failure and the actionable detail would never reach the user. Both error responses now build headers through one helper. |
| CodeRabbit Major — don't log the raw request URL | Fair. Logging `request.url.path`. |
| CodeRabbit Minor — `update.busy` still says "finish your dub" | Real. The guard now covers uploads, transcription, translation, export and synth, so the message can't name one. Rewritten and translated across all 21 locales. |
| CodeRabbit Minor — ru `common.dismissStatus` reads as "reject" | Real, missed in my earlier sweep. |

**Declined, with reasons:**

- **`(#NNN)` ref on the Highlights bullet** (raised twice) — Highlights bullets carry no issue refs by design; the refs live on the `###` subsection entries. That's the documented house style and `tests/test_changelog_style.py` enforces it, so the deterministic check is the authority here rather than the model's reading of the guideline.

- **Gate the cache re-download behind explicit user action** — the self-heal is pre-existing behaviour that already ran for the sibling wording of this same error; this change only stops it missing half the class. Gating it now would be a regression for the path that already works, and would leave users with a broken cache and no repair. It remains behind the existing `_hf_offline()` check (verified at `model_manager.py:1166`), so nothing downloads when `HF_HUB_OFFLINE`/`TRANSFORMERS_OFFLINE` is set — the guideline's actual concern.

Full suites green: 1581 frontend, and the backend suites this touches. The combined `tests/` + `backend/tests/` session leaves only the four pre-existing #1269 isolation failures — my own new tests initially added to that set (a duplicate-module import broke a bare `isinstance`) and are now isolation-clean.

Issue #1190, opened 2026-07-19
[Bug] 503 Service Unavailable: TTS generate exceeded 300s and was abandoned — the back


Rahul-kr-Gupta:
## Error

```
503 Service Unavailable: TTS generate exceeded 300s and was abandoned — the backend is running, but the job was too heavy for the available compute. Capacity was restored automatically; most often the GPU is VRAM-starved (a resident model and this job contend for memory). For a durable fix, Flush caches / Unload the resident model (top toolbar or Settings → Models) before retrying, try shorter text, a lighter engine, or set the engine to CPU in Settings → Models. (Raise OMNIVOICE_GENERATE_TIMEOUT_S for very long single generations.)
ApiError: 503 Service Unavailable: TTS generate exceeded 300s and was abandoned — the backend is running, but the job was too heavy for the available compute. Capacity was restored automatically; most often the GPU is VRAM-starved (a resident model and this job contend for memory). For a durable fix, Flush caches / Unload the resident model (top toolbar or Settings → Models) before retrying, try shorter text, a lighter engine, or set the engine to CPU in Settings → Models. (Raise OMNIVOICE_GENERATE_TIMEOUT_S for very long single generations.)
    at D (http://tauri.localhost/assets/client-D_2HFpEC.js:2:5623)
    at async fe (http://tauri.localhost/assets/VoiceProfile-CcXWi8Lq.js:1:15062)
```

## Environment

**Version:** `0.3.22`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36 Edg/150.0.0.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 4060 (8 GB VRAM)`
**CPU:** `Intel64 Family 6 Model 151 Stepping 5, GenuineIntel`
**RAM:** `31.8 GB`
**Disk free:** `32.5 GB`

## Recent actions

```
06:13:35 view:voice
06:13:38 view:projects
06:14:18 view:voice
06:14:38 view:launchpad
06:14:41 view:projects
06:14:43 view:transcriptions
06:14:54 view:audiobook
06:15:05 view:launchpad
06:15:08 view:studio
06:15:41 generate:start (audio)
06:18:07 view:projects
06:18:35 view:launchpad
[...]

PR #1292, opened 2026-07-28
feat(engines): add omnivoice-subprocess, a crash-isolated (killable) TTS engine

### LINKED ISSUE #1291: Residual: abandoned GPU worker still holds the device after the pool reset (#730/#1190); opt-in crash-isolated engine proposed
## The residual gap

#730 and #1190 are closed, and the current code mitigates the symptom well: on an execution-timeout the GPU pool is reset so new work isn't queued behind a wedged job, and the error messaging tells the user to wait or restart. Thank you for that.

But the underlying worker is a `ThreadPoolExecutor` thread, and Python cannot interrupt a thread that is blocked in a native torch / MPS / CTranslate2 call. `_reset_gpu_pool()` only swaps the inner pool (`_ResilientGpuPool.reset`); `ThreadPoolExecutor.shutdown` cannot stop the running thread. So the abandoned worker **keeps holding the device until it finishes on its own**, and an immediate retry still contends with the zombie. On a VRAM-tight MPS box one slow synth can cascade into a stuck, device-holding backend until a manual restart, which is the failure users still hit under sustained / unattended use.

Reproducer (stdlib, no model needed): a 1-worker `ThreadPoolExecutor` running a blocking call survives `asyncio.wait_for` cancellation + `shutdown(wait=False)` by its full runtime, holding the (simulated) device the whole time. The pool's own honest comment in `model_manager.py` already says as much.

## Proposed complementary fix: an opt-in crash-isolated engine

I've opened a PR that adds `omnivoice-subprocess`, an opt-in engine that runs the **same model in a child process** via the existing `SubprocessBackend` primitive. A child process *can* be hard-killed: on a recv-timeout the parent's watchdog calls `proc.kill()`, reclaiming VRAM/device, and the next request transparently respawns a fresh sidecar. The in-process `omnivoice` engine stays the default, so this changes nothing for existing users; it is an opt-in for unattended / scheduled / reaction-triggered synthesis where a stuck job must self-recover instead of hanging until a restart.

The PR also:

- Fixes a latent cold-load bug in `SubprocessBackend.generate()` where a sidecar's `{"op":"progress"}` frames during load made the first cold generate fail-then-work (additive; engines that reply with audio directly are unaffected).
- Adds `make_room_before_generate()` to free idle GPU memory before a warm, heavy generate (the cold-load path already evicted; the warm path skipped it).

Verified end-to-end against the live model and under a sustained + concurrent-pressure soak (killed-worker recovery 5/5, chunked long text 9/9, no memory leak).

Happy to adjust scope, naming, or split the PR if the maintainers prefer. This is a complement to the existing mitigations, not a replacement.

paoloantinori:
Closes #1291.

## Why

#730 and #1190 are closed, but the residual root cause remains: the GPU worker is a `ThreadPoolExecutor` thread, and on a timeout the pool reset can't kill it (Python can't interrupt a native torch/MPS call). The abandoned worker holds the device until it finishes on its own, so an immediate retry still contends with the zombie, and on a VRAM-tight MPS box one slow synth can cascade into a stuck backend until a manual restart.

## What

Adds an **opt-in** `omnivoice-subprocess` engine that runs the same model in a child process via the existing `SubprocessBackend`. A child process can be hard-killed: on a recv-timeout the parent's watchdog calls `proc.kill()`, reclaiming VRAM/device, and the next request transparently respawns a fresh sidecar. The in-process `omnivoice` engine stays the default, so there is no behavior change for existing users; this targets unattended / scheduled / reaction-triggered use where a stuck job must self-recover.

### Changed
- **New engine** `omnivoice-subprocess` (`backend/engines/omnivoice_subprocess/`): the sidecar loads the model via the parent's own `from_pretrained` path (`venv_python` = `sys.executable`, isolating for crash recovery rather than dependency pins). Opt-in via `OMNIVOICE_TTS_BACKEND=omnivoice-subprocess` or Settings -> Engines.
- **`SubprocessBackend.generate()`** now consumes non-terminal `{"op":"progress"}` frames a sidecar emits during a cold load (previously the first cold generate after spawn failed, then worked on retry). Additive: engines that reply with audio directly are unaffected.
- **`recv_timeout_s`** is overridable per engine (default 60s unchanged); the new engine aligns it with the generate budget so a long-but-valid synth is not falsely killed while a wedged one still is.
- **`make_room_before_generate()`**: free idle GPU memory before a warm, heavy generate (the cold-load path already evicted; the warm path skipped it). Env-tunable, no-op on a roomy machine.

### Tests / verification
- Unit: `test_omnivoice_subprocess.py` (registry, recv-timeout override, roundtrip, **hard-kill + recovery**, progress-loop) and `test_make_room_before_generate.py` (policy incl. malformed-env). 18 new tests; the repo's version-lockstep / CJK / analytics / docs-drift gates also green.
- End-to-end on the live model (cold / warm / recovery-after-kill) and a sustained + concurrent-pressure soak: killed-worker recovery 5/5, chunked long text 9/9, no memory leak.
- Docs: `docs/engines/omnivoice-subprocess.md` + a `docs/features.yaml` entry.

### Notes
- Same model and output quality as the default engine; slightly higher per-call overhead (one stdio round-trip). Does not carry the native advanced-param surface (`t_shift` etc.) or parent-side seed determinism, since the generic engine path doesn't forward those. Fine for clone/design synthesis.
- A complement to the existing reset/messaging mitigations, not a replacement.

Happy to split the PR (engine vs `make_room`) or rename if preferred.

Adds an opt-in `omnivoice-subprocess` TTS engine that runs OmniVoice in a crash-isolated child process via `SubprocessBackend`, hard-killing/respawning on receive timeouts and correctly handling cold-load progress frames and per-engine (env-overridable) receive-timeout settings. Centralizes VRAM/idle GPU memory “make room” behavior into `get_model()` with an `OMNIVOICE_FREE_VRAM_BEFORE_GENERATE` (auto/always/never) policy, and updates docs/tests to cover registration, protocol roundtrips, progress consumption, timeout recovery, and the warm-path memory policy. Main risk to review: subprocess stdio protocol/lifecycle correctness under progress + wedged-process timeout recovery, and the RAM-probe decisioning that gates VRAM eviction.

### DIFF
--- a/backend/api/routers/generation.py
+++ b/backend/api/routers/generation.py
@@ -931,6 +931,14 @@ async def generate_speech(
     except Exception:
         pass
 
+    # Free idle GPU before a warm, heavy generate (#730/#1190); no-op on a roomy
+    # machine or a short synth. Covers MCP generate_speech, which POSTs here.
+    # Run off the event loop: the eviction does gc.collect + cache drop + ASR
+    # teardown that can block for hundreds of ms when it trips.
+    from services.model_manager import make_room_before_generate
+    await asyncio.get_running_loop().run_in_executor(
+        None, make_room_before_generate, text)
+
     _model = None
     _backend = None
     if backend_cls is OmniVoiceBackend:
--- a/backend/api/routers/generation.py
+++ b/backend/api/routers/generation.py
@@ -931,13 +931,8 @@ async def generate_speech(
     except Exception:
         pass
 
-    # Free idle GPU before a warm, heavy generate (#730/#1190); no-op on a roomy
-    # machine or a short synth. Covers MCP generate_speech, which POSTs here.
-    # Run off the event loop: the eviction does gc.collect + cache drop + ASR
-    # teardown that can block for hundreds of ms when it trips.
-    from services.model_manager import make_room_before_generate
-    await asyncio.get_running_loop().run_in_executor(
-        None, make_room_before_generate, text)
+    # VRAM eviction runs in get_model()'s warm-return path now, so every native
+    # TTS generate (this route, WS TTS, dub, batch, audiobook) is covered.
 
     _model = None
     _backend = None
--- a/backend/api/routers/openai_compat.py
+++ b/backend/api/routers/openai_compat.py
@@ -385,6 +385,12 @@ async def create_speech(req: SpeechRequest):
     from services.text_normalization import normalize_for_tts
     text = normalize_for_tts(req.input, req.language)
 
+    # Free idle GPU before a warm, heavy generate (#730/#1190); no-op on a roomy
+    # machine or a short synth. Run off the event loop (gc.collect + cache drop).
+    from services.model_manager import make_room_before_generate
+    await asyncio.get_running_loop().run_in_executor(
+        None, make_room_before_generate, text)
+
     # ── #1033/#1037/#1014: warm the engine under the LOAD budget before the
     # generate clock starts. The T4 verification (#1014) measured a fresh
     # install's first /v1/audio/speech burning its whole 300s generate budget
--- a/backend/api/routers/openai_compat.py
+++ b/backend/api/routers/openai_compat.py
@@ -385,11 +385,8 @@ async def create_speech(req: SpeechRequest):
     from services.text_normalization import normalize_for_tts
     text = normalize_for_tts(req.input, req.language)
 
-    # Free idle GPU before a warm, heavy generate (#730/#1190); no-op on a roomy
-    # machine or a short synth. Run off the event loop (gc.collect + cache drop).
-    from services.model_manager import make_room_before_generate
-    await asyncio.get_running_loop().run_in_executor(
-        None, make_room_before_generate, text)
+    # VRAM eviction runs in get_model()'s warm-return path now, covering every
+    # native TTS generate (this route, WS TTS, dub, batch, audiobook).
 
     # ── #1033/#1037/#1014: warm the engine under the LOAD budget before the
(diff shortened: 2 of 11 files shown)


paoloantinori:
Two review findings addressed in f1d077d6:

- Greptile (long-text eviction never fires): correct, it was dead code. The long-text headroom boost became unreachable once the eviction moved into `get_model()` (which has no text), so I removed the long-text branch, the `text` param, and the now-unused `_env_float` helper rather than keep unreachable logic. The core RAM-tight gate is unchanged, which is the part that fires on a starved box.
- CodeRabbit (silent exception swallow): the `available_memory` probe failure now logs at debug level instead of being silently swallowed.

Tests updated for the text-agnostic policy; CI re-runs on this push.

paoloantinori:
CodeRabbit: pre-existing, not introduced here, so out of scope for this deadlock fix.

The off-pool slot acquire (the branch taken by the engine self-test and the diagnostic probe) submits a no-op that completes the instant a worker picks it up, so it has always been a queue-wait for "no pool worker is busy," not a reservation held across the synthesis. The existing `finally` comment already says so ("the slot returns to the pool when the no-op finishes, immediately after .result()").

The main dispatch path is different: `/v1/audio/speech`, `/generate`, audiobook, dub, and batch all run `generate()` ON a pool worker via `run_on_gpu_pool_guarded`, which holds `_running` for the whole sidecar exchange (verified in review), so the synthesis is correctly serialized there. The off-pool callers are bounded and rare. Tightening the off-pool path is a reasonable follow-up, but separate from this fix.

PR #1296, opened 2026-07-29
fix(engines): stop SubprocessBackend.generate() self-deadlock on 1-worker (MPS) pools

### LINKED ISSUE #1295: Subprocess TTS engines deadlock on MPS (1-worker pool) via /v1/audio/speech and /generate
## Bug

`SubprocessBackend.generate()` (the base for every subprocess-isolated TTS engine: IndexTTS-2, Supertonic-3, dots.tts, MOSS-TTS-v1.5, Confucius4, and the proposed `omnivoice-subprocess`) acquires a GPU-pool slot for accounting:

```python
slot_future = pool.submit(lambda: None)
slot_future.result(timeout=10)
```

But `/v1/audio/speech` and `/generate` (and audiobook / dub / batch) dispatch `backend.generate()` via `run_on_gpu_pool_guarded`, i.e. **already on a gpu-pool worker**. On a 1-worker pool (MPS / Apple Silicon) the inner `pool.submit` queues behind the very job running it, `result(timeout=10)` raises at ~10s before the sidecar ever spawns, and the request surfaces the in-process "exceeded 300s ... abandoned" 503 instead of synthesizing. Every subprocess engine is effectively unusable through the HTTP routes on MPS.

The engine works when called directly / off the pool, which is why it can pass unit tests yet fail in the running server.

## Fix

Skip the slot acquisition when `threading.current_thread()` is already a gpu-pool worker: the outer `run_on_gpu_pool_guarded` already accounts for the slot (it holds `_running` for the whole sidecar exchange via `_tracked`), so re-acquiring only double-counts and deadlocks. Direct off-pool callers (the engine self-test in `engines.py`, the diagnostic probe in `diagnose.py`) still take a slot as a queue-wait. Verified end-to-end on a live MPS server: `/v1/audio/speech` returns HTTP 200 with real audio, sidecar spawns; regression test added (generate dispatched on a pool worker).

This is the general bug; the same fix is present in #1292 (commit e9a15d53) for the omnivoice-subprocess engine. Linked: #730, #1190.

paoloantinori:
Closes #1295.

`SubprocessBackend.generate()` acquires a GPU-pool slot for accounting, but `/v1/audio/speech` and `/generate` (and audiobook / dub / batch) dispatch `backend.generate()` via `run_on_gpu_pool_guarded`, already ON a pool worker. On a 1-worker pool (MPS) the inner `pool.submit` queues behind the very job running it and `result(timeout=10)` raises before the sidecar spawns, so every subprocess engine (IndexTTS-2, Supertonic-3, dots.tts, MOSS-TTS-v1.5, Confucius4) surfaces the in-process 300s-abandon instead of synthesizing.

### Changed
- `generate()` skips the slot acquisition when `current_thread()` is already a gpu-pool worker; the outer `run_on_gpu_pool_guarded` already accounts for the slot (it holds `_running` for the whole sidecar exchange via `_tracked`). Direct off-pool callers (the engine self-test in `engines.py`, the diagnostic probe in `diagnose.py`) still take a slot as a queue wait.

### Tests / verification
- New regression test: `backend/tests/test_subprocess_slot_deadlock.py` (generate dispatched on a pool worker; pre-fix it raised the ~10s slot timeout before the sidecar spawned).
- Verified end-to-end on a live MPS server: `/v1/audio/speech` returns HTTP 200 with real audio, sidecar spawns.

### Notes
- Independent of #1292 (the omnivoice-subprocess feature, which carries the same fix at `e9a15d53`). This PR is the standalone general bug fix; if it merges first, #1292 rebases onto it.

Updated `SubprocessBackend.generate()` to skip the inner GPU-pool “slot” acquisition when the call already runs on a `gpu-pool` worker thread, preventing self-deadlock on one-worker pools (including MPS) while keeping slot accounting for direct off-pool callers. Added a regression test that monkeypatches `_get_gpu_pool` to a deterministic one-worker `gpu-pool` thread pool and uses a stub sidecar to assert `generate()` completes and returns the expected audio length without timing out. Risk is mainly the thread-name-based `gpu-pool` detection, which should be reviewed for brittleness across pool implementations.

### DIFF
--- a/backend/services/subprocess_backend.py
+++ b/backend/services/subprocess_backend.py
@@ -518,20 +518,24 @@ class SubprocessBackend(TTSBackend):
         sample rate. Decodes the int16 PCM the sidecar returns into float32
         in [-1, 1].
         """
-        # Lazy-import the GPU pool so importing this module doesn't pull in
-        # the entire model_manager + torch ecosystem at registry-listing time.
-        from services.model_manager import _get_gpu_pool
-
-        # Acquire a GPU pool worker for the duration of this generate. The
-        # try/finally guarantees the slot is released even if the sidecar
-        # dies mid-frame (T-02-02 / Pitfall 7).
-        pool = _get_gpu_pool()
-        slot_future = pool.submit(lambda: None)
-        try:
-            slot_future.result(timeout=10)  # wait for our turn
-        except Exception:
-            slot_future.cancel()
-            raise
+        # The HTTP routes, audiobook, dub, and batch all dispatch generate() via
+        # run_on_gpu_pool_guarded, i.e. already on a gpu-pool worker. On a
+        # 1-worker pool (MPS) a second slot submit would queue behind this very
+        # job and self-deadlock (.result(timeout=10) fires, no sidecar spawns),
+        # so skip the slot when already on the pool. The off-pool callers (the
+        # engine self-test and the diagnostic probe) still take a slot as a
+        # queue wait, so they yield to an in-flight pool job instead of racing.
+        if not threading.current_thread().name.startswith("gpu-pool"):
+            # Lazy-import the GPU pool so importing this module doesn't pull in
+            # the entire model_manager + torch ecosystem at registry-listing time.
+            from services.model_manager import _get_gpu_pool
+            pool = _get_gpu_pool()
+            slot_future = pool.submit(lambda: None)
+            try:
+                slot_future.result(timeout=10)  # wait for our turn
+            except Exception:
+                slot_future.cancel()
+                raise
 
         try:
             with self._lock:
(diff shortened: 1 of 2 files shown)


paoloantinori:
CodeRabbit: addressed in 20a2379f. The test now monkeypatches _get_gpu_pool to a fresh 1-worker `gpu-pool`-prefixed pool, so the self-deadlock reproduces deterministically regardless of the ambient pool's worker count. Verified fail-before (TimeoutError at ~10s with the fix reverted) and pass-after.

paoloantinori:
Superseded by #1298, which folds the on-pool skip (this PR) together with the off-pool hold in one path-aware slot block. Closing in favor of the complete fix.

Issue #1259, opened 2026-07-25
[Crash] Backend died (signal 9)


xXTungNDXx:
## Environment

**Version:** `0.4.0`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.5-arm64-arm-64bit`
**Python:** `3.11.15`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M1 Max`
**RAM:** `64 GB`
**Disk free:** `514.7 GB`
**Active TTS engine:** `mlx-audio`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 0 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-07-25T06:22:44.000Z (8 min ago)
**Exit:** `signal 9`
**Uptime before crash:** 24376 s
**Backend version:** `0.4.0`

```
… (truncated)
ython-3.11.15-macos-aarch64-none/lib -L~/Library/Application Support/com.debpalash.omnivoice-studio/project/.venv/lib/python3.11/site-packages/torch/lib

Output:
clang++: error: no such file or directory: 'Support/com.debpalash.omnivoice-studio/project/.venv/lib/python3.11/site-packages/torch/lib'


Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

2026-07-25 08:22:32,980 INFO [omnivoice.dub] bench[generate] total=6349.38s tts=6317.34s cache=0.00s diskw=0.00s mix=0.96s save=0.26s segs=275
2026-07-25 13:18:59,237 INFO [omnivoice.api] Dub mux wrote ~/Library/Application Support/OmniVoice/dub_jobs/ocrqbvxq/exports/dubbed_video_20260725T131747-9715e350.mp4 (901890374 bytes)
2026-07-25 13:18:59,374 INFO [omnivoice.api] Native save wrote ~/Downloads/Video/dubbed_video.mp4 (901890374 bytes)
~/.local/share/uv/python/cpython-3.11.15-macos-aarch64-none/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
[...]

PR #1310, opened 2026-07-29
fix(compile): skip torch.compile when the torch lib path has whitespace

### LINKED ISSUE #1266: [Bug] torch.compile always fails on macOS: unquoted -L path through "Application Support"
## What

On macOS, `torch.compile`'s inductor backend passes the venv's torch lib directory to `clang++` as an unquoted `-L` flag. OmniVoice's default data directory lives under `~/Library/Application Support/`, which contains a space, so the argument splits and the compile always fails:

```
… -L~/Library/Application Support/com.debpalash.omnivoice-studio/project/.venv/lib/python3.11/site-packages/torch/lib

Output:
clang++: error: no such file or directory: 'Support/com.debpalash.omnivoice-studio/project/.venv/lib/python3.11/site-packages/torch/lib'
```

Surfaced in the log tail attached to #1259 (macOS 26.5, Apple Silicon, MPS).

## Impact

Low but non-zero, and it is guaranteed rather than intermittent:

- Every affected load pays a full compile attempt that **cannot** succeed, then falls back to eager mode.
- The failure is swallowed by `except Exception as e: logger.info("torch.compile skipped: %s", e)` (`backend/services/model_manager.py`), so it appears as ordinary log noise and drowns out the tail of genuinely useful crash logs — which is exactly how it showed up in #1259, where it was *not* the actual problem but consumed a chunk of the captured output.

Generation is unaffected: `torch.compile` is an optimization and eager mode is the documented fallback.

## Why it isn't caught today

`should_torch_compile()` in `backend/services/engine_env.py` gates on device, Triton availability, the Settings escape hatch, prior runtime failure, and CUDA arch — all reasonable, none of which look at the **path**. The quoting bug is inside PyTorch's inductor, so we can't fix it directly; but a path we already know cannot compile is one we should not attempt.

## Suggested fix

Add a path check to `should_torch_compile()`: if the resolved torch lib directory (or the venv root) contains whitespace, skip compile and log the reason once, in the same shape as the existing skip reasons. Keep `OMNIVOICE_FORCE_TORCH_COMPILE=1` as the override, consistent with the arch gate.

That turns a repeated guaranteed-failure into a single explanatory INFO line, and stops it consuming crash-report tails.

## Not doing here

Relocating the data directory away from `Application Support` — that is the correct macOS location and changing it would break every existing install (`omnivoice_data/` backward compatibility is a hard project constraint).

---

Found while triaging #1259; filing separately because it is a distinct defect from that report's SIGKILL.

debpalash:
Closes #1266.

## The defect

Inductor passes the torch library directory to `clang++`/`g++` as an **unquoted** `-L` flag. A path containing a space splits into two arguments:

```
-L~/Library/Application Support/…/site-packages/torch/lib
clang++: error: no such file or directory: 'Support/…/site-packages/torch/lib'
```

The quoting bug is inside PyTorch, so we can't fix it. What we control is not spending a compile attempt on a path we already know cannot link.

## Why it mattered

Never a broken generation — `torch.compile` is an optimization and eager mode is the documented fallback. The cost was a **guaranteed-failing** compile on every load whose clang wreckage was swallowed by `except Exception: logger.info(...)` and then ate a chunk of the captured log tail.

That's exactly how it surfaced: while triaging #1259 it was consuming crash-report output while **not** being the fault being investigated.

Not platform-specific either — macOS keeps app data under `~/Library/Application Support/`, and a Windows profile is routinely `C:/Users/First Last`.

## Behaviour

Slots in beside the existing gates in `should_torch_compile()` and follows their conventions:

- one INFO line naming the path and *why* it can't work, instead of a clang dump
- `OMNIVOICE_FORCE_TORCH_COMPILE=1` still overrides, consistent with the arch gate
- an unreadable torch path **fails open** — no evidence is not evidence of a problem

## Tests

7 cases: clean path compiles, spaced path skips *and says why*, force-env overrides, unreadable path fails open, and mps/cpu/xpu still short-circuit on the device gate first.

Fails before the change; `engine_env`/`torch_compile` suites stay green.

Updated `should_torch_compile()` to avoid guaranteed-failing `torch.compile` when the resolved PyTorch library path contains whitespace, adding an INFO/skip reason with redacted path details and preserving the `OMNIVOICE_FORCE_TORCH_COMPILE=1` override. Added `_torch_lib_path_is_linkable()` with fail-open behavior when path probing/import errors occur, and expanded tests to cover clean vs spaced paths, override and log-message escape-hatch naming, unreadable path behavior, and non-CUDA device short-circuits (including asserting home directory paths are not logged verbatim). Main risk to review: enabling the force override on whitespace-containing paths will still attempt compilation and may fail as before.

### DIFF
--- a/backend/services/engine_env.py
+++ b/backend/services/engine_env.py
@@ -121,8 +121,18 @@ def _torch_lib_path_is_linkable() -> tuple[bool, str]:
     except Exception:
         return True, ""
     if any(ch.isspace() for ch in lib_dir):
+        # The path is genuinely useful for diagnosis, but it contains the user's
+        # home directory and this string is logged and lands in pasted bug
+        # reports — so it goes through the same redaction every other
+        # user-facing failure text uses (home → ~, secrets stripped).
+        try:
+            from core.failure import sanitize
+
+            shown = sanitize(lib_dir)
+        except Exception:
+            shown = os.path.basename(lib_dir.rstrip(os.sep)) or "<torch lib>"
         return False, (
-            f"the torch library path contains whitespace ({lib_dir!r}); inductor "
+            f"the torch library path contains whitespace ({shown!r}); inductor "
             f"passes it to the C++ linker unquoted, so every compile attempt fails"
         )
     return True, ""
--- a/backend/services/engine_env.py
+++ b/backend/services/engine_env.py
@@ -98,6 +98,36 @@ def _cuda_arch_supported_for_compile() -> "tuple[bool, str]":
         return True, ""
 
 
+def _torch_lib_path_is_linkable() -> tuple[bool, str]:
+    """``(ok, reason)`` — False when inductor's C++ link step is guaranteed to
+    fail because the torch library path contains whitespace (#1266).
+
+    Inductor passes the torch lib directory to ``clang++``/``g++`` as an
+    unquoted ``-L`` flag. A path with a space splits into two arguments and the
+    compile dies with ``no such file or directory: 'Support/...'``. The bug is
+    inside PyTorch, so we cannot fix the quoting — but a path we already know
+    cannot compile is one we should not spend a compile attempt on.
+
+    It is not hypothetical on any platform: the macOS data dir lives under
+    ``~/Library/Application Support/``, and a Windows user profile is routinely
+    ``C:/Users/First Last``.
+
+    Never raises — an unreadable torch path means "no reason to skip".
+    """
+    try:
+        import torch
+
+        lib_dir = os.path.join(os.path.dirname(torch.__file__), "lib")
+    except Exception:
+        return True, ""
+    if any(ch.isspace() for ch in lib_dir):
+        return False, (
+            f"the torch library path contains whitespace ({lib_dir!r}); inductor "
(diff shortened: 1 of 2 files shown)


debpalash:
Addressed the Major and both test findings. **Declining the Critical**, with reasoning:

> Keep the default compile policy cross-platform […] make the default policy uniform and expose compilation as an explicit opt-in

The rule it cites is about **user-visible default behaviour**. `torch.compile` is not user-visible — it is an internal optimization whose absence shows up only as speed, and CLAUDE.md's own wording allows platform-specific *implementation*.

More importantly, `should_torch_compile()` has always diverged by host, deliberately:

- `device != "cuda"` → skip (every MPS and CPU install)
- Triton not importable → skip (**every Windows install** — there is no Windows wheel)
- GPU arch not in the torch build's arch list → skip

This gate is the same shape as that last one: it declines a compile that is **guaranteed to fail** on this host. It adds no new divergence — a host with a space in its torch path could never have compiled successfully.

Following the suggestion would disable a working optimization for every Linux CUDA user in order to match the behaviour of hosts that cannot compile at all. That is the regression, not the fix.

Happy to revisit if you can point at a case where the gate skips a compile that would otherwise have **succeeded**.

Issue #1306, opened 2026-07-29
Proposal: add Kyutai's PocketTTS as a low-latency, CPU-only TTS engine option


paoloantinori:
## Proposal

Add [PocketTTS](https://github.com/kyutai-labs/pocket-tts) (Kyutai Labs) as a TTS engine option in OmniVoice, alongside the quality engines (omnivoice, IndexTTS, Supertonic-3, etc.).

## Why: latency (the interesting part)

The standout is **latency**. PocketTTS is purpose-built for it:

- **100M parameters** (vs OmniVoice's multi-GB resident model).
- **~200ms to the first audio chunk**, with streaming.
- **~6x real-time on a MacBook Air M4 CPU** (2 cores, no GPU).
- Runs entirely on CPU, so zero VRAM contention.

For latency-sensitive or low-resource scenarios (notifications, reaction-triggered voice, real-time feedback, CPU-only / edge devices) this fills a gap none of the current engines hit: near-instant, lightweight TTS where "good enough, very fast" beats "excellent, slow and heavy." It is the complementary opposite end of the spectrum from the resident OmniVoice model.

## Real-world signal: a two-way routing we already run

We (a downstream user) already run a **custom two-way routing** between OmniVoice and PocketTTS in our agents: OmniVoice for quality when latency is not critical and the GPU is free, PocketTTS for speed when low latency matters or the GPU is contended or unavailable. This is a production workaround for exactly the latency + resource gap, not hypothetical. Natively supporting PocketTTS as an OmniVoice engine would let this routing live in OmniVoice's engine picker (one process, one config, the existing engine-routing + GPU-preflight machinery) instead of a bespoke agent-level shim, and would benefit every user with the same latency/resource tradeoff.

## Left to the maintainer's evaluation

This is a proposal, not a request to merge anything. The call on whether to add it is entirely yours:

- **Languages**: PocketTTS currently covers 6 (en/fr/de/pt/it/es) vs OmniVoice's 600+, so it is a narrower but much faster niche.
- **Quality**: voice cloning + streaming, but a 100M model will not match the heavy engines on expressiveness.
- **License / fit / maintenance**: a Kyutai Labs project ([paper](https://arxiv.org/abs/2509.06926), [model card](https://huggingface.co/kyutai/pocket-tts)); license and dep-pin compatibility are for the repo to decide.

[...]

PR #1322, opened 2026-07-30
bench(engines): add CPU latency harness (PocketTTS evaluation, #1306)

paoloantinori:
## What

Adds `scripts/bench_engines_latency.py`, a reproducible CPU latency benchmark for the TTS engines (time-to-first-chunk and real-time factor). Built for the PocketTTS evaluation in #1306, where measured numbers and a way to verify them were requested.

**Data and methodology only.** This is not the PocketTTS engine integration: no adapter, sidecar, or CI smoke. It is a neutral measurement tool. Whether it belongs in-tree is your call.

## How it works

Mirrors `scripts/bench_pipeline.py`'s memory discipline: one engine at a time, model unloaded between engines, RAM-floor check (skips below the floor, and skips loudly when free RAM is unmeasurable rather than running blind). Each engine is driven through its own raw synthesis path, so no engine is penalised or flattered by OmniVoice's subprocess plumbing:

- **PocketTTS** via the `pocket_tts` library (`generate_audio` / `generate_audio_stream`).
- **supertonic3** via the raw SDK path used by `backend/engines/supertonic3/sidecar.py`.
- **omnivoice-gguf** via the in-tree `OmniVoiceGGUFBackend`.

Engines not installed in the running environment skip with a clear reason, so the same script runs cleanly wherever each engine is available.

Metrics: cold (load + first synth), warm (median of 3 passes), TTFC (first streamed chunk; batch engines marked N/A since they emit nothing incrementally), RTF (audio_s / synth_s). Records CPU, cores, RAM, OS.

```
uv run python scripts/bench_engines_latency.py
```

## Measured (Apple M3 Pro, 12 cores, CPU-only; run-to-run within ~10%)

| engine | warm RTF (short / long) | TTFC |
|---|---|---|
| PocketTTS 2.1.0 | ~8-9x / ~8-9x | ~33 ms (first chunk, streaming) |
| supertonic3 1.3.1 | ~4.3x / ~5.0x | N/A (batch; ~1 s full output) |
| omnivoice-gguf | pending | pending (needs the native binary built) |

Full methodology, fairness caveats, and the licence answer are in #1306.

Draft because this is shared for the evaluation, not a push to merge. Happy to mark ready if you want it in-tree.

Adds a standalone, reproducible CPU latency benchmark for PocketTTS, supertonic3, and omnivoice-gguf, measuring cold/warm synthesis, TTFC, RTF, hardware details, and memory-safe sequential execution. It improves cross-engine performance measurement without adding engine integration or CI changes. Review the raw engine paths and availability/memory handling for environment-specific behavior; unavailable engines are skipped with reasons.

### DIFF
--- a/scripts/bench_engines_latency.py
+++ b/scripts/bench_engines_latency.py
@@ -1,0 +1,340 @@
+#!/usr/bin/env python
+"""CPU latency benchmark for PocketTTS vs the OmniVoice incumbent TTS engines.
+
+DATA + METHODOLOGY ONLY. This measures each engine's raw synthesis speed on
+CPU (time-to-first-chunk and real-time factor), the numbers the OmniVoice
+maintainer asked for in #1306 before evaluating PocketTTS as an engine. It is
+NOT an integration: it builds no PocketTTS adapter, sidecar, or CI smoke. Each
+engine is driven through its OWN raw synthesis path, so no engine is penalised
+or flattered by OmniVoice's subprocess plumbing, and the same script reproduces
+the numbers on the maintainer's machine.
+
+Mirrors scripts/bench_pipeline.py's memory discipline:
+  * engines run one at a time, never concurrently;
+  * each model is dropped + gc'd between engines, so peak RSS is one model;
+  * before each engine we check free RAM and SKIP (not crash) below the floor
+    (and SKIP loudly when free RAM cannot be measured, rather than run blind);
+  * a small fixed number of warm passes, median reported, no loops to convergence.
+
+Metrics (all on CPU):
+  * cold  = model load + voice setup (paid once, at startup);
+  * warm  = median of N synthesize passes with the model resident;
+  * TTFC  = time to first playable audio. Reported ONLY for the streaming engine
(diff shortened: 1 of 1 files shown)


debpalash:
Happy to have this in-tree — full response on the evaluation is on #1306. Short version: licence cleared, numbers accepted, and a reproducible harness is worth more than the one set of numbers it produced.

One thing to fix before it goes out of draft:

The branch includes three files that look like your local task tracker rather than part of the harness:

- `backlog/config.yml`
- `backlog/tasks/task-1 - Investigate-OmniVoice-VRAM-lifecycle-stuck-model-loads.md`
- `backlog/tasks/task-2 - Exploration-durable-fixes-for-OmniVoice-VRAM-pressure.md`

`backlog/` doesn't exist in the repo and isn't gitignored, so merging would create it. Please drop them from the branch — `scripts/bench_engines_latency.py` is the whole change.

I'm adding `backlog/` to `.gitignore` on `main` so it can't catch you (or anyone else running that tool) again.

Otherwise: mark it ready whenever you like, and I'll fill the `omnivoice-gguf` row.

PR #1327, opened 2026-07-30
fix(release): rebuild + cryptographically verify the preview updater manifest (macOS Preview updates fail signature verification)

Pinkers01:
## Problem

macOS users on the **Preview** update channel get **"The signature verification failed"** on every update attempt.

**Root cause chain (verified against the live release):**

1. Since ~2026-07-13, **every** nightly matrix leg logs
   `Signature not found for the updater JSON. Skipping upload...`
   (e.g. run 30523998127, 2026-07-30, all 4 legs) — tauri-action uploads the bundles and their `.sig` companions but **never refreshes `latest.json`**. The preview manifest has been frozen at `2026-07-13T10:32:59Z` (version `0.3.21-86`).
2. The (otherwise correct) `Clear this arch's stale preview updater bundle` step deletes + replaces the **version-less** macOS updater bundles (`OmniVoice.Studio_aarch64.app.tar.gz`, `..._x64.app.tar.gz`) every night — most recently 2026-07-30 07:56/08:03 UTC.
3. Result: the manifest's darwin signatures point at files that no longer exist in that form. Linux/Windows entries still verify because their filenames are versioned (`..._0.3.21-86_...`) and the old files survive on the rolling release.

Independent verification against the updater pubkey in `tauri.conf.json` (blake2b-512 + ed25519, exactly what the bundled minisign check does):

```
FAIL darwin-aarch64:      signature in latest.json does NOT match OmniVoice.Studio_aarch64.app.tar.gz
FAIL darwin-aarch64-app:  signature in latest.json does NOT match OmniVoice.Studio_aarch64.app.tar.gz
FAIL darwin-x86_64:       signature in latest.json does NOT match OmniVoice.Studio_x64.app.tar.gz
FAIL darwin-x86_64-app:   signature in latest.json does NOT match OmniVoice.Studio_x64.app.tar.gz
OK   linux-x86_64         (versioned filename, old file still present)
OK   windows-x86_64       (versioned filename, old file still present)
```

The existing `Verify preview updater manifest` step checks prerelease flag, version format and platform parity — all of which kept passing while every darwin entry was cryptographically unverifiable.

## Fix (both in the single post-matrix `preview-notes` job — no per-leg race)

1. **Rebuild `latest.json` from the release's real assets** and their `.sig` companions, then `--clobber` upload. The manifest can no longer drift from the files it describes, regardless of what tauri-action's updater-JSON path does or skips. This also repairs the currently-broken manifest on the first nightly after merge.
2. **Extend the verify step with a cryptographic check**: every signature in `latest.json` must verify (minisign file sig + trusted-comment sig) against the artifact it points at, using the pubkey from `tauri.conf.json`. Any future drift fails the run loudly instead of shipping a broken channel.

Happy to adjust scope/style. The underlying "Signature not found for the updater JSON" from tauri-action is worth a separate look (it started around 2026-07-13), but with the manifest rebuilt post-matrix the channel stays correct either way.

Rebuilt the macOS Preview `latest.json` after the release matrix from current artifacts and signatures, uploading it with `--clobber`, and added cryptographic verification against the updater public key. This prevents stale signature mismatches caused by nightly replacement of version-less bundles. Human review should focus on artifact selection and signature validation across all platforms.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -919,6 +919,87 @@ jobs:
     permissions:
       contents: write
     steps:
+      # Needed by the manifest rebuild + signature check below: the updater
+      # pubkey lives in frontend/src-tauri/tauri.conf.json.
+      - uses: actions/checkout@v4
+
+      # ── Rebuild the preview updater manifest from what is ACTUALLY published ──
+      # Since ~2026-07-13 every matrix leg has logged "Signature not found for
+      # the updater JSON. Skipping upload..." — tauri-action uploads the bundles
+      # + .sig companions but never refreshes latest.json. Meanwhile the macOS
+      # updater bundles (version-less filenames) are deleted + replaced every
+      # night by "Clear this arch's stale preview updater bundle", so the
+      # manifest's darwin signatures stopped matching the published files:
+      # macOS Preview users hit "The signature verification failed" on every
+      # update attempt (latest.json frozen at 2026-07-13, tar.gz replaced
+      # nightly).
+      #
+      # Root fix: after the matrix completes, rebuild latest.json HERE — one
+      # job, no per-leg race — from the release's real assets and their .sig
+      # companions, then clobber-upload. The manifest can no longer drift from
+      # the files it describes, regardless of what tauri-action's own
+      # updater-JSON path does or skips.
+      - name: Rebuild preview updater manifest from published assets
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          gh release view preview --repo "$REPO" --json assets \
+            -q '[.assets[].name]' > /tmp/assets.json
+          WORK=$(mktemp -d)
+          python3 - "$WORK" <<'PY'
+          import datetime, json, os, re, subprocess, sys
+          work = sys.argv[1]
+          repo = os.environ["REPO"]
+          names = json.load(open("/tmp/assets.json"))
+
+          def sig(name):
+              subprocess.run(["gh", "release", "download", "preview", "--repo", repo,
+                              "-p", name + ".sig", "-D", work], check=True)
+              return open(os.path.join(work, name + ".sig")).read().strip()
+
+          # Versioned artifacts accumulate on the rolling release; the newest
+          # BASE-N (N = run number, monotonically increasing) is current.
+          def newest(pattern):
+              best, best_n = None, -1
+              for n in names:
+                  m = re.fullmatch(pattern, n)
+                  if m and int(m.group("n")) > best_n:
+                      best, best_n = n, int(m.group("n"))
+              return best, best_n
+
+          appimage, n1 = newest(r"OmniVoice\.Studio_[\d.]+-(?P<n>\d+)_amd64\.AppImage")
+          msi, n2 = newest(r"OmniVoice\.Studio_[\d.]+-(?P<n>\d+)_x64_en-US\.msi")
+          assert appimage and msi, f"missing versioned updater artifacts (AppImage={appimage}, MSI={msi})"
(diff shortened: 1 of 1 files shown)


debpalash:
Reviewed — this is a real root-cause fix for a genuinely nasty failure (a manifest silently describing files that no longer exist), and rebuilding it once after the matrix instead of per-leg is the right shape. Thanks @Pinkers01!

Three findings to resolve before merge. All three are fixed and tested on `review/1327-manifest-run-consistency` in this repo — I could not push to your fork despite `maintainerCanModify`, so please cherry-pick `66217b7f` onto `fix/preview-updater-manifest` (or say the word and I will land it as a follow-up commit crediting you):

**1. The manifest can still describe artifacts from two different runs (blocking).** `newest()` picks the AppImage and the MSI independently, then `appimage if n1 >= n2 else msi` takes whichever run number is larger as *the* version. If one matrix leg fails or is re-run, that publishes a manifest advertising `X.Y.Z-5` while pointing Windows users at the `-4` MSI — the same manifest/artifact drift this job exists to end, reintroduced by the fix for it. Fixed by requiring `n1 == n2` and failing loudly: leaving the previous manifest in place is a visible, already-understood state, whereas shipping a mismatched one is not. (The darwin tarballs carry no run number, so they cannot be checked this way — noted in a comment, since the signature verification in the next step is what pins those to the published bytes.)

**2. `persist-credentials: false` on the checkout.** Nothing in this job pushes to git, and the steps after it shell out to `gh` and install from PyPI.

**3. Floor-pin the `cryptography` install** (`>=42`, matching `docs-drift.yml`'s `pyyaml>=6`). This step is the gate that decides whether a signed release manifest is trustworthy — the one dependency here worth a bound.

Also added `tests/test_release_preview_manifest_rebuild.py`: it extracts this step's body from `release.yml` (so it cannot drift from the workflow) and runs it against a stubbed `gh` with synthetic asset lists — matched run numbers produce the expected manifest, mismatched ones fail and upload nothing, and a missing `.sig` companion or a missing darwin tarball each fail rather than silently dropping a platform. The mismatch tests fail against the current branch and pass with the fix.

One thing I checked that is correct as written: the verify step re-downloads `latest.json` over HTTP *after* the rebuild, so it validates the new manifest rather than a cached one.

debpalash:
@Pinkers01 — I have opened #1362, which carries **your commit with authorship intact** plus the three review fixes from my comment above. I could not push to your fork despite `maintainerCanModify`, and this had gone six hours, so rather than leave a real two-week outage fix waiting I unblocked it that way.

**The credit is yours** — the commit, the changelog entry and the PR description all say so. If you would rather push the fixes to your own branch and land this PR instead, say the word and I will close #1362 in favour of it. Genuinely happy either way; the diagnosis here was excellent and the "rebuild once after the matrix" shape is the right one.

Leaving this open until one of the two merges.

debpalash:
Superseded by #1362, which is now merged — it carries @Pinkers01's commit with authorship intact, plus the five review fixes that came out of the two bot passes and my own review.

Thank you @Pinkers01. The diagnosis here was the hard part: two weeks of macOS Preview updates failing signature verification, with CI green the whole time because parity and version-format checks never look at a signature. Rebuilding the manifest once after the matrix, from the assets actually published, is the right shape and it is what landed.

What changed on top of your commit, for the record:
- both versioned artifacts must come from the same run (a failed leg could otherwise advertise a version describing only half the payload);
- the macOS bundles are tied to the run by upload time — signature verification cannot catch a stale tarball, since it and its stale `.sig` match each other perfectly;
- verification runs *before* the upload, so a manifest that fails the check is never published;
- `persist-credentials: false` on the checkout, and a floor-pinned `cryptography`;
- the selection rules moved into `scripts/build_preview_manifest.py` with 14 tests, since three findings in a row were about which artifacts may be described together.

This takes effect on the next preview build.

PR #1328, opened 2026-07-30
feat(engines): PocketTTS CPU-only sidecar shape (#1306)

paoloantinori:
Draft: sidecar **shape** for review, per your "I'd rather review the sidecar shape before you build on top of it" (#1306). Not ready to merge: no CI smoke, no gated-weights preflight, no 4-platform install verification, no licence-accept gate yet. Those land on top once the shape is agreed.

## What
Adds PocketTTS (kyutai-labs/pocket-tts, 100M params) as an opt-in, CPU-only TTS engine for the "fastest CPU render / lowest latency" row, with **six languages** (en/fr/de/pt/it/es). Measured numbers and the licence answer are in #1306; the reproducible harness is #1322.

## Shape (mirrors the existing subprocess engines)
- `backend/engines/pockettts/__init__.py`: `PocketTTSBackend(SubprocessBackend)`. CPU-only (`gpu_compat=("cpu",)`), parent interpreter (pocket-tts deps sit at the host pins, same rationale as omnivoice-subprocess), `supports_cloning`, optional-dep `is_available()` gate. No `generate()` override; the base forwards JSON-safe kwargs (`text`, `ref_audio`, `language`, ...).
- `backend/engines/pockettts/main.py`: stdio sidecar (`ready` / `ping` / `synthesize` / `shutdown`). Lazy `TTSModel.load_model(language=...)` per language (cached), maps OmniVoice's `language` value to a pocket-tts model language, picks the default preset voice per language when no reference clip is given, `generate_audio` to int16 PCM. Byte-identical wire protocol to omnivoice-subprocess / supertonic3.
- Registered in `services/tts_backend.py` (`_LAZY_REGISTRY` + `_INSTALL_HINTS`).

## Language support: six, not one
PocketTTS ships one model per language: english, french, german, portuguese, italian, spanish (selectable via `load_model(language=...)`). The HuggingFace model card's "English only at the moment" line is stale; the [GitHub README](https://github.com/kyutai-labs/pocket-tts) and pocket-tts 2.1.0 confirm six. The sidecar cold-loads and caches a model the first time a language is used.

## One shape question for you
PocketTTS streams (`generate_audio_stream`, ~33 ms to first audio). This sidecar is **batch** (one audio frame per synth, the SubprocessBackend contract), so end-to-end TTFC is full-synth time (~0.3 s for a short line), still well under supertonic3's ~1 s, but the streaming advantage is lost. A streaming-aware protocol (incremental audio frames) would recover it. I kept this batch to match every other subprocess engine; happy to take it streaming if you'd rather.

## Deferred (on top of this shape, after your review)
- **Condition 6**: typed "weights are gated" preflight in `core/failure.py` (`POCKETTTS_GATED_WEIGHTS`), following `HF_AUTH_FAILED` / `PYANNOTE_LICENSE_REQUIRED`.
- **Condition 3**: stubbed CI smoke (no GPU, no gated pull), pattern `test_omnivoice_subprocess.py`.
- **Condition 4**: four-platform install verification.
- Licence-accept gate (MIT code + CC-BY-4.0 weights), like supertonic3's TTS-05.

Adds an opt-in, CPU-only `PocketTTSBackend` with a batch stdio sidecar for six languages, lazy model loading, voice handling, and Int16 PCM output. Registers the backend and documents it in the TTS inventory. Before merge, review the multi-dimensional audio validation fix and complete CI smoke coverage, gated-weight preflight, four-platform installation checks, licence-acceptance gating, and the `Unreleased` changelog entry.

### DIFF
--- a/backend/engines/pockettts/__init__.py
+++ b/backend/engines/pockettts/__init__.py
@@ -1,0 +1,96 @@
+"""pockettts: Kyutai PocketTTS as a crash-isolated, CPU-only TTS sidecar (#1306).
+
+PocketTTS (kyutai-labs/pocket-tts, 100M params) is hired for the "fastest CPU
+render / lowest latency" job, the row the engine-acceptance framework leaves
+unheld: every CPU engine OmniVoice ships is either English-only or a quality
+engine falling back to CPU. PocketTTS is the complementary opposite end of the
+spectrum from the quality engines (omnivoice, IndexTTS, Supertonic-3): small,
+fast, CPU-only, zero-shot cloning from a reference clip. Measured ~8-9x
+real-time on an Apple M3 Pro (see scripts/bench_engines_latency.py, PR #1322).
+
+This engine runs PocketTTS in a child process via :class:`SubprocessBackend`,
+mirroring engines/omnivoice_subprocess and engines/supertonic3. Crash isolation:
+a wedged generate is hard-killed by the parent's watchdog, reclaiming the
+child's memory, the thing an in-process engine structurally cannot do. CPU-only
+by design (``gpu_compat = ("cpu",)``): Kyutai observes no GPU speedup for this
+100M, batch-1 model.
+
+Opt-in (Settings -> Engines, or ``OMNIVOICE_TTS_BACKEND=pockettts``); the
+default ``omnivoice`` engine is unchanged, so existing users see no behaviour
+change.
+
+Licence: MIT (code) + CC-BY-4.0 (weights), both commercial-OK (cleared from
+primary sources in #1306). The weights are gated on HuggingFace (an access
+agreement plus an acceptable-use clause); the engine must surface that honestly
+at first-run rather than failing inside a download (condition 6 of the #1306
+acceptance). That preflight is built on top of this shape, not in it.
+
+Streaming note: PocketTTS streams audio (``generate_audio_stream``), but this
+batch sidecar returns one audio frame per synth, matching the SubprocessBackend
+contract every other subprocess engine uses. A streaming-aware variant
+(incremental audio frames) is a documented opportunity to recover PocketTTS's
+~33 ms time-to-first-audio end-to-end; out of scope for this shape, raised on
+the PR.
+"""
+from __future__ import annotations
+
+import logging
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from services.subprocess_backend import SubprocessBackend
+
+if TYPE_CHECKING:
+    import torch  # noqa: F401
+
+logger = logging.getLogger("omnivoice.pockettts")
+
+
+class PocketTTSBackend(SubprocessBackend):
+    """Kyutai PocketTTS in a killable, CPU-only sidecar process."""
+
+    id = "pockettts"
+    display_name = "PocketTTS (Kyutai, CPU-only, low-latency, MIT/CC-BY-4.0)"
+    _DEFAULT_SAMPLE_RATE = 24_000
+    # CPU-only by design (honest hardware reporting, like supertonic3): Kyutai
+    # ships no CUDA/MPS path and reports no GPU speedup for this model.
+    gpu_compat: tuple[str, ...] = ("cpu",)
+    supports_cloning = True  # zero-shot clone from a reference clip
+
+    @classmethod
+    def is_available(cls) -> tuple[bool, str]:
+        # Optional-dep gate: the pocket-tts wheel is installed only when the user
+        # opted in. The interpreter is the parent's own (sys.executable), so
+        # there is no separate venv to validate.
+        try:
+            import pocket_tts  # type: ignore[import-not-found]  # noqa: F401
+        except Exception as e:
+            return False, (
+                "pocket_tts package not installed. Enable in Settings -> Engines "
+                "(pip install pocket-tts)."
+            )
+        return True, "ready (CPU-only)"
+
+    @classmethod
+    def venv_python(cls) -> Path:
+        # Parent interpreter: pocket-tts deps (torch>=2.5, scipy, beartype) sit
+        # happily at the parent's pins, so this isolates for crash recovery, not
+        # dependency pins (same rationale as omnivoice-subprocess).
+        return Path(sys.executable)
+
+    @classmethod
+    def sidecar_script(cls) -> Path:
+        return Path(__file__).resolve().parent / "main.py"
+
+    @property
+    def sample_rate(self) -> int:
+        return self._DEFAULT_SAMPLE_RATE
+
+    @property
+    def supported_languages(self) -> list[str]:
+        # Protocol tag; per-engine language coverage follows the model card.
+        return ["multi"]
+
+
+__all__ = ["PocketTTSBackend"]
--- a/backend/engines/pockettts/__init__.py
+++ b/backend/engines/pockettts/__init__.py
@@ -5,8 +5,10 @@ render / lowest latency" job, the row the engine-acceptance framework leaves
 unheld: every CPU engine OmniVoice ships is either English-only or a quality
 engine falling back to CPU. PocketTTS is the complementary opposite end of the
 spectrum from the quality engines (omnivoice, IndexTTS, Supertonic-3): small,
-fast, CPU-only, zero-shot cloning from a reference clip. Measured ~8-9x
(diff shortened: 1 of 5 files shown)


paoloantinori:
Hi, gentle nudge. I've flipped this to ready-for-review in case the draft flag made it look like work-in-progress: it's intentionally just the sidecar shape, per your "review the shape before you build on top." The CI smoke, gated-weights preflight, four-platform install, and licence-accept gate all build on top once you've eyeballed the shape.

One concrete call I'd value before I add the rest: the sidecar is batch (one audio frame per synth, the SubprocessBackend contract), so PocketTTS's streaming TTFC (~33 ms) is lost end-to-end. A streaming-aware protocol would recover it, but diverges from every other subprocess engine. I kept it batch to match; happy to take it streaming if you'd rather.

No rush.

debpalash:
Shape review — the answer you asked for, plus one code finding still open.

## Keep it batch. Ship this shape.

**Keep batch.** The right fix for the lost TTFC is a streaming-aware `SubprocessBackend` protocol that *every* sidecar can opt into, not a second wire contract that only PocketTTS speaks. Forking the protocol for one engine buys ~250 ms on one engine and costs a permanent divergence in the layer every future engine has to be written against — and the next contributor has to guess which of the two shapes to copy. Your instinct to match the existing engines was right.

The streaming work is welcome as its own change, judged on its own merits, with the base class and all three existing sidecars moving together. That is a bigger and more interesting PR than this one, and it should not be gated on PocketTTS landing.

Also: at ~0.3 s full-synth for a short line, this already wins the "fastest CPU render" row it is being added for. The 33 ms number matters for real-time dictation, which is not the job this engine is being brought in to do.

## Shape itself: approved

It mirrors the existing subprocess engines closely enough that I have nothing structural to add. Specifically checked and correct:

- CPU-only via `gpu_compat=("cpu",)` and the parent interpreter, same rationale as omnivoice-subprocess.
- Opt-in with an `is_available()` optional-dep gate, so it satisfies the cross-platform rule by construction — nothing changes for a user who never enables it.
- No `generate()` override; the base's JSON-safe kwarg forwarding is doing the work.
- Byte-identical wire protocol to omnivoice-subprocess / supertonic3.

Good calls on the iterations since the first review, all of which I would otherwise have raised: raising on an unsupported language instead of mispronouncing it in English, the cold-load heartbeat so a gated-weights download doesn't get a healthy sidecar killed, the LRU bound on the voice cache, `st_mtime_ns` in the fingerprint, and rejecting URLs in `ref_audio`.

## One finding still open

`_tensor_to_pcm_b64`, the defensive downmix (CodeRabbit flagged this and it is still live):

```python
arr = np.asarray(audio, dtype=np.float32).squeeze()
if arr.ndim > 1:
    arr = arr.mean(axis=0)   # assumes channels-first
```

For a channels-last `(N, 2)` array, `squeeze()` keeps both axes and `mean(axis=0)` averages across *time* — `n_samples` becomes 2 and the frame is garbage, not a downmix. Since this branch only exists to be defensive about a shape you don't expect, it should not itself be able to produce silent noise. Either pick the channel axis as the shorter dimension, or drop the guesswork and raise on `ndim > 1` so an upstream shape change surfaces as an error frame. I'd take the raise: this engine returns mono today, so the branch is unreachable in practice, and an assertion that fails loudly is worth more than a downmix that quietly might not be one.

## Then the deferred list

Once that's in, the four you named — CI smoke, gated-weights preflight, four-platform install verification, licence-accept gate — are what stand between this and merge. Two notes so they don't need a second round:

- **Licence gate:** the local-first guarantee means the accept has to be an explicit user action with a real decline path, and declining must leave the rest of the app fully functional. No default-on, no "continue implies accept".
- **CHANGELOG:** an `### Added` one-liner under `## [Unreleased]` with the `(#1306)` ref and your credit, per the repo's changelog rule.

Thanks for the patience on the shape question, and for pushing the fixes rather than arguing them.

debpalash:
All three fixes look right, and the framing one in particular — the `stop.set()` / `join()` pair made the window narrow rather than closed, so the lock is the correct shape.

One thing left before I merge: **there are no tests.** 428 lines, and the comparable sidecars both have suites (`tests/test_supertonic3.py`, `tests/test_confucius4_sidecar.py`).

That matters more than usual here because every finding you fixed this morning is silent by construction:

| fix | what it looked like without the fix |
|---|---|
| unsupported language | fluent, confident, **wrong** audio — English phonemes on Italian text |
| channels-last downmix | averages across **time**, not channels — noise, not a downmix |
| `_send` lock | a length header lands inside another frame's body; the pipe never resynchronizes |
| mtime+size fingerprint | re-recording the clip changes nothing, forever |

None of them raise. A smoke test that asserts audio came back passes through all four.

I've written them rather than asking you to: **paoloantinori/OmniVoice-Studio#1**, opened against your branch so this PR stays yours. 49 tests over the pure logic (language selection, PCM, wire framing, the LRU cache) and the backend surface (recv-timeout guards, CPU-only declaration, sample-rate lockstep, lazy registration). The model is mocked and the sidecar is stdlib-only at import time, so nothing needs the `pocket-tts` wheel or a child process.

Verified fail-before/pass-after by reverting your lock and multi-channel guard — the framing test fails with a length header decoded from inside another frame's body, which is the corruption itself rather than a proxy for it.

Merge that and I'll merge this. Your four deferred items (CI smoke, gated-weights preflight, four-platform install, licence-accept gate) stay deferred as agreed — this is the shape PR.

Separately, and **not** your problem: `mean(axis=0)` with the same channels-first assumption is in five already-merged sidecars (`moss_tts_v15`, `dots_tts`, `supertonic3`, `confucius4`, `omnivoice_subprocess`). Latent in all of them since they return mono, but it's the same class. I'll fix those on main — no scope creep here.

debpalash:
Tests merged, thanks. CI clean, 90 passing locally on the head. Merging.

On greptile's last P1 (cold load exceeds the 300s generation budget) — **real, but not yours, and not a blocker here.**

Every sidecar loads its model inside `_handle_synthesize`, not before it:

| sidecar | loader | weights |
|---|---|---|
| `supertonic3` | `_load_tts` | ~400 MB |
| `confucius4` | `_load_model` | ~5 GB |
| `dots_tts` | `_load_runtime` | ~9 GB |
| `moss_tts_v15` | `_load_model` | ~16 GB |
| `pockettts` | `_load_model` | **~100 MB** |

So the shape is shared by all five, and PocketTTS is the *least* exposed of them — it is the smallest model in the table by two orders of magnitude. Asking this PR to fix a condition every existing sidecar already has, and has less of, would be the wrong place for it. Its suggested remedy also names an `ensure_ready()` that `SubprocessBackend` does not have.

I've filed it as its own issue against the shared path, where the real risk lives (a 9 GB first download on a slow line blows the 300s budget having already fetched most of it — the outer job times out even though the sidecar's own watchdog is healthy and progress frames are arriving).

Your four deferred items stand as agreed: CI smoke, gated-weights preflight, four-platform install verification, licence-accept gate.

debpalash:
Filed as #1367 — with the finding that it is recoverable (the pool reset reclaims nothing by design, so the sidecar keeps downloading and a retry succeeds), which bounds it to a first-run message problem rather than a block. Thanks for the engine.

paoloantinori:
While this sits in review, one upstream fact worth folding into the engine's docs (and possibly the adapter): **kyutai won't add number normalization**. I asked ([kyutai-labs/pocket-tts#234](https://github.com/kyutai-labs/pocket-tts/issues/234)) and they closed it with "run your own text preprocessing; the model is too small to read numbers." Fair for a library — but VoiceStudio is a GUI app: users type text with numbers and there is **no caller side** to push preprocessing to.

Two suggestions, in increasing size:

1. **Engine guide caveat (docs-only, now)**: the pockettts guide under `docs/engines/` should note that the model reads raw digits poorly, especially in non-English languages, and that callers who use the API directly should pre-normalize numbers to words.
2. **Opt-in number normalization in the adapter (feature, your call on shape)**: a per-language pre-pass before tokenization (e.g. `num2words`, opt-in via engine setting, default off for English-only users). I have a working reference implementation running in production (locale-aware `it`/`en` parsing, regex boundaries that leave dates `2026-08-17`, times `12:30`, and versions `v2.0.1` untouched, safe degradation when the dep is missing) — happy to vendor it into the adapter here or as a fast-follow PR, whatever fits the engine shape you prefer.

Not blocking for #1328 itself — the engine works as-is; this is about what users hear when they type "il totale è 1.234.567 euro".

Issue #1333, opened 2026-07-30
[Bug] Linux AppImage cannot enumerate microphones while raw binary works


Kakuzen93:
> **AI disclosure:** This report was written and submitted with assistance from OpenAI Codex after interactive troubleshooting with the affected user. The commands and results below were observed on the user’s machine.

## What happened?

On Linux, the locally built AppImage launched by `bun run desktop-prod:run` cannot find any microphone. Starting a recording shows:

> No microphone found. Connect or enable a microphone and try again.

The frontend maps this to `getUserMedia()` rejecting with `NotFoundError`. Launching `frontend/src-tauri/target/debug/omnivoice-studio` directly from the same build immediately fixes recording.

## Steps to reproduce

1. On Linux, build with `bun run desktop-prod`.
2. Relaunch with `bun run desktop-prod:run`; the launcher prefers `target/debug/bundle/appimage/OmniVoice Studio_0.4.2_amd64.AppImage`.
3. Attempt a voice recording.
4. Observe “No microphone found…” and no WebKit recording stream in pavucontrol.
5. Quit and launch `frontend/src-tauri/target/debug/omnivoice-studio` directly.
6. Attempt the same recording; microphone capture works.

## Expected behavior

The AppImage should enumerate and capture from the same microphones as the raw binary.

## Environment

- OmniVoice Studio: 0.4.2, `main` at `5f39f9ff`
- OS: Debian 13, kernel 6.12.96, KDE Plasma on Wayland
- WebKitGTK: 2.52.5 (`libwebkit2gtk-4.1-0`)
- PipeWire: 1.4.2 with `pipewire-pulse` and WirePlumber
- GStreamer: 1.26.2; `pipewiresrc` and `pulsesrc` installed
- Default input: Elgato Wave:3

## Diagnostics

The host audio stack is healthy:

- `pactl list short sources` lists four active microphone sources.
- `wpctl status` lists the Elgato Wave:3 as the default source.
- Both `gst-launch-1.0 pulsesrc ... ! fakesink` and `gst-launch-1.0 pipewiresrc ... ! fakesink` capture successfully.
- `gst-device-monitor-1.0 Audio/Source` enumerates the microphones.
- PipeWire, pipewire-pulse, WirePlumber, xdg-desktop-portal, and the KDE portal backend are active.
- The raw debug binary links against the system WebKitGTK/GStreamer libraries and records successfully.

[...]

PR #1339, opened 2026-07-31
fix(scripts): desktop-prod:run wiped the data it was documented to preserve (#1333)

debpalash:
Secondary finding in #1333, and it costs real data.

`scripts/desktop-prod.sh` emulates a first install, so wiping is its default: it removes the app data dir, **`~/.omnivoice` — the SQLite database, every voice profile and all outputs** — the Tauri logs, and the WebKit profile. `--keep-data` is the only thing that suppresses that block.

`--skip-build` is an independent flag that only skips the cargo compile. `desktop-prod:run` passed `--skip-build` alone:

```json
"desktop-prod:run": "bun scripts/desktop-prod.mjs --skip-build",
```

while the script's own header calls it a re-launch:

```
#   bun desktop-prod:run      # re-launch last build (skip compile)
```

and its closing banner tells you to use it that way (`To re-run without rebuilding: bun desktop-prod:run`). So "just start it again without recompiling" silently deleted the developer's voice profiles and project database, every time.

### Fixed

- `desktop-prod:run` and `desktop-prod:run:pill` now pass `--keep-data`.
- The `--help` text and the script header state that `--skip-build` does **not** imply `--keep-data`.

The fix is deliberately in the package scripts, **not** in the flag parsing: making `--skip-build` imply `--keep-data` inside the script would remove a legitimate combination (a fresh-data run that skips the 1–3 min compile). The two stay independent.

`desktop-fresh:run` is left alone on purpose — that script is a stricter new-user emulation, so wiping is the point of its name.

### Tests

`tests/test_desktop_prod_scripts.py` — three cases: the re-launch aliases keep data; the fresh-install emulation still wipes (so the test can't pass by making everything harmless); and `--skip-build` stays independent of `--keep-data` in the script.

Confirmed fail-before by reverting just the `desktop-prod:run` line. `AppRun.test.sh` still 17/17, `bash -n` clean. Scripts-only change, so `bun.lock` is untouched and Docker's `--frozen-lockfile` is unaffected.

### Not addressed here

The primary bug in #1333 — the AppImage enumerating no microphones — is diagnosed but not fixed in this PR; I can't run an AppImage to verify, and that launcher just took a delicate Mesa fix. Analysis and a decisive test are on the issue.

Production desktop relaunch scripts now preserve user data, while `--skip-build` remains independent and `desktop-fresh:run` still wipes data. Running instances are terminated before launch, with process matching scoped to this checkout so installed applications are not killed. Human review should confirm that preserved-data relaunches and fresh-install cleanup remain correctly isolated.


debpalash:
Bot triage:

- **greptile P1 "Pill mode argument is discarded"** — not valid against this branch. It reads `--keep-data` as bypassing `kill_running_instances`, but the diff moves that call *out* of the `KEEP_DATA = false` branch precisely so it runs unconditionally; `test_running_instances_are_killed_regardless_of_data_policy` pins it there.
- **greptile P1 "Installed app is force-terminated"** — valid, and fixed in `4bb02f5`-and-following. Making the kill unconditional did widen the `pgrep` blast radius: `"OmniVoice Studio.app"` also matches an installed `/Applications` copy, so `desktop-prod:run` would kill the shipped app a developer was using. Previously masked, because the kill only ran on wipe runs where a clean slate had been asked for explicitly. The pattern is now scoped to `${TAURI_DIR}/target/debug/` (covers both the raw binary and the `.app` bundle, nothing else), and an installed instance gets a warning instead of a kill — it still has to be named, since single-instance keys on the bundle id and would otherwise swallow the launch with no explanation. Guarded by `test_kill_is_scoped_to_this_checkouts_build` (fails against the old pattern).

Issue #1258, opened 2026-07-24
[BUG] AppImage v0.4.0 blank window on Arch/CachyOS — AMD GPU, Mesa 26.1.5


rvasilev:
**Title:** AppImage v0.4.0 blank window on Arch/CachyOS — AMD GPU, Mesa 26.1.5

Downloaded the AppImage, made it executable, ran it — just a white rectangle. The app noticed something was wrong and tried reloading three times, but nothing changed. Error in the terminal:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
[app_lib::blank_guard][WARN] blank window detected (#root children = -2); reload 1/3
[app_lib::blank_guard][WARN] blank window detected (#root children = -2); reload 2/3
[app_lib::blank_guard][WARN] blank window detected (#root children = -2); reload 3/3
[app_lib::blank_guard][ERROR] window still blank after 3 reloads
```

I worked through every workaround in the install docs — `WEBKIT_DISABLE_DMABUF_RENDERER=1`, adding `LIBGL_ALWAYS_SOFTWARE=1`, `WEBKIT_DISABLE_COMPOSITING_MODE=1`, various combos. Same error every time.

**Building from source works perfectly.** Cloned the repo, `bun install`, `bun run desktop-prod` — launches and runs great. Same machine, same graphics driver, same everything. The app itself is solid — the issue is in the AppImage packaging.

**My system:**
- CachyOS (Arch), kernel 7.1.3
- AMD Radeon RX 6800 XT
- Mesa 26.1.5
- KDE Plasma 6, Wayland
- System WebKit: 2.52.5 (Arch package)

**What I think is happening:** The AppImage bundles an older WebKit that doesn't get along with Mesa 26.1.5. My system's own WebKit (2.52.5, compiled by Arch against the same Mesa I'm running) handles it fine — that's why the source build works. The workaround env vars don't help because the crash happens during the initial display connection setup, before any rendering method is chosen.

Related: issue #961 (same error, Ubuntu 26.04).

[...]

Issue #961, opened 2026-07-05
[Install] Ubuntu 26.04 white screen at first launch


bistory:
### Where did it fail?

App won't launch / blank or broken window

### The error

```text
Hi,
I tried to install OmniVoice on Ubuntu 26.04.
I downloaded the latest AppImage file, then followed the documentation (chmod +x ...)
Then ran :
`WEBKIT_DISABLE_COMPOSITING_MODE=1 ./OmniVoice.Studio_*.AppImage`

Then I got those errors/warnings in the Terminal and the screen stays blank :

/usr/lib/x86_64-linux-gnu/gvfs/libgvfscommon.so: undefined symbol: g_task_set_static_name
Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so
/usr/lib/x86_64-linux-gnu/gvfs/libgvfscommon.so: undefined symbol: g_task_set_static_name
Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so
[2026-07-05][06:41:03][app_lib][INFO] Global shortcut 'CmdOrCtrl+Shift+Space' registered
[2026-07-05][06:41:03][app_lib][INFO] WebKitGTK: media-stream enabled on 'main'
[2026-07-05][06:41:03][app_lib][INFO] WebKitGTK: media-stream enabled on 'widget'
[2026-07-05][06:41:03][app_lib][INFO] First run — awaiting setup screen confirmation before installing
/usr/lib/x86_64-linux-gnu/gvfs/libgvfscommon.so: undefined symbol: g_task_set_static_name
Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so
/usr/lib/x86_64-linux-gnu/gvfs/libgvfscommon.so: undefined symbol: g_task_set_static_name
Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...

** (omnivoice-studio:12443): WARNING **: 08:41:28.431: atk-bridge: get_device_events_reply: unknown signature
```

### Operating system

Linux (AppImage)

### How are you installing it?

Desktop app (installer / AppImage)

### Version

v0.3.10

### Network conditions (model/dependency downloads)

Normal / unrestricted

[...]

PR #1354, opened 2026-08-04
fix(appimage): let the host GStreamer win, and stop sharing its registry (#1333)

debpalash:
Fixes the microphone half of #1333. (The launcher data-wipe half was #1339, already merged.)

## What the reporter measured

Recording from the AppImage fails with "No microphone found", on a machine whose audio stack is demonstrably fine — `pactl list short sources` lists four sources, `wpctl status` names the default, and `gst-launch-1.0` captures through both `pulsesrc` and `pipewiresrc`. The same build's raw binary (`target/debug/omnivoice-studio`) records without trouble.

`GST_DEBUG=2` gives the whole answer in two lines:

```
WARN GST_REGISTRY gst_registry_binary_check_magic:
  Binary registry magic version is different : 1.23.90 != 1.3.0
GStreamer element appsink not found. Please install it.
```

## Root cause

linuxdeploy bundles `libgstreamer-1.0` because WebKitGTK links it — but not the plugins, which are `dlopen`'d at runtime and so invisible to anything static. `AppRun` puts the bundle first on `LD_LIBRARY_PATH`, so the **bundled core** runs and falls back to its compile-time default plugin directory: the **host's**. Those plugins were built against the host's core, the magic check rejects them, and the scan produces nothing.

`appsink` is one of the elements that goes missing, and it is precisely the one WebKit hands a capture stream to — so `getUserMedia()` rejects with `NotFoundError` and the UI reports no microphone. The raw binary works because it sets no `LD_LIBRARY_PATH`: core and plugins are both the host's, and they agree.

This is the same class as #1258 — a frozen bundled library paired against a host that moved on — in a different library, which is why `OMNIVOICE_PREFER_SYSTEM_WEBKIT=1` did nothing for the reporter. The WebKit fix cannot reach a problem one layer down.

## Fix

**Prefer the host's GStreamer.** Unlike WebKit there is no version comparison worth making: we ship no plugins, so a bundled core can only ever pair with host plugins it was not built against. Any host core beats that. `OMNIVOICE_PREFER_SYSTEM_GSTREAMER=0` falls back to the bundled core for a host whose own GStreamer is broken.

**Give the app a private registry cache.** This is a second, independent defect that the first one hid. GStreamer caches its plugin scan at `~/.cache/gstreamer-1.0/registry.<arch>.bin`, keyed by architecture alone — so two cores of different versions overwrite each other's file. That makes the failure depend on which application ran last, and it means the AppImage corrupts the cache for every other GStreamer app on the machine. Both directions disappear with `GST_REGISTRY_1_0` pointed at `~/.cache/OmniVoice/`.

## Tests

Three cases in `AppRun.test.sh` — host GStreamer present, absent, and opt-out — all failing against the previous `AppRun` and passing with this one. Suite: 20 pass / 0 fail.

## Verification still wanted

I do not have a Debian 13 + PipeWire host to confirm end-to-end capture. @Kakuzen93 — if you are willing to build this branch and try recording, that would close the loop; `GST_DEBUG=2` output either way is the useful thing.

## Docs

`docs/install/linux.md` gains a section for this symptom (the `GST_DEBUG` signature is distinctive, and "my mic works everywhere else" is exactly the search someone will make), and `CHANGELOG.md` an entry under `## [Unreleased]`.

The AppImage now preloads a loadable host `libgstreamer-1.0.so.0` without changing `LD_LIBRARY_PATH`, with an opt-out override and a private GStreamer registry cache. This prevents microphone capture failures caused by incompatible bundled GStreamer plugins and isolates cache data from host GStreamer. Review host-library detection and preload compatibility across Linux environments, including fallback to the bundled core.


debpalash:
Bot triage — greptile P1 "Host path overrides bundled libraries" was right, and the fix is now narrower.

Prepending the host GStreamer's directory was too blunt: on Debian that directory is `/usr/lib/x86_64-linux-gnu`, so putting it ahead of `${HERE}/usr/lib` replaced **every** bundled library with the host's copy, not just GStreamer — loader symbol errors, startup crashes or a blank window on a distro we never built against. Exactly one library needs to come from the host, so the mechanism has to be that narrow.

It now uses `LD_PRELOAD` with the full path to `libgstreamer-1.0.so.0`. That names one library and leaves the search path untouched, which also means the `#1258` WebKit ordering is unaffected — and it removes the composed-`LD_LIBRARY_PATH` block that only existed to stop the two prepends fighting each other. The preload is inherited by the Python backend, where nothing links GStreamer and it is inert; that is the accepted cost of the narrower mechanism, noted in the code.

`AppRun.test.sh` now asserts both halves — the library **is** preloaded, and the libdir is **not** hoisted — so a future "simplification" back to a path prepend fails.

Separately, the earlier CI failure on this branch was a genuine regression rather than a flake: the GStreamer prepend landed ahead of the host WebKit dir, and "host WebKit first" is the invariant #1258 turns on. It only showed on Linux because the WebKit ldpath cases do not stub away a real host GStreamer, so the runner had one to find and macOS did not. Reproduced locally with an `ldconfig` shim; the `LD_PRELOAD` approach makes the whole interaction moot.

debpalash:
greptile P1 "Host core keeps bundled dependencies" — valid, and the more dangerous of the two findings on this branch, because the failure mode is worse than the bug being fixed.

The host GStreamer links GLib, and the bundle ships GLib too, resolved bundle-first. A host core built against newer GLib than we bundle fails its relocations, so the app would not start **at all** — a broken microphone traded for a dead launcher. Taking host GLib along with it is not an escape: GLib is what WebKit is built against, so pulling that from the host reopens #961/#1258.

Rather than predict the pairing from version numbers, it now **tests** it. The dynamic loader processes `LD_PRELOAD` for any binary, so running `true` under the exact environment the app is about to get is a complete check of whether that library loads there — a missing dependency or an unresolved version tag (`version 'GLIB_2.84' not found`) fails it, and nothing else runs. Cheap, and decisive where a version comparison would be guesswork.

On failure the preload is skipped and the app starts on the bundled core, with a warning naming the mismatch — a silent fallback would leave the user with the original symptom and no thread to pull.

`OMNIVOICE_APPRUN_PRELOAD_PROBE` lets the suite choose the probe outcome (same precedent as the existing `OMNIVOICE_APPRUN_WK_MARKER` fixture hook), and the new "unloadable host core is skipped" case fails if the guard is removed. Suite: 22 pass / 0 fail.

debpalash:
Both of your Majors here were valid and I merged before reading them — my error. Corrected immediately in a follow-up rather than left as backlog, since finding 1 meant this PR shipped a no-op.

`command -v true` returns the shell builtin (the bare word `true`, not a path), so `[ -x "$probe" ]` was false on every host, the probe always failed, and the preload never happened. Verified: `bash -c 'p="$(command -v true)"; [ -x "$p" ]'` is false. A builtin never involves the dynamic loader either, so it could not have tested anything even if the guard had passed. And the probe took our library alone while the exported value appends any inherited `LD_PRELOAD`, so it could pass on a value the app never receives.

The suite missed both because every GStreamer case overrides the probe via `OMNIVOICE_APPRUN_PRELOAD_PROBE` — the default path real users take was never exercised. The follow-up covers that, not just the corrected lines.

Issue #1348, opened 2026-08-02
[Bug] TTS generate ran for more than 300s of actual compute time and was abandoned — t


vanderlpp:
## Error

```
TTS generate ran for more than 300s of actual compute time and was abandoned — the backend is running, but this job was too heavy for the available compute. The abandoned job cannot be killed: it keeps running and keeps holding the device until it finishes on its own, so an immediate retry competes with it. Wait for the current job to drain (or restart the backend) before retrying; this machine renders on CPU, where long generations are compute-bound. For a durable fix try shorter text or a lighter engine (OmniVoice GGUF and Supertonic-3 are CPU-tuned). If you expect very long single generations, raise OMNIVOICE_GENERATE_TIMEOUT_S.
StreamingPreviewError: TTS generate ran for more than 300s of actual compute time and was abandoned — the backend is running, but this job was too heavy for the available compute. The abandoned job cannot be killed: it keeps running and keeps holding the device until it finishes on its own, so an immediate retry competes with it. Wait for the current job to drain (or restart the backend) before retrying; this machine renders on CPU, where long generations are compute-bound. For a durable fix try shorter text or a lighter engine (OmniVoice GGUF and Supertonic-3 are CPU-tuned). If you expect very long single generations, raise OMNIVOICE_GENERATE_TIMEOUT_S.
    at handleEvent (http://192.168.1.159:3901/src/utils/streamingTts.js:311:11)
    at _streamGenerateSpeech (http://192.168.1.159:3901/src/utils/streamingTts.js:325:15)
    at async withTtsInflight (http://192.168.1.159:3901/src/api/generate.ts:17:10)
    at async http://192.168.1.159:3901/src/hooks/useTTS.js:210:6
```

## Environment

**Version:** `0.4.2`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36`
**OS:** `Linux-7.0.14-2-pve-x86_64-with-glibc2.36`
**Python:** `3.11.2`
**Compute device:** `cpu`
**CPU:** `AMD Ryzen 3 3200G with Radeon Vega Graphics`
**RAM:** `16 GB`
**Disk free:** `13.8 GB`
**Active TTS engine:** `omnivoice-gguf`

## Backend reachability

**Deployment mode:** `dev`
**Last backend response:** 2 s before this report

[...]

Issue #1338, opened 2026-07-31
[Bug] TTS generate ran for more than 300s of actual compute time and was abandoned — t


Kazhwaan:
## Error

```
TTS generate ran for more than 300s of actual compute time and was abandoned — the backend is running, but this job was too heavy for the available compute. The abandoned job cannot be killed: it keeps running and keeps holding the device until it finishes on its own, so an immediate retry competes with it. Wait for the current job to drain (or restart the backend) before retrying; most often the GPU is VRAM-starved (a resident model and this job contend for memory). For a durable fix, Flush caches / Unload the resident model (top toolbar or Settings → Models) before retrying, try shorter text, a lighter engine, or set the engine to CPU in Settings → Models. (Raise OMNIVOICE_GENERATE_TIMEOUT_S for very long single generations.)
StreamingPreviewError: TTS generate ran for more than 300s of actual compute time and was abandoned — the backend is running, but this job was too heavy for the available compute. The abandoned job cannot be killed: it keeps running and keeps holding the device until it finishes on its own, so an immediate retry competes with it. Wait for the current job to drain (or restart the backend) before retrying; most often the GPU is VRAM-starved (a resident model and this job contend for memory). For a durable fix, Flush caches / Unload the resident model (top toolbar or Settings → Models) before retrying, try shorter text, a lighter engine, or set the engine to CPU in Settings → Models. (Raise OMNIVOICE_GENERATE_TIMEOUT_S for very long single generations.)
    at d (http://tauri.localhost/assets/main-app-B6MZUbYV.js:16:28277)
    at Si (http://tauri.localhost/assets/main-app-B6MZUbYV.js:17:54)
    at async http://tauri.localhost/assets/main-app-B6MZUbYV.js:17:3640
```

## Environment

**Version:** `0.4.2`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36 Edg/150.0.0.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 3050 (8 GB VRAM)`
**CPU:** `Intel64 Family 6 Model 151 Stepping 5, GenuineIntel`
**RAM:** `31.8 GB`
**Disk free:** `2726.2 GB`
**Active TTS engine:** `omnivoice`

[...]

Issue #1329, opened 2026-07-30
[Bug] 0.4.2: dubbing stalls after ~one sentence and transcription is slower than before


debpalash:
Reported on Discord: dubbing on **0.4.2** stalls, and the whole pipeline feels slower than it used to.

### What was reported

> Today, after some time, I wanted to do some dubbing. I updated the application to version 0.4.2, and for some reason, it feels slower than before, **especially the transcription**. But besides that, now I can't even dub a single sentence; **it just gets stuck in a loop and doesn't advance**.

Following up a minute later:

> Well, writing this it seems like it's only advanced one sentence
>
> It used to be much faster
>
> I couldn't say for sure, but it seems the problem is with **voice cloning**; I'm using **OmniVoice** in settings → TTS

Two screenshots of the stalled dub were attached (not reposted here).

### Why this one matters

This is a returning user comparing 0.4.2 against a version that worked for them, so it reads as a **regression**, not a first-run setup problem. Two distinct symptoms:

1. **Transcription noticeably slower** than the previous version they used.
2. **Dubbing advances roughly one sentence and then stalls** — no error surfaced, it just stops making progress.

Their own guess is that voice cloning is the stuck part, with the OmniVoice engine selected for TTS.

### What we need

- Exact previous version, to bracket the regression.
- OS, GPU, and whether compute device shows CUDA/MPS/CPU.
- Backend log covering a stalled dub (Settings → Logs → Backend) — whether the pipeline is waiting on the clone step, the ASR step, or neither.
- Whether a dub with a **preset** voice (no cloning) advances normally, which would isolate it to the clone path.

### Notes

No existing issue covers this. Possibly related in mechanism to the audiobook hang fixed in #1321 (a bare `StopIteration` from a generator-driven engine could never reach the awaiting future, so the caller waited forever with no error) — the "advances one sentence then stops with no error" shape is similar, and the dub path has its own `run_in_executor` call sites. Worth checking whether the dub loop can wedge the same way now that the pool-level guard is in.

---
Reported by `nanai97` on Discord — https://discord.com/channels/1497926884707799144/1497926886020354102/1532378473581252751

debpalash:
Related: #1355 (merged) makes the abandoned-job case diagnosable — when a generate overruns its budget the backend now records the stack of the worker that is stuck, rather than only the fact that it timed out.

[...]

PR #1355, opened 2026-08-04
fix(gpu): record WHERE a wedged job was stuck when its budget expires (#1338)

debpalash:
Makes #1338 / #1329 / #1348 diagnosable. Does **not** fix the underlying hang — see "What this is not" below.

## The problem with the reports

Three reporters on v0.4.2 hit *"TTS generate ran for more than 300s of actual compute time and was abandoned"*. Two of them were on an **RTX 3050** and an **RTX 3060**, rendering a **single sentence**. `GpuJobTimeoutError` is only raised once a worker has picked the job up — queue wait is bounded separately and surfaces as a retryable 503 — so that is 300+ seconds of *execution*, not of waiting in line.

On a 12 GB card, one sentence is not a workload. But the message says "too heavy for the available compute", because that is the only story the timeout path is able to tell.

And nothing in the log could contradict it. The timeout branch logged *that* the budget was exceeded, reset the pool, and returned. The worker thread cannot be cancelled, so it was still running, still on a real stack — and we discarded it. Every report of this class therefore arrives undiagnosable, and the only advice available is "reproduce it under a debugger", which is not a reasonable thing to ask of someone who just wanted to dub a video.

## The change

`log_gpu_pool_worker_stacks()` — at timeout, before abandoning the job, dump the current frame of every GPU-pool worker.

`sys._current_frames()` reads the frame of every live thread *including one wedged inside a C call*, which is precisely the case that matters: the worker is stuck somewhere in the model, and it is the one thing we cannot ask it about.

- **Filtered to gpu-pool workers**, by the existing `_GPU_POOL_THREAD_PREFIX`. The process also holds a web server, a watchdog, and a watermark pool; a dump of everything is a dump nobody reads.
- **Capped at 25 frames** — deep enough to cross the engine adapter into the model's own stack, shallow enough that a 1-worker and an 8-worker host both produce something a human will read.
- **Can never raise.** It runs inside the timeout handler, so a diagnostic that throws would replace a real `GpuJobTimeoutError` with an unrelated crash.
- **Runs before `reset()`,** and that ordering is asserted rather than assumed. `reset()` swaps in a fresh executor, and once the wedged thread is no longer named as a pool worker the capture finds nothing — it would still run, still log, and be empty, which is worse than not having it because it looks like it worked.

It logs at WARNING, so it lands in `backend.log` and therefore in diagnostic bundles and auto-filed bug reports without the user doing anything.

## What this is not

This does not fix the hang. It makes the *next* report of it carry the one fact needed to fix it: which frame the job died in. If the deepest frame is inside the model's forward pass, it is a hang and the "too heavy for the available compute" wording is actively misleading; if it is in a host-to-device copy, the VRAM story was right after all. Right now we cannot tell those apart, which is why #1338 has been stuck on "please attach a log" for a week.

## Tests

`tests/test_gpu_wedge_stack_capture.py` — 5 tests, all failing before this change:

- the capture reaches the frame the worker is actually parked in (asserted by function name, since that is the only part a maintainer reads);
- non-pool threads are excluded;
- no pool worker at all returns empty rather than erroring — the timeout path must not depend on it finding anything;
- it never raises, even when thread enumeration itself throws;
- end-to-end through `run_on_gpu_pool_guarded`: the capture happens *before* `reset()`, and the captured text names the wedged function.

Related suites green: `test_generate_timeout_730`, `test_gpu_pool_queue_accounting_1190`, `test_recurrence_hardening`.

One thing I noticed while running these and did **not** touch here: `tests/test_load_budget_split_1033.py::test_speech_survives_a_load_slower_than_the_generate_budget` fails when the file is run on its own and passes in the full suite, on clean `main` as well as on this branch. That is order-dependence in the opposite direction from #1269 and belongs in its own change.

When a GPU job exceeds its execution-time budget, the timeout path logs up to 25 sanitized stack frames for each GPU-pool worker before resetting the pool. The capture filters unrelated and stale workers, supports C calls, and suppresses diagnostic failures so timeout handling continues. Reviewers should assess WARNING-level log volume and confirm sanitization remains fail-closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 - Linux AppImage: `OMNIVOICE_PREFER_SYSTEM_WEBKIT=1` forces your own WebKitGTK for hosts where its version can't be read automatically (no `pkg-config`), and `=0` forces the bundled one (#1258)
 - Dubbing: the transcription overlay said "Transcribing with Whisper…" whatever ASR engine was actually running — it now names the stage, in all 21 languages — thanks @paoloantinori! (#1352)
 - Error messages no longer arrive with terminal colour codes spliced into the sentence (`download: ^[[0;31mERROR:^[[0m …`) — every surfaced failure is cleaned now, whichever tool produced it. (#1344)
+- A TTS job abandoned for exceeding its compute budget now records where it was actually stuck, so a hang stops being reported as a machine that is merely too slow. (#1338, #1329, #1348)
 
 ### Docs
 
--- a/backend/services/model_manager.py
+++ b/backend/services/model_manager.py
@@ -583,7 +583,7 @@ async def run_on_gpu_pool_guarded(fn, *, what: str = "GPU job",
         # Capture the stacks BEFORE reset(): reset() replaces the executor, and
         # once the wedged thread is no longer a pool worker we can no longer
         # tell it apart from any other thread in the process.
-        log_gpu_pool_worker_stacks(what, timeout)
+        log_gpu_pool_worker_stacks(what, timeout, executor=ex)
         _reset = getattr(ex, "reset", None)
         if callable(_reset):
             try:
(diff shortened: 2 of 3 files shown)


debpalash:
CI caught a real defect in my own test rather than in the code: `test_capture_is_empty_when_no_pool_worker_exists` asserted on the *ambient* thread list, and in a full session earlier tests build a real GPU pool whose idle workers outlive them. So it passed alone and failed in the suite — the same order-dependence class as #1269, in the opposite direction.

Fixed by manufacturing the absence instead of assuming it: the filter prefix is monkeypatched to something nothing uses, which tests the branch rather than the environment. Verified both ways — `pytest tests/test_gpu_pool_queue_accounting_1190.py tests/test_gpu_wedge_stack_capture.py` is 20 passed with the fix and 1 failed without it.

debpalash:
Both findings valid and fixed.

**CodeRabbit (CWE-532) — the important one.** `traceback.format_stack()` emits absolute source paths, which on a user's machine begin with their home directory, i.e. their account name. This log goes to `backend.log`, and `backend.log` goes into diagnostic bundles and prefilled bug reports — so an unredacted capture would have leaked the username of everyone who ever hit a timeout, in a feature whose whole purpose is to get more of these logs sent to us. Now routed through `core.failure.sanitize` (the repo's existing answer, not a second one), which also covers HF tokens and `*TOKEN*/*KEY*/*SECRET*` env values. If sanitizing itself throws, the stacks are dropped rather than logged raw.

**greptile P1 — also real.** A wedged worker survives `reset()`: it cannot be cancelled, so it keeps running under the same `gpu-pool` prefix the replacement pool uses. The second timeout in a session would log both with nothing to distinguish them, and the stale one is the more misleading of the two because it names an operation that is *not* the job that just failed. The live pool is now identified through its own thread set; everything else is marked `STALE — a worker abandoned by an earlier timeout, still running; not the job that just failed`.

That set comes from `ThreadPoolExecutor._threads`, which is private, so I added a third test for the degrade path: unknown internals must cost the *label*, never the *stacks*. A diagnostic that disappears because an attribute moved in a future Python is worse than an unlabelled one.

Three new tests; the redaction and stale-label ones both fail against the previous commit. Related suites green (43 passed).

Issue #963, opened 2026-07-05
[Bug] Some errors found


camilocarrai1997:
### Before filing

- [x] I searched [existing issues](https://github.com/debpalash/OmniVoice-Studio/issues?q=is%3Aissue) and this isn't a duplicate.
- [x] I'm on the latest release (or `main`) — older builds may already be fixed.

### What happened?

Several problems persist after the updates, some of them are new.
1- The boxes that appear below the video with the dialogue are invisible; they appear and disappear as the video plays. This prevents us from moving and adjusting them properly if we want to make any precise changes.

2- There are still several problems detecting the dialogue spoken by the speakers; they are constantly getting mixed up.

3- There's quite a bit of desynchronization with the dubbed version. In the dub, there are moments where the speakers start talking much earlier or later than in the original video. I don't know if this could be the reason, but perhaps when a sound is detected in the original video, it's being mistaken for the start of the conversation, making, for example, the sound of footsteps appear as the beginning of the conversation.

4- I'm in the configuration setting up the use of Ollama locally, it seems to work correctly, but when I restart Omnivoice it seems that the changes are not saved and I have to do them again.

### Steps to reproduce

1. Make any attempt at dubbing

### What did you expect instead?

That the speakers are correctly detected in their dialogues and that these do not get mixed up.

Make sure the dialog boxes that appear at the bottom don't disappear and are visible at all times.

Being able to save the settings when selecting a model, for example Ollama, upon restarting Omnivoice

Correct synchronization where each person's dialogue begins at the same moment as it does in the original video.

### Operating system

Windows (x64)

### How did you install it?

Desktop app (installer / AppImage)

### Version

v0.3.10

### Compute device

NVIDIA GPU (CUDA)

### Logs / diagnostic bundle

```text

```

### Anything else?

It would be good to improve the detection of when multiple speakers are talking simultaneously; there are moments in videos where one speaker says something while another is still talking. Overlaying one dialogue over the other in a more natural way would be ideal.

debpalash:
Thank you for re-testing on v0.3.10 so quickly — this is exactly the feedback loop that gets things fixed. Point by point, honestly:

**1. Invisible boxes — this one stings.** v0.3.10 shipped a structural fix for exactly this (we stopped GPU-compositing the box lane entirely and made the fills fully opaque), and you're telling us it didn't cure your case. Two things would pin down what's really happening on your machine:
[...]

PR #1359, opened 2026-08-04
fix(llm): LM Studio translation failed on a placeholder model name (#1332)

### LINKED ISSUE #1332: [Bug] Translation fails with LM Studio as the local LLM provider (Ollama works)
Reported on Discord (in Russian; translation below).

### What was reported

> Не работает перевод через LM Studio. Через ollama все ок.

Translated:

> Translation via **LM Studio** doesn't work. Via **Ollama** everything is fine.

Two screenshots were attached (not reposted here).

### Why this is actionable as filed

It's a clean A/B: the same user, same machine, same translation task — one local-LLM provider works and the other doesn't. That rules out the surrounding pipeline and points at the LM Studio integration specifically.

Both are OpenAI-compatible local servers, so the usual suspects are narrow:

- Base URL / path assumptions (LM Studio serves `/v1` on `:1234` by default; Ollama's OpenAI-compat endpoint differs).
- A required `model` name that LM Studio rejects when it doesn't match the loaded model exactly.
- Request fields Ollama tolerates and LM Studio rejects (or the reverse) — e.g. `format`, `options`, response-format/JSON-mode handling.
- Streaming vs non-streaming response shape.

### What we need

- The error the app shows (and the backend log line for the failed translation).
- LM Studio version, the base URL configured in OmniVoice, and the exact model name loaded.
- Whether LM Studio's own server log shows the request arriving at all — that splits "never reached it" from "rejected it".

---
Reported by `biga73` on Discord — https://discord.com/channels/1497926884707799144/1497926886020354102/1531605067629854792

debpalash:
Fixes #1332.

## The A/B was the whole diagnosis

> Не работает перевод через LM Studio. Через ollama все ок.
> (Translation via **LM Studio** doesn't work. Via **Ollama** everything is fine.)

Same user, same machine, same task — which rules out the pipeline and points at the LM Studio integration. The difference turns out to be one line in the provider table:

```python
Provider("ollama",   ..., "llama3.1",     local=True, ...)
Provider("lmstudio", ..., "local-model",  local=True, ...)
```

`local-model` is a **placeholder, not a model id**. LM Studio serves whatever the user has loaded and rejects a name it does not know, so every request 404s. Ollama's default is `llama3.1` — a real name people actually pull — so the identical code path worked there. Anyone who never opened Settings and typed a model name got a translation feature that could not work.

## Why not just ship a better default

There isn't one. The correct value depends on which model that user loaded, so any string we hardcode is a guess that is wrong for most people. So the fix asks the server.

`resolve_model()` now discovers from `/v1/models` for providers whose default is flagged as a placeholder. Position in the chain is the load-bearing part:

```
env override  →  stored setting  →  discovered  →  built-in default
```

- **Below** any explicit setting, so a deliberate choice is never silently replaced by whatever the server happens to be serving. That would be a worse bug than the one being fixed.
- **Above** the default, so a server that is down or has nothing loaded leaves the caller exactly where it was rather than raising inside a translation.
- **Off entirely** for providers with real defaults — probing OpenAI to resolve `gpt-4o-mini` would add a network round-trip to every call for no benefit.

Cached per provider, because translation resolves the model *per segment* and a round-trip each time would trade a broken setup for a slow one. Dropped whenever a `base_url` or `model` edit could invalidate it — a stale id would make the user's own change look like it did nothing, which is the failure mode that produced #963.

Where several models are loaded it picks deterministically (sorted, not "whatever the server listed first") and logs that it chose, so two runs on one machine agree and a bug report stays reproducible.

## The error message was also sending people the wrong way

A 404 from a **local** provider is almost never a wrong URL — the request reached the server. The generic *"Not found (404) — check the model name and Base URL path"* sends the user to audit a URL that is fine.

A local 404 now says which models the server actually has:

> LM Studio (local) is running, but has no model named `local-model`. Loaded right now: qwen2.5-7b-instruct, llama-3.2-3b. Pick one in the Model field above.

…or, when nothing is loaded, that it has none and one needs loading first.

## Tests

`tests/test_lmstudio_model_discovery.py`, 10 cases, all failing before this change. The ones that matter most are the guards rather than the happy path:

- an explicit model (stored **or** env) wins, **and discovery does not even run** — asserted by call count, not just by the result;
- providers with real defaults never probe;
- a server that is down, and a server with an empty model list, both fall back instead of raising;
- the choice is deterministic across differently-ordered responses;
- discovery is cached (5 resolves, 1 round-trip);
- saving a new base URL drops the cache;
- a guard on the placeholder flag itself, since adding one puts a network probe in the resolve path and should be deliberate.

The suite also neutralises the developer's own env and settings store — otherwise this passes on one machine and fails on another.

Related suites: 170 passed.

## What I could not verify

I don't have LM Studio here, so the round-trip against a real server is untested. The failure mode is well-determined from the provider table and the 404, but @biga73 confirming on their setup would close it properly.

LM Studio now discovers the loaded model through `/v1/models` instead of sending the `local-model` placeholder, while preserving explicit settings and adding cached, invalidated fallback behavior. This fixes translation failures caused by model-name mismatches and improves local-provider 404 messages. Review cache expiry, invalidation, and fallback paths because unavailable servers or configuration changes can affect model selection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 - Linux AppImage: recording failed with "No microphone found" on hosts whose GStreamer is newer than the build runner's, even with a verified-healthy audio stack — your own GStreamer now takes precedence, and the plugin cache is app-private so it can neither be confused by nor corrupt the one other apps use — thanks @Kakuzen93! (#1333)
 - Linux AppImage: that GStreamer preference actually takes effect — the check guarding it could never pass, so it had been silently doing nothing. (#1333)
 - A TTS job abandoned for exceeding its compute budget now records where it was actually stuck, so a hang stops being reported as a machine that is merely too slow. (#1338, #1329, #1348)
+- Translation through LM Studio works. The built-in model name was the placeholder `local-model`, which LM Studio rejects because it serves whatever you have loaded — OmniVoice now asks it, and a 404 from a local server names the models that ARE loaded instead of telling you to check a URL that was fine — thanks @biga73! (#1332)
 
 ### Docs
 
--- a/backend/api/routers/settings.py
+++ b/backend/api/routers/settings.py
@@ -462,12 +462,21 @@ def test_llm_provider(provider_id: str):
                     + (" …" if len(available) > 10 else "")
                     + ". Pick one in the Model field above."
                 )
-            else:
+                # The cached discovery, if any, produced a name this server
+                # rejects — most likely the user swapped the loaded model
+                # inside the local app. Drop it so the next attempt re-asks
+                # rather than repeating the same 404 until the TTL expires.
+                llm_providers.forget_discovered_models(p.id)
+            elif available == []:
                 detail = (
                     f"{p.display_name} is running, but reports no loaded models, "
                     f"so {asked!r} cannot be served. Load a model in "
                     f"{p.display_name} first, then test again."
                 )
+                llm_providers.forget_discovered_models(p.id)
+            # available is None: the model listing itself failed, so nothing
+            # here is established. Keep the generic 404 text rather than
+            # inventing a diagnosis the lookup did not support.
         return {
             "ok": False,
             "kind": kind,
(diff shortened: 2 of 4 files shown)


debpalash:
All three findings valid — thanks, they were all about the cache being unbounded in one direction or missing in the other.

**greptile P1 / CodeRabbit Major — failed discovery retried indefinitely.** The sharp version: `resolve_model()` runs once per *translated segment*, so a stopped LM Studio cost a 5s timeout per segment. A 200-segment dub would spend ~1000s discovering nothing — strictly worse than the bug this PR fixes. Failures are now remembered for 30s: long enough to stop the per-segment cost, short enough that starting the server recovers in seconds instead of needing an app restart.

**greptile P1 — discovered model goes stale.** Correct, and the trigger needs no OmniVoice action at all: the user swaps the loaded model inside LM Studio and every translation 404s until restart. Now a 300s TTL, *plus* an immediate invalidation when a local 404 proves the cached name is one the server rejects — so the common case self-heals on the next attempt rather than at the next expiry.

Both of those mean the cache cannot be a `dict[str, str]` any more: "no entry" and "we looked and there was nothing" have to be distinguishable, or the negative case is not cacheable at all. It stores `(value_or_None, expiry)`.

**CodeRabbit — the 404 message claimed more than the lookup proved.** `_local_models` collapsed a *failed* listing into `[]`, so the error could state "reports no loaded models" about a lookup that never happened — swapping a vague right answer for a confident wrong one. `None` and `[]` are now distinct, and when the listing itself fails the generic 404 text stands.

Five new tests, three failing before the change. One note on how they are written: they age the cache entry directly rather than patching `time.monotonic`, because that name is the stdlib's — shared with sqlite and logging — and freezing it broke the settings store underneath the test on the first attempt.

Issue #1330, opened 2026-07-30
[Bug] Generation silently drops the last few sentences of the input


debpalash:
Reported on Discord: generation silently drops the end of the text.

### What was reported

> also when i make voices this app dosent generate me the last few sentences. for rest this app is a banger voices are perfect no glitches love this app

So: the audio that comes back is **clean**, it is just **missing the tail** of the input. No error, no warning — the user only notices because the text they pasted is longer than the audio they got.

### Why this is worth chasing

Silent truncation is the worst shape for this class of bug: the output looks fine, so a user only catches it by reading along. It also lands squarely on the project's core promise — the failure never announces itself, so there is nothing for the user to act on.

Likely places to look:

- Chunking/segmentation dropping the final chunk when the text doesn't end on a sentence boundary (or ends without terminal punctuation).
- A trailing chunk below some minimum length being filtered out before synthesis.
- `trim_trailing_silence` over-trimming a quiet final segment.
- A generation cap (token/duration) hit without surfacing anything.

### What we need

- The exact text that reproduces it (or any text where the tail is dropped), plus how much is missing — last sentence, last paragraph, a fixed duration?
- Engine in use (Settings → TTS) and app version.
- Whether the input ends with punctuation, and whether adding a trailing newline or full stop changes anything.
- Whether it reproduces with short text (2–3 sentences) or only with long input — that separates a chunking bug from a cap.

---
Reported by `tipoo_st` on Discord — https://discord.com/channels/1497926884707799144/1497926886020354102/1532438016621154385

debpalash:
Partly addressed on `main` (#1360) — the silence is fixed, the underlying cause is not, and the fix is what makes the cause findable.

Three of the four hypotheses in the issue are ruled out, by probe rather than by reading, and each elimination is now pinned by a test so nobody has to redo it:

- **chunking** loses nothing — every non-whitespace character survives into some chunk, including text with no terminal punctuation and text ending mid-word;
- **concatenation** joins everything it is given;
- **`trim_trailing_silence`** cuts only from the last *voiced* sample, so it cannot remove speech.

That left one line at the top of the join:

```python
chunks = [c for c in chunks if c is not None and c.shape[-1] > 0]
```

When an engine returns nothing for one slice of text, skipping it is still the right *joining* behaviour — the alternative is a crash or a gap mid-sentence. Doing it **silently** is not. If the empty chunk is the last one, the output simply ends early: "doesn't generate me the last few sentences", exactly.

[...]

PR #1360, opened 2026-08-04
fix(tts): a chunk that renders to nothing is no longer dropped in silence (#1330)

debpalash:
Addresses #1330. Makes the failure findable; does **not** fix whatever makes the engine return nothing.

## Narrowing it

The report is the hardest shape to chase:

> also when i make voices this app dosent generate me the last few sentences. for rest this app is a banger

Clean audio, correct voice, just short. The issue listed four hypotheses; three are eliminated here by probe rather than by reading, and each elimination is now a test so nobody has to redo it:

- **`split_text_into_chunks` loses nothing.** Every non-whitespace character survives into some chunk, across text with no terminal punctuation, text ending mid-word, 500-char runs and multi-paragraph input.
- **`concatenate_audio_chunks` joins everything it is given.**
- **`trim_trailing_silence` cannot remove speech** — it cuts only from the last *voiced* sample, keeping 0.3 s past it.

That leaves one line, at the top of the join:

```python
chunks = [c for c in chunks if c is not None and c.shape[-1] > 0]
```

## The defect

When an engine returns nothing for one slice of text, skipping it is still the **right joining behaviour** — the alternative is a crash or a gap in the middle of a sentence. Doing it **silently** is not.

The waveform comes back clean and simply short. Nothing errors, nothing looks wrong, and the only way to notice is to read along while listening — which is precisely how it was reported, and why it went unnoticed long enough to reach Discord as a compliment with a complaint attached.

If the empty chunk is the **last** one, the output ends early: "doesn't generate me the last few sentences", exactly.

## Change

The join now counts the drops, logs at **WARNING** — this is output the user spent compute on, and the log is the only place it can surface — and names the lost text when the caller can supply it. Wired through all three generation call sites and the audiobook renderer.

Audiobook additionally pre-filtered before calling:

```python
rendered = [r for r in rendered if r is not None and ...]
```

which hid the same drop a second time **and** misaligned `rendered` from `chunks`, so the concat could not have named what was lost even with the text in hand. It passes the full list now and uses a separate count for its own branching.

The reporting cannot break the join — including when the logger itself is what broke, since turning missing audio into a failed render would be strictly worse than the bug.

## Tests

`tests/test_dropped_chunk_is_not_silent.py`, 9 cases, 6 failing before this change:

- a dropped chunk is logged, and the surviving chunks are still joined;
- the log names the lost text, and **not** the chunks that rendered fine;
- the tail case specifically (last chunk empty → short output);
- **nothing is logged on a healthy render** — a warning on every render would train everyone to ignore the one that matters;
- `None` counts the same as an empty tensor, so an engine returning either does not get a quieter path;
- all-empty still returns a tensor, leaving the total-failure case to the downstream dead-render guards;
- a caller passing fewer texts than chunks still gets the count, rather than losing the diagnostic to an `IndexError`;
- the reporting cannot break the join;
- the chunking-loses-nothing property, so a future splitter change cannot quietly reintroduce the hypothesis this ruled out.

Related suites: 234 passed.

## What is still open

Why an engine returns no audio for a chunk. That needs a reproducing input, and this is what makes one obtainable — the next report of #1330 will carry the sentence that failed instead of "the last few sentences".

TTS concatenation now reports dropped `None` or empty chunks with source text while preserving rendered audio. Generation and audiobook paths use `join_rendered_chunks` to maintain chunk alignment and report one-survivor and zero-survivor cases consistently. The underlying empty-audio engine issue remains unresolved, so review diagnostic behavior when chunk metadata is incomplete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 - Linux AppImage: recording failed with "No microphone found" on hosts whose GStreamer is newer than the build runner's, even with a verified-healthy audio stack — your own GStreamer now takes precedence, and the plugin cache is app-private so it can neither be confused by nor corrupt the one other apps use — thanks @Kakuzen93! (#1333)
 - Linux AppImage: that GStreamer preference actually takes effect — the check guarding it could never pass, so it had been silently doing nothing. (#1333)
 - A TTS job abandoned for exceeding its compute budget now records where it was actually stuck, so a hang stops being reported as a machine that is merely too slow. (#1338, #1329, #1348)
+- Generation that silently dropped the end of the input now says so. When an engine returns no audio for part of the text the result sounds clean and is simply short, so the only way to notice was to read along; the backend log now names the sentences that produced nothing. (#1330)
 
 ### Docs
 
--- a/backend/api/routers/generation.py
+++ b/backend/api/routers/generation.py
@@ -551,7 +551,8 @@ def _run_inference(
                     if used_seed is not None:
                         torch.manual_seed(used_seed + i)
                     parts.append(_gen(chunk_text, None)[0])
-                audio_out = concatenate_audio_chunks(parts, sr, _xfade_ms)
+                audio_out = concatenate_audio_chunks(parts, sr, _xfade_ms,
+                                                     texts=text_chunks)
             else:
                 audio_out = _gen(text, duration)[0]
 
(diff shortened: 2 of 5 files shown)


debpalash:
Both of you caught the same hole and it was real: the one-survivor branch in `audiobook.py` returned that chunk directly, skipping the join — and therefore skipping the reporting the join does. A chapter whose text mostly failed came back short and said nothing about it, which is this very bug one branch over. Zero survivors had the same problem.

Rather than add two more conditions to an inline branch, the decision moved into `chunked_tts.join_rendered_chunks()`: kept/dropped, report, and `None` when nothing rendered — `None` rather than a silence buffer, so the caller's dead-render handling still owns that case instead of being handed something that looks like audio. One place that can be wrong, and a testable one. A second inline copy is exactly how the hole appeared, so there is a test pinning that audiobook routes through the shared function.

Four new cases on the helper; `test_one_survivor_of_several_is_reported` fails if the short-circuit is restored.

On the test note (CodeRabbit): fair on the first half, and fixed — `test_reporting_never_breaks_the_join` now asserts the report was *attempted*, not merely that nothing blew up, so it can no longer pass on a build where the reporting does not exist.

I have kept `test_chunking_itself_loses_no_text` rather than removing it, and made the docstring say plainly that it pins an *eliminated hypothesis* rather than a fixed defect — it does not fail before this change and is not meant to. "The chunker drops the tail" was the first and most plausible explanation for #1330; ruling it out took a probe, and without this the next person reading the issue redoes that work. A future splitter change that did start losing the tail would reproduce the reported symptom exactly, and this is what would catch it.

Issue #1331, opened 2026-07-30
[Bug] Re-dubbing a single sentence loses the cloned voice — only a full re-dub keeps it


debpalash:
Reported on Discord (in Russian; translation below).

### What was reported

> если переозвучивать отдельные предложения то голос не берется из видео, приходится все переозвучивать чтобы клон голоса сработал

Translated:

> If you re-dub **individual sentences**, the voice isn't taken from the video — you have to re-dub **everything** for the voice clone to work.

A screenshot was attached (not reposted here).

### The bug

Re-rendering a **single line/segment** of an existing dub does not pick up the cloned reference voice from the source video. The result is a segment that doesn't match the rest of the dub. The only workaround the user found is to re-dub the entire video, which throws away every already-correct segment — expensive, and it defeats the purpose of per-line editing.

That points at the single-segment re-dub path resolving its voice differently from the full-run path: the reference audio extracted from the video appears to be attached to the batch run rather than persisted per segment, so a later one-off render falls back to a default/uncloned voice.

### What we need

- Whether the affected segment uses a *default* voice or *no* voice conditioning at all (audibly different from both the clone and the rest of the dub?).
- App version and OS.
- Whether it also happens when the clone is a saved voice profile rather than one extracted from the video — that separates "reference not persisted" from "reference not passed on the re-render path".

### Note

Not a duplicate of #486 (multi-speaker per-speaker voice auto-assignment, closed) — this is single-speaker, and about the *re-render* path specifically.

---
Reported by `xipb14` on Discord — https://discord.com/channels/1497926884707799144/1497926886020354102/1531041313062129674

debpalash:
Partly addressed on `main` (#1361) — the mechanism is identified and now visible; the root cause is not fixed yet.

The clue was the reporter's own workaround. Re-dubbing everything is **not** a different code path — the UI regenerates a subset through the same `/dub/generate` with `regen_only`, so the voice resolution is shared. What a full re-dub does differently is **re-extract the clips**.

[...]

PR #1361, opened 2026-08-04
fix(dub): say when a clone reference is gone instead of rendering a default voice (#1331)

debpalash:
Addresses #1331. Explains the reported behaviour and makes it visible; does **not** stop the clips being cleaned up.

## What the report actually describes

> если переозвучивать отдельные предложения то голос не берется из видео, приходится все переозвучивать чтобы клон голоса сработал
>
> "If you re-dub **individual sentences**, the voice isn't taken from the video — you have to re-dub **everything** for the voice clone to work."

That last clause is the clue, and it points somewhere specific: re-dubbing everything is not a different *code path* — the UI regenerates a subset through the same `/dub/generate` with `regen_only`, so the resolution logic is shared. What a full re-dub does differently is **re-extract the clips**.

## Mechanism

Clone references are **file paths** into the job's extracted-clip directory (`speaker_clones`, `segment_clones`). The entire job dict — those paths included — is persisted to `dub_history.job_data` so saved projects reopen after a restart.

So the job outlives its clips. Reopen a saved dub after the clip directory has been cleaned, regenerate one line, and every resolution branch in `dub_generate.py` happily produces a path that is no longer there. **Nothing checked it**, and an engine handed a missing reference renders *uncloned* rather than failing — so the line comes back in a default voice, matching nothing else in the dub, with no error anywhere.

Re-running the full dub re-extracts the clips. Which is precisely why that "works", and why the reporter found that workaround on their own without knowing the cause.

## The change: diagnostic only

`warn_if_ref_missing()` is called immediately before the engine call in both the full render and the single-segment preview — the last point where every resolution branch has converged on a path. It logs which segment lost its reference and to what path, and it **returns the reference unchanged**.

That restraint is deliberate, and I got it wrong first:

- Nulling the reference would not alter what the user hears. The engine already falls back; that is the entire problem.
- It would decide, on the engine's behalf, that a path OmniVoice cannot `stat` is unusable — untrue for anything resolved inside a sidecar's own namespace.
- My first cut did null it, and broke **7 existing tests** in `test_dub_voice_match.py` that legitimately assert a synthetic path reaches the engine. That was the right signal, not a test problem.

The defect is the silence, not the fallback.

Warns **once per segment per job**: a 300-segment dub whose clips were cleaned should name which lines lost their reference, not repeat itself on every retry of one line. "Some of them" is not actionable.

## Tests

`tests/test_missing_clone_ref_is_not_silent.py`, 10 cases:

- an existing reference passes through untouched;
- a missing one is **still passed to the engine** — the restraint above, pinned so a later "cleanup" cannot quietly turn this into a behaviour change;
- the log names the segment, the path, and what the user will actually hear (`DEFAULT voice`) — a message that only said "reference missing" reads as housekeeping rather than as the explanation for their bug;
- **no** warning when there is no reference at all — designed voices, presets and plain TTS are the ordinary case and would bury the real one;
- once per segment, but per-segment, and not shared between jobs that reuse segment ids;
- an unreadable path (dead mount, permissions) is treated the same as a deleted one rather than escaping through an `OSError`;
- both call sites route through the helper, so a resolution branch added later cannot bypass it.

Full dub suite: 326 passed.

## What is still open

Why the clips go missing — whether that is temp-dir cleanup, a job-cleanup path, or the user clearing storage. That needs a reproduction, and this is what makes one useful: the next report will carry the segment ids and the paths instead of "the voice isn't taken from the video".

Added warnings for missing or unreadable voice-clone references during full renders and segment previews, with per-job, segment, and path deduplication. The engine still receives the original path and falls back to the default voice when the reference is unavailable. The underlying clip-cleanup issue remains unresolved and requires human review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 - A TTS job abandoned for exceeding its compute budget now records where it was actually stuck, so a hang stops being reported as a machine that is merely too slow. (#1338, #1329, #1348)
 - Translation through LM Studio works. The built-in model name was the placeholder `local-model`, which LM Studio rejects because it serves whatever you have loaded — OmniVoice now asks it, and a 404 from a local server names the models that ARE loaded instead of telling you to check a URL that was fine — thanks @biga73! (#1332)
 - Generation that silently dropped the end of the input now says so. When an engine returns no audio for part of the text the result sounds clean and is simply short, so the only way to notice was to read along; the backend log now names the sentences that produced nothing. (#1330)
+- Dubbing: a re-rendered line that quietly came back in a default voice instead of the cloned one now says why in the backend log — the clone clips are extracted per job and a saved dub outlives them, so regenerating after cleanup loses the reference with no error. (#1331)
 
 ### Docs
 
--- a/backend/api/routers/dub_generate.py
+++ b/backend/api/routers/dub_generate.py
@@ -203,6 +203,63 @@ def _speaker_key_for_segment(job: dict, sid) -> str | None:
     return None
 
 
+#: Segment ids already warned about, per job, so a 300-segment dub whose clip
+#: directory was cleaned logs once per segment rather than once per retry.
+_MISSING_REF_WARNED: dict = {}
+
+
+def warn_if_ref_missing(ref_audio, *, job_id: str = "", seg_id="", where: str = "dub"):
+    """Say so when a clone reference points at a file that is gone (#1331).
+
+    Reported as: re-dubbing a single sentence loses the cloned voice, while
+    re-dubbing everything keeps it.
+
+    Clone references are FILE PATHS into the job's extracted-clip directory,
+    and the whole job dict — those paths included — is persisted to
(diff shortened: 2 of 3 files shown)


debpalash:
Valid, and it is the more interesting half of the design: the de-duplication meant to stop repetition was swallowing new facts.

The preview endpoint has no segment identity to pass — it is a "render this text" call — so every preview shared the key `"preview"`, and only the **first** missing reference in a job was ever reported. Every later one, with a different path, was silenced.

Keyed on `(segment, path)` now, which fixes it without an API change and is also more correct on the render path: a segment rebound to a *second* missing clip is no longer mistaken for the one already reported. A distinct path is distinct information wherever it comes from.

I also took your suggestion of carrying the identity, as an **optional** `segment_id` on `SegmentPreviewRequest` — diagnostic-only, defaulted to `None`, so existing callers are unaffected, and a caller that supplies it gets the line named instead of a bare `preview`.

Three tests. The distinct-paths one fails against the segment-only key; the repeat-suppression one keeps passing, so the fix cannot be a blanket removal of the de-duplication it exists for.

PR #1362, opened 2026-08-04
fix(release): rebuild + cryptographically verify the preview updater manifest (#1327)

debpalash:
Supersedes #1327, which is @Pinkers01's work — their commit is preserved here with authorship intact, and my three review fixes sit on top. I could not push to their fork despite `maintainerCanModify`, and the PR had gone six hours without a response, so this unblocks it rather than leaving a real outage fix sitting.

**@Pinkers01 — the credit is yours.** If you'd rather push the fixes to your own branch and land #1327 instead, say so and I'll close this.

## The outage

Since ~2026-07-13 every matrix leg logged *"Signature not found for the updater JSON. Skipping upload…"* — `tauri-action` uploaded the bundles and their `.sig` companions but never refreshed `latest.json`. Meanwhile the version-less macOS updater bundles are deleted and replaced every night by the stale-bundle cleanup step, so the manifest's darwin signatures stopped matching the published files.

Net effect: macOS Preview users hit *"The signature verification failed"* on **every** update attempt, for over two weeks, while CI stayed green. Parity and version-format checks both passed the whole time, because neither one looks at a signature.

## The fix (Pinkers01)

Rebuild `latest.json` **after** the matrix completes — one job, no per-leg race — from the release's real assets and their `.sig` companions, then clobber-upload. The manifest can no longer drift from the files it describes regardless of what `tauri-action`'s own updater-JSON path does or skips.

Plus the check that was missing: every signature in the manifest is verified with the updater pubkey baked into `tauri.conf.json`, against the bytes actually published. That is the gate that would have caught this on day one.

## My three fixes on top

1. **The manifest could still describe artifacts from two different runs (blocking).** `newest()` picked the AppImage and the MSI independently by run number, then took whichever `-N` was larger as *the* version. A matrix where one leg failed or was re-run would publish a manifest advertising `X.Y.Z-5` while handing Windows users the `-4` MSI — the same manifest/artifact drift this job exists to end, reintroduced by the fix for it. Now requires `n1 == n2` and fails loudly: leaving the previous manifest in place is a visible, already-understood state; shipping a mismatched one is not. (The darwin tarballs carry no run number, so they cannot be checked this way — the signature verification in the next step is what pins those.)
2. **`persist-credentials: false`** on the checkout — nothing here pushes to git, and the following steps shell out to `gh` and install from PyPI.
3. **Floor-pinned the `cryptography` install** (`>=42`, matching `docs-drift.yml`'s `pyyaml>=6`). This step decides whether a signed release manifest is trustworthy; it is the one dependency here worth a bound.

## Tests

`tests/test_release_preview_manifest_rebuild.py` extracts the step body from `release.yml` — so it cannot drift from the workflow — and runs it against a stubbed `gh` with synthetic asset lists:

- matched run numbers produce the expected manifest, with every platform stable serves;
- mismatched run numbers fail and upload **nothing** (fails against the pre-fix version);
- a missing `.sig` companion fails rather than publishing an entry no client can verify;
- a missing darwin tarball fails rather than silently dropping Intel Mac from the manifest.

One thing I checked and it is correct as written: the verify step re-downloads `latest.json` over HTTP *after* the rebuild, so it validates the new manifest rather than a cached one.

## After merge

This only takes effect on the next preview build. Worth watching that run's **step log** rather than just job-green, since the whole failure mode here was a step that skipped its work and said so quietly.

The preview workflow now rebuilds `latest.json` from published assets and signatures, verifies each signature with the updater public key, and rejects mismatched runs, stale macOS assets, or missing artifacts before upload. This fixes stale or mismatched manifests that caused macOS Preview update verification failures. Review platform asset selection, freshness checks, and signature verification edge cases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -919,6 +919,87 @@ jobs:
     permissions:
       contents: write
     steps:
+      # Needed by the manifest rebuild + signature check below: the updater
+      # pubkey lives in frontend/src-tauri/tauri.conf.json.
+      - uses: actions/checkout@v4
+
+      # ── Rebuild the preview updater manifest from what is ACTUALLY published ──
+      # Since ~2026-07-13 every matrix leg has logged "Signature not found for
+      # the updater JSON. Skipping upload..." — tauri-action uploads the bundles
+      # + .sig companions but never refreshes latest.json. Meanwhile the macOS
+      # updater bundles (version-less filenames) are deleted + replaced every
+      # night by "Clear this arch's stale preview updater bundle", so the
+      # manifest's darwin signatures stopped matching the published files:
+      # macOS Preview users hit "The signature verification failed" on every
+      # update attempt (latest.json frozen at 2026-07-13, tar.gz replaced
+      # nightly).
+      #
+      # Root fix: after the matrix completes, rebuild latest.json HERE — one
+      # job, no per-leg race — from the release's real assets and their .sig
+      # companions, then clobber-upload. The manifest can no longer drift from
+      # the files it describes, regardless of what tauri-action's own
+      # updater-JSON path does or skips.
+      - name: Rebuild preview updater manifest from published assets
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          gh release view preview --repo "$REPO" --json assets \
+            -q '[.assets[].name]' > /tmp/assets.json
+          WORK=$(mktemp -d)
+          python3 - "$WORK" <<'PY'
+          import datetime, json, os, re, subprocess, sys
+          work = sys.argv[1]
+          repo = os.environ["REPO"]
+          names = json.load(open("/tmp/assets.json"))
+
+          def sig(name):
+              subprocess.run(["gh", "release", "download", "preview", "--repo", repo,
+                              "-p", name + ".sig", "-D", work], check=True)
+              return open(os.path.join(work, name + ".sig")).read().strip()
+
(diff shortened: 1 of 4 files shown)


debpalash:
Both findings valid, and both are cases the existing checks structurally could not see.

**greptile P1 — verification ran *after* the upload.** That is the more serious one: a manifest failing the check was already live and stayed served, with the job merely going red. Every macOS Preview user would stay broken until a human noticed — the same two-week silence this PR exists to end, in a new costume. Verification now runs against the file about to be published, and the upload is the last statement in the step. A refusal leaves the previous manifest in place, which is a visible, already-understood state.

**CodeRabbit — the darwin entries were not bound to this run.** Also right, and I had explicitly waved this away in my earlier comment ("the signature check pins them to the published bytes"). That was wrong: signature verification pins the tarball to *its own* `.sig`, and a stale pair matches perfectly. A run whose macOS legs never uploaded would advertise this version while serving Mac users the previous build — and because those clients keep reporting the old version, the updater re-offers the same update forever. They are bound by upload time now, with two minutes of slack so legs that finish apart do not fail a healthy run.

**One structural change on the back of it.** Three findings in a row here have been about *which artifacts may be described together*, so the selection rules move out of the YAML heredoc into `scripts/build_preview_manifest.py`. `build_manifest` is pure — assets in, manifest out, `ManifestRefused` on anything it will not describe — which means the tests call the rules directly instead of extracting a heredoc and stubbing a shell. The previous test file was asserting against `gh` stubs, which is a lot of machinery to test a `max()`.

14 tests. Both new refusals fail against the previous version, and two of them pin that the workflow still calls the module and still uploads last — an inline copy would pass everything else and ship the original bug.

Issue #1223, opened 2026-07-22
[Crash] Backend died (exit code 1)


xipb14:
## Environment

**Version:** `0.4.0`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36 Edg/151.0.0.0`
**OS:** `Windows-10-10.0.22631-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 3060 (8 GB VRAM)`
**CPU:** `Intel64 Family 6 Model 60 Stepping 3, GenuineIntel`
**RAM:** `15.9 GB`
**Disk free:** `51.2 GB`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 1 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-07-21T21:40:28.000Z (8 h ago)
**Exit:** `exit code 1`
**Uptime before crash:** 58 s
**Backend version:** `0.4.0`

```
… (truncated)
ble_http_manager] StreamableHTTP session manager started
2026-07-22 00:40:14,235 INFO [omnivoice.api] MCP server mounted at /mcp
INFO:     Application startup complete.
ERROR:    [Errno 10048] error while attempting to bind on address ('127.0.0.1', 3900): обычно разрешается только одно использование адреса сокета (протокол/сетевой адрес/порт)
INFO:     Waiting for application shutdown.
2026-07-22 00:40:14,248 INFO [omnivoice.api] Shutdown: cleaning up…
2026-07-22 00:40:14,249 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager shutting down
2026-07-2026-07-22 00:40:17,359 WARNING [omnivoice.setup.models] is_cached: scan_cache_dir failed (OSError: [WinError 1392] Файл или папка повреждены. Чтение невозможно: 'G:\\OS\\models--Systran--faster-whisper-medium\\snapshots\\08e178d48790749d25932bbc0822026-07-22 00:40:17,593 INFO [omnivoice.model] Model load aborted: shutd
Loading weights:   0%|          | 0/313 [00:00<?, ?it/s]
Loading weights:   0%|          | 1/313 [00:00<04:36,  1.13it/s]
[...]

Issue #1357, opened 2026-08-04
[Install] Colab notebook install (Cell 2) leaves torch/torchvision mismatched → RuntimeError: operator torchvision::nms does not exist


HanzlahCh:
### Where did it fail?

Python / uv environment bootstrap

### The error

```text
Cell 2 ("Clone + install") completes without error, but the import sanity check that
follows it fails with:

RuntimeError: operator torchvision::nms does not exist

...raised from transformers trying to import HiggsAudioV2TokenizerModel, which pulls
in torchvision via transformers/image_utils.py. Full traceback:

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/transformers/utils/import_utils.py", line 2305, in __getattr__
    module = self._get_module(self._class_to_module[name])
  ...
  File "/usr/local/lib/python3.12/dist-packages/torchvision/_meta_registrations.py", line 163, in <module>
    @torch.library.register_fake("torchvision::nms")
  ...
RuntimeError: operator torchvision::nms does not exist

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.12/dist-packages/transformers/utils/import_utils.py", line 2393, in __getattr__
    raise ModuleNotFoundError(
ModuleNotFoundError: Could not import module 'HiggsAudioV2TokenizerModel'. Are this
object's requirements defined correctly?

Root cause: Cell 2 installs torch 2.8.0+cu128 and torchvision 0.26.0+cu128 together,
but that torchvision version doesn't correspond to torch 2.8.0 (expected ~0.23.x) —
they end up internally mismatched.
```

### Operating system

Linux (other / from source)

### How are you installing it?

From source (uv sync)

### Version

v0.4.2 — OmniVoice Studio

### Network conditions (model/dependency downloads)

Not sure

### Logs / diagnostic bundle

```text
Reproduced twice from a completely fresh Colab runtime (Runtime → Restart, then
re-ran Cell 2 from scratch both times) — same mismatch both times, so this isn't a
one-off caching artifact.

Confirmed fix, applied after Cell 2:

!/usr/bin/python3 -m pip uninstall -y torch torchvision torchaudio
!/usr/bin/python3 -m pip install torch==2.8.0 torchvision==0.23.0 torchaudio==2.8.0 \
    --index-url https://download.pytorch.org/whl/cu128

[...]

PR #1370, opened 2026-08-04
fix(startup): explain a lost port race instead of exiting 1 in silence (#1364)

### LINKED ISSUE #1364: [Crash] Backend died (exit code 1)
## Environment

**Version:** `0.4.2`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cpu`
**CPU:** `Intel64 Family 6 Model 154 Stepping 3, GenuineIntel`
**RAM:** `15.6 GB`
**Disk free:** `21.7 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Last backend response:** 1 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-08-04T13:41:41.000Z (20 min ago)
**Exit:** `exit code 1`
**Uptime before crash:** 13 s
**Backend version:** `0.4.2`

```
… (truncated)
nerate.types
2026-08-04 19:11:38,063 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.constraints
2026-08-04 19:11:38,064 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.defaults
2026-08-04 19:11:38,064 INFO [alembic.runtime.plugins] setup plugin alembic.autogenerate.comments
2026-08-04 19:11:38,316 INFO [omnivoice.model] Preload skipped: k2-fsa/OmniVoice is not installed locally — the model will load (and download if requested) on first use.
2026-08-04 19:11:38,317 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager started
2026-08-04 19:11:38,318 INFO [omnivoice.api] MCP server mounted at /mcp
INFO:     Application startup complete.
ERROR:    [Errno 10048] error while attempting to bind on address ('127.0.0.1', 3900): only one usage of each socket address (protocol/network address/port) is normally permitted
INFO:     Waiting for application shutdown.
2026-08-04 19:11:38,320 INFO [omnivoice.api] Shutdown: cleaning up…
2026-08-04 19:11:38,321 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager shutting down
2026-08-04 19:11:38,623 INFO [omnivoice.api] Shutdown: done.
INFO:     Application shutdown complete.
```

## Recent actions

```
18:59:21 view:launchpad
```

debpalash:
Fixes #1364. Follow-up to #1223 — same crash, still unexplained in one case.

## What #1223 left open

#1223 gave a port conflict a dedicated exit code (78) and handles the race between our pre-probe and uvicorn's real bind by **re-probing** after uvicorn dies.

That only helps while the other process is *still* holding the port. The common case is an orphaned backend from the previous session which is itself shutting down: it releases the port between uvicorn's failed bind and our re-probe, the probe says "free", and the user gets a bare `exit code 1` — for a crash we had already fully diagnosed.

The report shows the tell-tale ordering:

```
INFO:     Application startup complete.          <- uvicorn's lifespan runs BEFORE the bind
ERROR:    [Errno 10048] error while attempting to bind on address ('127.0.0.1', 3900)
INFO:     Waiting for application shutdown.
```

...then `Backend died (exit code 1)`.

## The fix

uvicorn already hands us the answer. Its startup does `logger.error(exc); sys.exit(1)` with the **OSError itself** as the record's message, so the errno is available as an object — no locale-dependent string matching, which is precisely the trap #1223 exists to avoid (that report was in Russian). Observe that record, believe it first, keep the re-probe as the fallback.

The watcher also pins `uvicorn.error` to at most `ERROR`: a filter only runs on records the logger actually emits, so a higher level would drop the bind failure and silently restore the unexplained exit 1. No-op today — uvicorn defaults to INFO and nothing here raises it — it just stops the mechanism being disarmed at a distance later.

## Tests

Two regressions, both driving the **real** uvicorn:

- The race, simulated deterministically by making both probes report the port free while it is genuinely held — exactly the state the race leaves us in. Verified: **exit 1 without the watcher, 78 with it.**
- The watcher ignores unrelated `OSError`s (EACCES, ECONNREFUSED) and plain string messages, so a permission failure doesn't get mislabelled as a port conflict.

Two things worth flagging in how these are written:

1. The assertion is on **our** wording, not `"already in use"`. I checked: that phrase is also in uvicorn's own English log line, so asserting on it passes against the *unfixed* build — and it would be the very locale-dependent match #1223 forbids.
2. The pre-existing extraction test needed `import logging` added to its generated script, since the guard region it copies out of `main.py` now defines a `logging.Filter`.

`tests/test_port_in_use_exit.py`: 11 passed. `backend/tests/`: 237 passed.

The backend now detects locale-independent `EADDRINUSE` failures from `uvicorn.error` and reports the dedicated port-conflict error when a bind race frees the port before re-probing. Regression tests cover bind races, unrelated errors, uvicorn logging behavior, and production wiring, with 11 port-conflict tests and 237 backend tests passing. Human review should verify that the guarded `uvicorn.run()` keeps a log level no higher than `ERROR`, because the watcher depends on that behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 ### Fixed
 
 - Every subprocess TTS engine would have turned a stereo render into noise: the mono downmix always averaged axis 0, which is time rather than channels for channels-last audio. Unreachable today since every engine returns mono, fixed in all five before it isn't. (#1328)
+- A port conflict that resolves itself while the backend is dying no longer reports a bare "Backend died (exit code 1)" — the conflict is named even when the other process has already let the port go. (#1364, #1223)
 - Fresh installs failing to import `transformers.HiggsAudioV2TokenizerModel` with "RuntimeError: operator torchvision::nms does not exist" are fixed by pinning `torchvision==0.23.0` to match `torch 2.8.0` — thanks @HanzlahCh! (#1358, #1357)
 - Every RTX 40-series card (4060–4090) was declared unsupported and silently run on the CPU. The compatibility gate demanded an exact `sm_89` match, but PyTorch ships `sm_86` kernels that already cover Ada. (#1285)
 - Under-provisioned hardware is now flagged **before** a synthesis starts instead of after the full compute budget expires. (#1240, #1246, #1248, #1277, #1283, #1284)
--- a/backend/main.py
+++ b/backend/main.py
@@ -1412,6 +1412,28 @@ if __name__ == "__main__":
         )
         sys.exit(_EXIT_PORT_IN_USE)
 
+    class _BindErrorWatcher(logging.Filter):
+        """Remembers the EADDRINUSE uvicorn logged on its way out (#1364).
+
+        uvicorn's startup does ``logger.error(exc); sys.exit(1)`` with the
+        OSError itself as the record's message, so the errno is available as an
+        object — no locale-dependent string matching. Passing every record
+        through untouched; this only observes.
+        """
+
+        def __init__(self) -> None:
+            super().__init__()
+            self.bind_error: "OSError | None" = None
+
+        def filter(self, record: logging.LogRecord) -> bool:
+            msg = record.msg
+            if isinstance(msg, OSError) and (
+                msg.errno in (48, 98, 10048)
+                or getattr(msg, "winerror", None) == 10048
+            ):
+                self.bind_error = msg
+            return True
+
     # #1223: uvicorn does NOT let a bind failure reach the caller — it logs the
     # raw errno and raises SystemExit(1) from inside its startup, so an
     # `except OSError` around uvicorn.run() never fires (verified, not assumed).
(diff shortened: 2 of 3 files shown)


debpalash:
Both findings addressed — and the P1 was half right in a way worth writing down, because it caught a real mistake via a mechanism that turned out not to be the one described.

**greptile P1 — "uvicorn removes the bind watcher."** Measured against the installed uvicorn rather than reasoned about:

```
before:                                  level 40  filters 1
after configure_logging(critical):       level 50  filters 1   filter still attached: True
after configure_logging(default):        level 20  filters 1
```

`dictConfig` replaces a logger's **handlers** and leaves its **filters** alone, so the filter survives — and the end-to-end test was already driving the real `uvicorn.run` and getting exit 78, which is independent evidence.

But the conclusion was right for a different reason, and it caught something I got wrong: uvicorn resets the logger's **level** from its config during startup, *after* the defensive `setLevel(ERROR)` I had added. That guard was dead code offering false assurance. Removed.

The actual precondition is that the guarded `uvicorn.run()` must not raise `log_level` above ERROR, since a filter only runs on records the logger emits. Now pinned by a test — scoped to the guarded call, because the `--health-check` smoke path sets `log_level="warning"`, which is below ERROR and irrelevant.

Both uvicorn behaviours are pinned by tests that measure the installed version, so an upgrade that starts clearing filters, or a change that quietens the serve call, fails loudly instead of silently restoring the unexplained exit 1.

**CodeRabbit — the tests rebuild the guard from extracted source, so they'd pass if main.py stopped wiring it.** Correct. Extracting the whole block into a callable is a larger refactor of startup than this fix justifies, so I've pinned the production wiring directly instead: the class exists, the filter is attached, and the result is consulted. That closes the specific hole — main.py losing any of the three now fails.

15 tests passing.

debpalash:
Second round, both addressed.

**log_level checked by AST, not regex** — valid and worth taking. The regex only recognised string literals, so `log_level=settings.level` or any computed value matched nothing and the assertion passed while verifying nothing. That is the same class of bug as #1357's pin that did not apply: a check that looks present and is inert. A non-literal is now an explicit failure rather than a silent skip, and the guarded call is located by walking to the `addFilter` and taking the `uvicorn.run` after it, rather than by source order.

Verified against three mutations of `main.py`:

```
literal critical    -> FAILED
computed value      -> FAILED
literal warning     -> passed   (below ERROR, must not fail)
```

**"Exercise the production bind-failure wiring"** — repeated from the first round; my answer is unchanged, and I want to be explicit that it is a judgment call rather than an oversight.

The block lives inside `main.py`'s argument-dispatch body, so making it directly invocable means extracting the port guard, the watcher, and the serve call into a module-level function — a refactor of process startup, in a PR whose job is to stop one crash being unexplained. The risk of that refactor is larger than the risk it removes.

What the finding correctly identifies is that the end-to-end tests rebuild the guard from extracted source and would survive `main.py` dropping the wiring. That specific hole is closed directly: `test_main_actually_wires_the_watcher` asserts the class exists, the filter is attached, and the result is consulted. Losing any of the three now fails.

Extraction is the better long-term shape and I'd take it as its own change against startup, not smuggled into this one.

15 tests passing.

Issue #1376, opened 2026-08-05
[Crash] Backend died (exit code 1)


yashrio:
## Environment

**Version:** `0.4.2`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36 Edg/151.0.0.0`
**OS:** `Windows-10-10.0.26100-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce GT 1030 (2 GB VRAM)`
**CPU:** `Intel64 Family 6 Model 94 Stepping 3, GenuineIntel`
**RAM:** `31.9 GB`
**Disk free:** `22.4 GB`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 1 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-08-05T08:32:03.000Z (29 min ago)
**Exit:** `exit code 1`
**Uptime before crash:** 28 s
**Backend version:** `0.4.2`

```
… (truncated)
alash.omnivoice-studio\project\.venv\Lib\site-packages\transformers\utils\import_utils.py", line 2184, in __getattr__
    raise ModuleNotFoundError(
ModuleNotFoundError: Could not import module 'GenerationMixin'. Are this object's requirements defined correctly?

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "~\AppData\Local\com.debpalash.omnivoice-studio\project\backend\services\model_manager.py", line 1426, in _load_model_sync
    OmniVoice = _lazy_omnivoice()
                ^^^^^^^^^^^^^^^^^
  File "~\AppData\Local\com.debpalash.omnivoice-studio\project\backend\services\model_manager.py", line 40, in _lazy_omnivoice
    from omnivoice.models.omnivoice import OmniVoice as _OV
  File "~\AppData\Local\com.debpalash.omnivoice-studio\project\omnivoice\models\omnivoice.py", line 45, in <module>
    from transformers import (
  File "~\AppData\Local\com.debpalash.omnivoice-studio\project\.venv\Lib\site-packages\transformers\utils\import_utils.py", line 2184, in __getattr__
    raise ModuleNotFoundError(
ModuleNotFoundError: Could not import module 'AutoFeatureExtractor'. Are this object's requirements defined correctly?
```

## Recent actions

```
[...]

PR #1377, opened 2026-08-05
fix(errors): the transformers-import advice names torchvision now (#1376)

debpalash:
For #1376 (and the class behind #1357/#1294).

## The gap

The `TRANSFORMERS_IMPORT` hint and the ASR pipeline error both said: reinstall `torch torchaudio transformers`.

**torchvision was the one package omitted** — and it is the package whose ABI mismatch actually produces this exact wording. `operator torchvision::nms does not exist` gets wrapped by transformers' lazy `__getattr__` into `Could not import module 'AutoFeatureExtractor'`, so the surface error names transformers while the broken package is torchvision. A user following our advice to the letter reinstalled everything except the thing that was broken.

That is #1376: a Windows desktop install dying on `GenerationMixin` / `AutoFeatureExtractor` at startup, where reinstalling transformers cannot help.

## The fix

Both surfaces (the shared hint in `core/failure.py`, the ASR pipeline error in `asr_backend.py`) now:

- name the torch/torchvision mismatch as a cause, noting it fails with exactly this wording
- prescribe the **pinned** reinstall: `uv pip install --reinstall --constraint deploy/torch-constraints.txt torch torchaudio torchvision transformers`

Pinned deliberately: an unpinned reinstall of the trio can itself resolve a drifted pair — the exact bug the constraint file (#1369) exists to prevent. The constraint pins carry no local segment, so CUDA/ROCm vendor builds match rather than being replaced.

## Docs (same-PR sync)

`docs/install/troubleshooting.md` gains §1a-bis: same wording, different cause. The existing §1a covers only the missing-file cause, and its fix (`--reinstall transformers`) genuinely does nothing for the mismatch — a user who hit §1a, tried it, and saw no change had no next step.

## Tests

Regression test pins that the hint names torchvision and the constraint file. Related suites: 175 passed (asr/transformers selection), changelog style green.

Updated Transformers import troubleshooting to identify `torchvision` ABI mismatches and recommend pinned package reinstallation. Added matching ASR error guidance, documentation, changelog text, and regression coverage. Review risk is limited to verifying that all pinned versions remain compatible.


debpalash:
greptile's P1 is confirmed and fixed — good catch, and it invalidated the exact command I told the #1376 reporter to run.

Verified against `tauri.conf.json`: the bundle ships `pyproject.toml`, `uv.lock`, `omnivoice/`, `backend/` — **no `deploy/`**. So the `--constraint deploy/torch-constraints.txt` form fails with *file not found* on desktop installs, which are exactly the users who hit this error.

Every advice surface now carries the **literal pinned versions** instead, guarded by a lockstep test that reads `deploy/torch-constraints.txt` and fails if any surface (hint, ASR error, docs) advises different versions — bump the pins, the build stays red until the advice matches. Verified fail-on-drift by bumping torchvision in a copy. Shipping the file into the bundle would have needed Rust + packaging changes and still left every existing install broken; the literals work everywhere immediately.

I've also corrected the command in my #1376 comment.

debpalash:
Second round, two of four taken:

**greptile — target the venv explicitly.** The stated mechanism is off (uv does discover `.venv` from the working directory, which is why the advice says "in the project folder") but the hardening is free: `--python .venv` makes the command survive being run from the wrong place or with another environment active. Taken on all three surfaces.

**CodeRabbit — assert the command, not substrings.** Valid. The lockstep test now asserts the exact command, whitespace/quote-normalized so `asr_backend.py`'s split string literals count as what the user sees. Verified failing on both mutations: a wrong pin *inside* the command with the right pin in a comment, and a reworded command.

**Declined — "include the transformers pin in the lockstep check."** The advice deliberately leaves `transformers` unpinned, as does `deploy/torch-constraints.txt` — its version floats in pyproject, and pinning it in a repair command would fight the next `uv sync`. There is no transformers pin to be in lockstep *with*.

**No action — the Python-availability chain** resolved to the same venv-targeting concern greptile raised; covered by `--python .venv`.

PR #1381, opened 2026-08-05
fix(dub): a job's clone references live in its own directory (#1331)

debpalash:
Root cause fix for the identified class behind #1331, which #1361 made visible but did not fix.

## The mechanism, concretely

1. Dub the same video twice → `find_cached_job` (content-hash cache) points the new job's `vocals_path` into the **old** job's directory. Cache working as designed.
2. `dub_core` then extracted the new job's clone references into `os.path.dirname(vocals_path)` — **into the old job's directory too**.
3. Delete that older dub from history — an ordinary, sanctioned action that `rmtree`s its dir.
4. Every clone reference of the newer job now dangles. Each single-segment regen silently renders in the default voice (the engine falls back without error — the #1361 diagnostic is what logs it now), and a full re-dub "fixes" it only because prep re-extracts.

Which matches the report exactly: *only a full re-dub keeps the cloned voice*.

## The fix

One argument: clones are written into the **current** job's own dir (`safe_job_dir(job_id)`, `dirname(vocals)` as fallback for the degenerate case).

## Known residual, deliberately out of scope

`vocals_path` itself is still shared across jobs by the cache, so deleting the older job still costs the newer one its stems for re-separation-dependent features (export-stems, re-transcribe). Fixing that means refcounting or copying multi-hundred-MB artifacts — a different tradeoff, worth its own issue if it bites. The clone references — the thing the user hears on every regen — no longer ride on a neighbour's lifetime.

## Tests

- Two pin the call site by AST (out_dir not derived from the vocals path; anchored to the job's own id) — **both fail against main**.
- One functional: vocals in job A's dir, clones requested for job B, `rmtree` job A — references survive.

Adjacent dub suites: 31 + 34 passed.

The fix stores clone and per-segment references in the current job directory, and preserves directories that other saved jobs reference when history entries are deleted. This prevents cached-job deletion from orphaning references needed by newer jobs while still removing unreferenced directories. AST, functional, and deletion tests cover extraction paths, cross-directory references, and JSON-escaped Windows paths; shared `vocals_path` behavior remains unchanged.


debpalash:
Both findings taken — and both reviewers caught the same real miss: `extract_segment_refs` is the **default** path (`per_segment_refs=true`), so the first version of this fix covered the less-travelled branch while the reported behaviour survived on the main road.

Same one-argument change at the second call site. The AST test now sweeps **every** extraction call in `dub_core` rather than pinning one by name, so a third copy added later inherits the rule instead of reintroducing the class — plus the functional test for segment refs (clips extracted for job B from job A's vocals survive job A's deletion).

5 tests, the sweep failing against the pre-push state.

debpalash:
Scope note: a third commit lands the **deletion half** of the same class, because the extraction fix alone only protects jobs created from now on.

Existing users' jobs already carry cross-directory references, and `vocals_path` is shared across cache-hit jobs by design — for all of those, deleting the older history entry still `rmtree`'d files a surviving job renders from. `delete_single_dub_history` now scans every other row's persisted `job_data` for the dir prefix first (textual on purpose: the path-bearing keys have grown before and a scan cannot fall behind the schema; trailing-separator needle so prefix-sharing ids cannot collide; JSON-escaped Windows spelling matched too). If anyone still references the dir: row deleted — the entry disappears from the UI as asked — directory kept, holder logged. Clear-all untouched, since deleting everything leaves no survivors to protect.

8 new tests including both router-level ends. With this, #1331 is covered from both directions: new jobs never write into a neighbour's dir, and old jobs' neighbours can no longer be deleted out from under them.

Issue #982, opened 2026-07-06
[Bug] Windows: dictation pill steals focus → auto-paste lands in OmniVoice, pill never dismisses


greenmojo2:
## Environment
- OS: **Windows 11** (x64)
- OmniVoice Studio: **v0.3.11** (official MSI install)
- Dictation engine: `sherpa-parakeet-tdt-v3` (live transcription itself works and is accurate)

## Symptom
Using the global dictation shortcut on Windows:
1. Pressing the shortcut moves focus to the **OmniVoice Studio** window (the app/pill activates and grabs the foreground).
2. Because the cursor is no longer in the target app, the final **auto-paste (`Ctrl+V`) types into OmniVoice** instead of the document where the cursor was.
3. The pill widget **never auto-dismisses** after the final — it stays on screen.

Only the focus / auto-paste / dismiss behavior is broken; the transcription is correct.

## Root cause
`simulate_paste` (`frontend/src-tauri/src/commands.rs`) writes the transcript to the clipboard and then sends `Ctrl+V` via `enigo` to whatever window is in the foreground. The pill is *intended* to be unfocused — `lib.rs` even comments *"deliberately unfocused so the auto-paste lands in the focused app"* — and this holds on macOS.

On **Windows**, however, showing an always-on-top top-level `WebviewWindow` **activates** it (Win32 gives a normal top-level window foreground activation on show). The pill is built with `.always_on_top(true).skip_taskbar(true)` but nothing marks it **non-activating**, so it steals foreground → the synthesized `Ctrl+V` lands in OmniVoice, and the pill's state machine stalls before the auto-dismiss.

This is the Windows counterpart of **#287** ("macOS auto-paste — don't steal focus, write clipboard natively"), which was implemented for macOS only.

## Suggested fix
Make the Windows pill non-activating so it never takes the foreground:
- Apply `WS_EX_NOACTIVATE` to the pill's HWND (e.g. `SetWindowLongPtrW(hwnd, GWL_EXSTYLE, cur | WS_EX_NOACTIVATE)` via the `windows` crate on the raw HWND from `window.hwnd()`), **and/or** show it with `ShowWindow(SW_SHOWNOACTIVATE)` rather than a focus-activating `show()`.
- Confirm the final transcript still routes through the auto-dismiss path so the pill hides after paste.

Happy to test a patched build (Windows 11 / RTX 3070).

debpalash:
[...]

Issue #287, opened 2026-06-06
[Bug] macOS: Dictation auto-paste lands in the widget / pastes stale clipboard content


geektf:
## Environment

- OmniVoice Studio **0.3.5**, built from source (current `main`)
- macOS 26 (Darwin 25.5.0), Apple Silicon M4 Pro, 48 GB RAM
- Accessibility permission granted (enigo logs *"The application has the permission to simulate input"*)
- ASR engine: mlx-whisper (whisper-large-v3-turbo)

## Summary

Dictation via the global shortcut (`⌘+⇧+Space`) transcribes correctly, the widget reports success — but the dictated text never appears in the target app (tested with Microsoft Word and Notes). After partially fixing the focus issue, a second, related bug surfaces: the **previous** clipboard content is pasted instead of the transcript. Both bugs together make dictation auto-paste fully unusable on macOS.

## Bug 1 — Widget steals focus, so the simulated ⌘V lands in the widget

In `frontend/src-tauri/src/lib.rs`, the `ShortcutState::Pressed` handler shows the widget and then calls:

```rust
let _ = win.show();
let _ = win.set_focus();
```

On macOS this makes the widget window the frontmost app. When `simulate_paste()` (enigo) later sends `⌘+V`, the keystroke is delivered to the **widget**, not to the app the user is dictating into. The widget reports success, the target document stays empty. The transcript *is* in the clipboard at this point, so a manual `⌘+V` works — which is what masked the bug for us initially.

**Fix:** skip `set_focus()` on macOS so the target app keeps focus:

```rust
let _ = win.show();
// Don't steal focus on macOS: the simulated ⌘V from
// simulate_paste() must land in the app the user is
// dictating into — focusing the widget would swallow it.
#[cfg(not(target_os = "macos"))]
let _ = win.set_focus();
```

## Bug 2 — Unfocused widget cannot write the WebView clipboard (fails silently)

Once the widget no longer takes focus, `CaptureWidget.jsx` → `copyText()` breaks: `navigator.clipboard.writeText()` requires a focused document in WKWebView, and the `document.execCommand('copy')` fallback also fails in an unfocused window. `copyText()` returns `false` instead of throwing, so `simulate_paste` runs anyway and pastes **whatever was previously in the clipboard** — silently inserting unrelated content into the user's document.

**Fix:** write the transcript to the clipboard **natively (OS-side)** in the `simulate_paste` command itself, which needs no window focus. Using the `arboard` crate:

[...]

PR #1399, opened 2026-08-06
fix(dictation): stop the widget stranding an empty square, repair the swept data dirs

### LINKED ISSUE #1398: [Bug] Dictation hotkey leaves a blank square stuck on the desktop and does not dictate
Reported with a screenshot: pressing the dictation hotkey leaves a bare dark rectangle on screen — no pill, no text, no way to dismiss it — and no dictation happens.

## Root cause

Three defects compound.

**1. The widget window paints an opaque background.** It is declared `transparent: true` + `decorations: false` in `tauri.conf.json`, but it loads the same `index.html` as the main window, so `body { background-color: var(--chrome-bg) }` (`index.css`) painted an opaque, square-cornered rect across all 300x64 of it. Window transparency only shows through if every ancestor box is transparent.

**2. Hotkey presses could be dropped.** The `tray-dictate` listener effect in `CaptureWidget.jsx` depended on `[state]`, so every state change tore the Tauri subscription down and re-attached it across an `await import()` + `await listen()`. A press landing in that gap was lost.

**3. A lost press strands the window.** `lib.rs` shows the widget window *before* it emits `tray-dictate`, and `dismiss()` is the only path that hides it — reachable only from the X button, Esc, or a post-session timer. None can fire for a session that never started: `state === 'idle'` renders `null` so there is no X, and Esc never arrives because the widget deliberately refuses focus on macOS/Windows (#287, #982).

Net effect: window shown, pill idle, nothing rendered, opaque body — an unrecoverable square.

## Fix

- Scope a transparent background to `html`/`body`/`#root` behind a `data-window="widget"` marker set from the window label before first render.
- Subscribe the tray listener exactly once, reading state through a ref (the pattern the `mode`/`enabled` refs beside it already use).
- Enforce the invariant directly rather than patching call sites: idle + a visible widget window reconciles to hidden. Covers a press while dictation is disabled, a dropped event, and any path added later.

Regression tests fail before and pass after.

debpalash:
Closes #1398.

Two related fixes: the dictation bug the screenshot showed, and a data-dir regression the brand sweep introduced that I found while auditing the rename leftovers.

## The stranded square (#1398)

Three defects compounded into one unrecoverable artefact.

| # | Defect | Effect |
|---|---|---|
| 1 | `body { background-color: var(--chrome-bg) }` applies to the widget window, which loads the same `index.html` | `transparent: true` is defeated — an opaque, square-cornered rect |
| 2 | The `tray-dictate` listener effect depended on `[state]`, re-attaching across `await import()` + `await listen()` | A press landing in that gap is lost |
| 3 | Rust shows the window **before** emitting; `dismiss()` is the only hide path | A lost press strands the window |

Defect 3 is why it was unrecoverable: `idle` renders `null`, so there is no X button, and Esc never arrives because the widget deliberately refuses focus on macOS/Windows (#287, #982).

Rather than patch each call site, the third fix enforces the invariant itself — idle plus a visible widget window reconciles to hidden. That covers a press while dictation is disabled, a dropped event, and any path added later. It is scoped to the window labelled `widget`, so the in-app pill can never hide the main window.

## The data dirs the sweep broke

The rename rewrote three path literals that had to stay `OmniVoice`, because they name where the backend actually writes:

- `smoke-test.sh` resolved `%APPDATA%/VoiceStudio` on Windows and an XDG path on Linux — it verified a directory the backend never writes, so it can pass while nothing was produced.
- `desktop-prod.sh` wiped `%APPDATA%/VoiceStudio`, silently no longer clearing backend state on Windows.

Both are invisible on macOS, which is where they are normally run. The Linux branch of `smoke-test.sh` was independently wrong *before* the rename too (XDG vs `~/.omnivoice`), so all three platforms now match `backend/core/config.py` rather than only the two that were noticed.

The identity guard missed these because it covered only the Rust resolvers and the Windows fixtures. It now pins the shell scripts too, asserting on the **assignment lines** specifically — these files legitimately mention the brand in banners and comments, and a file-wide substring check is exactly the too-weak assertion that let this through.

Two remaining rebrand leftovers were factually wrong rather than merely stale: the GGUF engine told macOS users to `xattr -cr /Applications/OmniVoice Studio.app`, which no longer exists, and `uninstall.ps1` named the wrong Add/Remove Programs entry.

## Verification

- New tests fail before, pass after: 4/5 in `CaptureWidgetStrandedPill`, both halves of `widgetWindowTransparency`, and the new shell-script guard (verified by re-introducing the regression).
- Backend: 4362 passed, 25 skipped.
- Frontend: 210 files, 1685 tests passed. `typecheck:ci`, `format:check` clean; `lint` exits 0 and this change **reduces** CaptureWidget warnings 9 → 8.

The PR prevents stranded dictation widget windows, stabilizes tray listeners, scopes transparency to the widget, restores `OmniVoice` data paths, and updates model sizes and threading. These changes address lost or disabled dictation events that could leave an opaque, empty window with no dismissal method. Human review should focus on widget visibility, cross-platform path assignments, and model thread-selection behavior.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -6,7 +6,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Thanks for helping improve OmniVoice Studio! 🎙️
+        Thanks for helping improve VoiceStudio! 🎙️
 
         **Fastest path to a fix:** **Settings → About → "Save diagnostic bundle"** makes a
         zip (self-check + recent errors + scrubbed log tails) — drag it onto this issue and
--- a/.github/ISSUE_TEMPLATE/sponsor.yml
+++ b/.github/ISSUE_TEMPLATE/sponsor.yml
@@ -6,7 +6,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Thanks for considering sponsoring **OmniVoice Studio** 💛
+        Thanks for considering sponsoring **VoiceStudio** 💛
 
         OmniVoice is free, local-first, and AGPL-3.0 — sponsorship keeps development going.
         See **[SPONSORS.md](https://github.com/debpalash/VoiceStudio/blob/main/SPONSORS.md)** for tiers, placements, and logo guidelines.
--- a/.github/ISSUE_TEMPLATE/sponsor.yml
+++ b/.github/ISSUE_TEMPLATE/sponsor.yml
@@ -1,5 +1,5 @@
 name: 🤝 Sponsorship inquiry
-description: Support OmniVoice and (optionally) claim a logo slot. Not for bugs or feature requests.
+description: Support VoiceStudio and (optionally) claim a logo slot. Not for bugs or feature requests.
 title: "Sponsorship inquiry: "
 labels: ["sponsor"]
 body:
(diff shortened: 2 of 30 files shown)


debpalash:
Reopening to trigger CI — the GitHub Actions outage swallowed this PR's workflow run.

Issue #1416, opened 2026-08-07
[Bug] TTS generate ran for more than 300s of actual compute time and was abandoned — t


sciencefiction879-cmyk:
## Error

```
TTS generate ran for more than 300s of actual compute time and was abandoned — the backend is running, but this job was too heavy for the available compute. The abandoned job cannot be killed: it keeps running and keeps holding the device until it finishes on its own, so an immediate retry competes with it. Wait for the current job to drain (or restart the backend) before retrying; most often the GPU is VRAM-starved (a resident model and this job contend for memory). For a durable fix, Flush caches / Unload the resident model (top toolbar or Settings → Models) before retrying, try shorter text, a lighter engine, or set the engine to CPU in Settings → Models. (Raise OMNIVOICE_GENERATE_TIMEOUT_S for very long single generations.)
d@tauri://localhost/assets/main-app-B6MZUbYV.js:16:28283
Si@tauri://localhost/assets/main-app-B6MZUbYV.js:17:55
```

## Environment

**Version:** `0.4.2`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.2-arm64-arm-64bit`
**Python:** `3.11.15`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M1 Pro`
**RAM:** `16 GB`
**Disk free:** `19.8 GB`
**Active TTS engine:** `mlx-audio`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 1 s before this report

## Recent actions

```
18:14:59 view:studio
18:16:06 generate:start (audio)
18:17:34 view:settings
18:18:56 view:studio
18:21:18 generate:stream-retryable-abort
18:23:22 generate:start (audio)
18:23:45 view:launchpad
18:23:56 view:settings
18:25:14 engine:tts=mlx-audio
18:25:20 view:launchpad
18:25:23 view:studio
18:25:46 view:dub
18:26:38 dub:ingest-url
18:27:06 view:launchpad
18:27:22 view:stories
18:27:24 view:launchpad
18:27:29 view:studio
18:28:22 generate:stream-retryable-abort
```

debpalash:
[...]

PR #1418, opened 2026-08-07
fix(tts): stop the cold model load racing the event loop (#1417)

### LINKED ISSUE #1417: [Bug] v0.4.2: TTS broken after VoiceStudio rename — cannot import name 'VoiceStudio' from omnivoice.models.omnivoice, /v1/audio/speech 500
## Describe the bug

After the VoiceStudio product rename (commit `5cab8e01`, 2026-08-06), all TTS paths are broken on installs that have the `omnivoice` TTS engine active:

- `POST /v1/audio/speech` → `500 {"detail": "<asyncio.locks.Lock object ...> is bound to a different event loop"}`
- `POST /generate` (native form endpoint) → hangs, then `Model load exceeded 1200.0s; resetting GPU pool.`
- Startup preload fails as "non-fatal" → the app runs with dead TTS and no visible error.

## Root cause

`backend/services/model_manager.py:32` (and the fallback at line 36) imports `VoiceStudio` from `omnivoice.models.omnivoice`, but the class in that module is still named `OmniVoice` (line 242) — the rename commit changed the import but never added a `VoiceStudio` alias. Confirmed present in both v0.4.2 and current `main`.

Once the ImportError is removed, a second latent bug surfaces on `/v1/audio/speech`: `tts_backend.py:564` calls `asyncio.run(get_model())` from a GPU-pool worker thread, while `get_model()` acquires a module-level `asyncio.Lock` (`_model_lock`, `model_manager.py:949`) that is bound to the Uvicorn event loop → `RuntimeError: ... is bound to a different event loop`.

## Repro

1. Launch v0.4.2 (preview channel), macOS arm64, MPS.
2. Log: `ImportError: cannot import name 'VoiceStudio' from 'omnivoice.models.omnivoice'` on every preload attempt.
3. `curl -X POST http://localhost:3900/v1/audio/speech -H "Content-Type: application/json" -d '{"model":"tts-1","voice":"alloy","input":"test"}'` → 500 (lock bound to a different event loop).
4. `/generate` → after retries, `Model load exceeded 1200.0s; resetting GPU pool.`

## Environment

- Version: 0.4.2 (preview channel, auto-updated from 0.4.0)
- OS: macOS 14 (arm64, Apple Silicon MPS)
- Python: 3.11.15 (uv-managed)
- Active TTS engine: `omnivoice` (in-process, MPS, routing: accelerated)
- RAM: 18 GB total, ~2.5 GB free at failure time
- Model cache: `k2-fsa/OmniVoice` present (3.0 GB); model had loaded successfully 230+ times before the rename commit

## Logs

```
ImportError: cannot import name 'VoiceStudio' from 'omnivoice.models.omnivoice'
  (/Users/.../project/omnivoice/models/omnivoice.py)
Traceback: model_manager.py:32 `from omnivoice.models.omnivoice import
  VoiceStudio as _OV` → _load_model_sync (line 1737) → preload_model (2165)
  — caught as non-fatal

POST /v1/audio/speech:
  tts_backend.py:562 asyncio.get_running_loop() → RuntimeError: no running
  event loop
  (during handling): tts_backend.py:564 asyncio.run(get_model()) →
  model_manager.py:2007 `async with _model_lock` →
  RuntimeError: <asyncio.locks.Lock object ... [locked, waiters:4]> is bound
  to a different event loop

/generate:
  Model load exceeded 1200.0s of EXECUTION time — abandoned the GPU-pool
  worker ... (repeated retries)
```

## Expected behavior

- The `VoiceStudio` symbol resolves (or the import is reverted to `OmniVoice`).
- `/v1/audio/speech` returns synthesized audio.
- The model lock is acquired within a single event loop.

## Suggested fix (for maintainers)

The rename must be completed properly, not patched:

1. In `omnivoice/models/omnivoice.py`, either rename the class to `VoiceStudio` (with a backwards-compatible `OmniVoice = VoiceStudio` alias) or keep `OmniVoice` and add a `VoiceStudio` alias.
2. In `tts_backend.py._ensure_loaded`, don't `asyncio.run(get_model())` from a pool thread while `_model_lock` is bound to the server loop — create the lock inside `get_model()`'s loop, or use a `threading.Lock` for load-once semantics.
3. Add tests: module import of `VoiceStudio`, a cold-start TTS round-trip through `/v1/audio/speech`, and the `/generate` path — so the rename can never ship broken again.
4. Verify the end-to-end use case: launch → preload → synthesize → transcribe.

## Quick runtime workaround (user-side, NOT a fix)

Verified working locally on v0.4.2 (macOS arm64/MPS): append `VoiceStudio = OmniVoice` at the end of `omnivoice/models/omnivoice.py`. This unblocks TTS end-to-end (`/v1/audio/speech` 200, cloned-voice synthesis OK) but is overwritten on every app update, which is why the proper rename fix above is needed.

debpalash:
Reported by @mmospanenko with a precise root cause for each of two defects. The first — `model_manager.py` importing a `VoiceStudio` class the library doesn't export — **landed separately in #1420**; this PR is the second defect, which the first was hiding.

## The cold load raced the event loop

`_model_lock` is a module-level `asyncio.Lock`, but `OmniVoiceBackend._ensure_loaded` bootstraps a **fresh** loop with `asyncio.run(get_model())` from a GPU-pool worker thread.

Depending on which loop touched the lock first, this either raises the reported

```
RuntimeError: <asyncio.locks.Lock …> is bound to a different event loop
```

(the `/v1/audio/speech` 500) **or deadlocks outright** — the pool thread binds the lock and then waits on a future the releasing loop cannot touch. The deadlock branch is a plausible contributor to the "ran for more than 300s and was abandoned" reports (#1416, #1419) and to #1329's "advances one sentence, then stops with no error": neither produces an error, both look like slow hardware.

`_heal_tts_placement` already carries a `running_on_gpu_pool()` guard with a comment describing this exact situation; the cold-load path in `get_model()` never got the same treatment.

The load also has to run **inline, in this thread**. Routing it through `_load_model_with_timeout()` would hand `_load_model_sync` back to `_get_gpu_pool()` — the pool we are currently occupying — and MPS pins that pool to a single worker, so it would wait on itself (CodeRabbit caught this in the first revision). Exclusion comes from a loop-agnostic `threading.Lock` rather than from holding a GPU slot, since a CUDA host's pool has more than one worker and holding one slot excludes nobody.

## Regression cover

- **The lock path, driven through real contention.** My first attempt passed with *and* without the fix: an uncontended `asyncio.Lock.acquire()` takes a fast path that never calls `_get_loop()`, so it never binds and never complains. The error only exists on the waiting path, so the test holds the lock on a live foreign loop — and now reproduces the reporter's exact string. Its worker is a daemon thread because the unfixed behaviour can hang; a non-daemon one would take the suite down at interpreter exit instead of reporting a failure. Both joins assert the worker actually finished before any shared module state is reset, so a timed-out regression cannot leak a live thread into the next test.
- **The pool is never re-entered.** A stand-in pool that refuses `submit()` turns the self-wait into an instant, readable failure instead of a hang.
- **Off-pool callers still take the lock** — that exclusion is what stops two cold loads racing into memory, and it would be easy to disarm by accident.
- **Import integrity** (from the original #1417 work, still worth keeping): an AST scan, no torch import, 0.7 s, pinning every `from omnivoice… import X` in the backend against what that module actually defines. It catches the whole class of "a sweep renamed one side of an import" rather than the one symbol #1420 fixed. Includes a guard-the-guard test, since a scan that silently matched nothing would pass while protecting nothing.

Backend suite green.

Closes #1417

Restores `OmniVoice` loading and uses a process-wide `threading.Lock` with inline cold loading on GPU-pool workers. This fixes TTS startup and first-request failures after the VoiceStudio rename without cross-event-loop lock errors or GPU-pool self-deadlocks. Review GPU-pool worker detection because incorrect classification could bypass the model lock.


debpalash:
Both findings taken — the Critical one was right, and it mattered.

`_load_model_with_timeout()` dispatches `_load_model_sync` back into `_get_gpu_pool()`, so calling it from a pool worker re-queues behind the slot the caller already holds. MPS pins that pool to a single worker, so it would have waited on itself: **the same deadlock this PR set out to remove, on the same platform, just reached by a different route.** Worse than the original in one respect — it would have looked fixed.

The test was complicit. Faking `_load_model_with_timeout` meant the real dispatch never ran, so it proved the guard was *taken*, not that the load could *finish*. Exactly as you said: exercise the real dispatch path.

Changes:

- The pool path loads **inline** via `_load_model_sync()`, keeping `_make_room_before_tts_load()` so memory-tight machines still get the reclaim.
- Exclusion moves to a module-level `threading.Lock`. My original rationale — "holding the GPU slot is already the mutual exclusion" — **was wrong**, and I've corrected it in the code comment: a slot is not exclusion when the pool has more than one worker, which CUDA hosts do. A loop-agnostic lock holds from any thread or loop.
- The regression now patches the **leaf** loader and installs a pool whose `submit` raises. Re-introducing the re-submit fails in 0.7s with a readable message rather than hanging — deliberate, because a wedged pool thread blocks interpreter exit and would take the suite down instead of reporting a failure. Verified by putting the bug back.
- Minor finding taken too: the lock-holder polled `asyncio.sleep(0.01)`; it now waits on the event in an executor, per the no-sleeps-as-synchronization rule.

Backend suite: **4227 passed, 32 skipped**.

Issue #1006, opened 2026-07-08
[Bug] Transcribe stream dropped before emitting any segments. Likely ASR backend faile


Labpar000:
## Error

```
Transcribe stream dropped before emitting any segments. Likely ASR backend failed to load — check backend log + Settings → Models.
Error: Transcribe stream dropped before emitting any segments. Likely ASR backend failed to load — check backend log + Settings → Models.
    at EventSource.<anonymous> (http://tauri.localhost/assets/main-app-BTIqFh6_.js:16:22599)
```

## Environment

**Version:** `0.3.11`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36 Edg/149.0.0.0`

## Recent actions

```
16:22:16 view:donate
03:43:37 view:settings
03:43:39 view:launchpad
03:43:50 view:studio
03:44:00 generate:start (audio)
03:44:43 generate:start (audio)
03:45:08 view:dub
03:46:14 dub:upload
03:46:40 view:settings
```

debpalash:
Thanks for filing. I dug into what that specific message means: it's a fallback the frontend shows only when the transcribe SSE connection drops *without* the backend ever getting a chance to report why (no structured error, no `done` event) — i.e. the backend was interrupted or crashed mid-stream, rather than failing normally (normal ASR-load failures — missing weights, wrong compute type, etc. — already surface with a specific, actionable message, which is not what you got).

Two most likely causes, and I can't tell which without more data:
1. A backend crash/restart that happened to land mid-transcribe (unrelated to the transcribe itself).
2. A VRAM-related issue — your action log shows two standalone audio generations right before the dub upload, so if a model was still resident when ASR tried to load, that's worth ruling out.

Could you check **Settings → Logs → Backend** (or the `backend.log` file) for whatever's around the timestamp of the dub upload, and paste it here? That'll show whether the process actually crashed (and if so, on what) versus something else. Also worth trying: does the same dub upload fail again on a fresh app restart with nothing else running?

debpalash:
One more lead while waiting on your backend log: your report was on v0.3.11, and **v0.3.13** (just released: https://github.com/debpalash/OmniVoice-Studio/releases/latest) fixes a backend shutdown race that produced exactly this kind of mid-operation connection drop for several other reporters — quitting or restarting while a model was still loading could poison the process without a proper error.

[...]

PR #1447, opened 2026-08-09
fix(security): constrain endpoint probes to trusted origins

debpalash:
## Summary
- allow endpoint-race network probes only to VoiceStudio's two shipped HTTPS origins
- reject alternate schemes, lookalike hosts, credentials, non-default ports, paths, queries, and fragments before `urlopen`
- remove side effects and unclosed reads from assertions in two backend tests

Addresses Bandit code-scanning alerts #936 and #937, CodeQL #988–#990, and #1006.

## Validation
`PYTHONPATH=$PWD:$PWD/backend pytest -q tests/test_endpoint_race.py backend/tests/test_batch.py backend/tests/test_run_sentinel.py` — 75 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ The bundled TTS model package (`pyproject.toml`) is versioned independently.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


debpalash:
@coderabbitai review

PR #1497, opened 2026-08-11
WIP: recover cleanly from unreachable remote backends

### LINKED ISSUE #1496: [Remote backend] Unreachable/TLS endpoints trap the app in setup
## Finding

When a user saves a remote endpoint such as `https://192.168.0.110:7443`, the desktop app repeatedly requests `/model/status` and `/setup/status` from that address. A self-signed or otherwise untrusted certificate is rejected by the WebView before the response can be read. The current UI then appears stuck and may return to the local model-install/setup wizard.

## Reproduction

1. Run the desktop app.
2. In Settings → Sharing → Remote backend, save `https://192.168.0.110:7443` (or another unreachable/self-signed endpoint).
3. Reload.
4. Observe repeated `Unacceptable TLS certificate` / CORS-style fetch errors and setup/install UI instead of a recoverable connection state.

## Root cause

- Port `7443` is the remote-worker gRPC/TLS control-plane port in PR #1495, not the HTTP VoiceStudio backend API port normally used by the Remote backend setting.
- `RemoteBackendPanel` validates only URL syntax and can save an unverified endpoint after confirmation.
- `App.jsx` uses `/setup/status` as the local first-run gate. A configured remote backend is not distinguished from a local backend, so an unavailable remote target can block startup and incorrectly route the user to `SetupWizard`.
- There is no startup recovery action that clears `ov_backend_url` and returns to the local backend.

## Expected behavior

- Explain the difference between the HTTP backend URL and the remote-worker control-plane endpoint; do not suggest bypassing TLS verification.
- Treat a configured remote backend as remote: never show the local model-install wizard because the remote machine owns its models.
- On transport/TLS/certificate failure, show a clear non-blocking connection state with `Use local backend` / `Disable remote backend`, `Retry`, and a link to the Remote backend settings.
- Disabling the remote backend must clear `ov_backend_url` and the API key if appropriate, then reload/reconnect to the local backend without requiring a reinstall.
- Connection probing must classify certificate, network, CORS, HTTP, and wrong-service/port failures with actionable copy.
- Add regression coverage for the setup gate and recovery action; include all 21 locale files for new user-facing strings.

## Verification evidence

During local PR #1495 testing:

- PR frontend ran on `http://localhost:3911` and PR backend on `http://127.0.0.1:3910`.
- With `OMNIVOICE_UI_PORT=3911`, local CORS was corrected and `/model/status` returned `Access-Control-Allow-Origin: http://localhost:3911`.
- The failing remote target remained `https://192.168.0.110:7443`, where the WebView reported `Unacceptable TLS certificate`.

Related: #1495

debpalash:
## Handoff

Tracks the remote-backend recovery work found while validating PR #1495. The full finding and reproduction are in #1496.

This draft is intentionally a small handoff PR for another contributor to carry. It records the observed WebView TLS failure, the distinction between the HTTP backend API and the 7443 remote-worker control plane, and the acceptance contract.

## Required follow-up

- Classify certificate, transport, CORS, wrong-port/service, and HTTP failures.
- Never route a configured remote backend into the local model-install wizard.
- Add a recoverable screen with Retry, Disable remote backend / Use local backend, and Settings access.
- Clear `ov_backend_url` and the API key when disabling, then reload into local mode.
- Add frontend regressions and translations across all 21 locales.

Closes #1496 when the implementation and regression coverage land.\n\n

### DIFF
--- a/docs/handoffs/remote-backend-recovery-1496.md
+++ b/docs/handoffs/remote-backend-recovery-1496.md
@@ -1,0 +1,44 @@
+# Remote backend recovery handoff (#1496)
+
+This draft PR is a handoff for the remote-backend recovery work described in
+[#1496](https://github.com/debpalash/VoiceStudio/issues/1496). It is based on
+`main`; PR #1495 adds the remote-worker control plane involved in the report.
+
+## Observed failure
+
+Saving `https://192.168.0.110:7443` as the Remote backend makes the desktop
+WebView request `/model/status` and `/setup/status` from that URL. The WebView
+rejects the self-signed certificate, and the app can fall back into the local
+model setup path even though the configured backend is remote.
+
+## Important distinction
+
+- `7443` is the remote-worker gRPC/TLS control-plane port.
+- The Remote backend setting expects the HTTP VoiceStudio API, normally on
+  `3900` (or another explicitly configured HTTP API port).
+- Certificate verification must remain enabled. The fix should explain a
+  certificate/transport failure rather than weakening TLS validation.
+
+## Intended implementation
+
+1. Add a bounded remote-backend probe that classifies certificate, network,
+   CORS, wrong-port/service, and HTTP failures.
+2. If `ov_backend_url` is configured, never route `/setup/status` failure into
+   the local `SetupWizard`; remote model installation belongs to the remote
+   host.
+3. Add a recoverable connection screen with `Retry`, `Use local backend` /
+   `Disable remote backend`, and a Settings link.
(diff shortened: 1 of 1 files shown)


debpalash:
Superseded by #1501, which implemented this handoff completely: classified bounded remote health probing, retry/settings/use-local recovery, clearing the remote URL and API key, all 21 locale strings, and focused regressions. #1501 passed full backend/frontend, Security, and Linux/macOS/Windows smoke + Tauri gates and is already on main. Closing this draft avoids reintroducing its old-base conflicts and weaker pre-review probe behavior.

Issue #1185, opened 2026-07-18
[Bug] ASR backend initialization failed: No module named 'lightning_fabric'


gitmeatarru:
## Error

```
ASR backend initialization failed: No module named 'lightning_fabric'
Error: ASR backend initialization failed: No module named 'lightning_fabric'
    at EventSource.<anonymous> (http://tauri.localhost/assets/main-app-BajjIO_S.js:17:7555)
```

## Environment

**Version:** `0.3.21`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36 Edg/147.0.0.0`
**OS:** `Windows-10-10.0.26100-SP0`
**Python:** `3.11.14`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 3060 Laptop GPU (6 GB VRAM)`
**CPU:** `Intel64 Family 6 Model 154 Stepping 3, GenuineIntel`
**RAM:** `63.7 GB`
**Disk free:** `49.1 GB`
**Active TTS engine:** `omnivoice`

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-07-18T08:43:25.000Z (5 min ago)
**Exit:** `exit code -1073741510`
**Uptime before crash:** 2304 s
**Backend version:** `0.3.21`

```
… (truncated)
 (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
2026-07-18 11:43:07,668 ERROR [asyncio] Exception in callback _ProactorBasePipeTransport._call_connection_lost(None)
handle: <Handle _ProactorBasePipeTransport._call_connection_lost(None)>
Traceback (most recent call last):
  File "~\AppData\Roaming\uv\python\cpython-3.11-windows-x86_64-none\Lib\asyncio\events.py", line 84, in _run
    self._context.run(self._callback, *self._args)
  File "~\AppData\Roaming\uv\python\cpython-3.11-windows-x86_64-none\Lib\asyncio\proactor_events.py", line 165, in _call_connection_lost
    self._sock.shutdown(socket.SHUT_RDWR)
2026-07-18 11:43:15,312 ERROR [asyncio] Exception in callback _ProactorBasePipeTransport._call_connection_lost(None)
handle: <Handle _ProactorBasePipeTransport._call_connection_lost(None)>
Traceback (most recent call last):
[...]

Issue #1032, opened 2026-07-09
[Bug] TTS processing is much slower than v0.3.5, missing Clear History, and no way to stop audio preview


hokimin535-dev:
### Before filing

- [x] I searched [existing issues](https://github.com/debpalash/OmniVoice-Studio/issues?q=is%3Aissue) and this isn't a duplicate.
- [x] I'm on the latest release (or `main`) — older builds may already be fixed.

### What happened?

First of all, thank you for the continuous improvements to OmniVoice Studio.

I've noticed a few issues in the recent versions compared to OmniVoice Studio v0.3.5.

1. Text-to-Speech processing has become significantly slower.
Using exactly the same model, settings, hardware, and input text, OmniVoice Studio v0.3.5 generates speech much faster than the latest releases. The slowdown is very noticeable during daily use.

2. The "Clear History" button appears to have been removed.
I can no longer find an option to quickly clear previous TTS jobs. This feature was very useful when generating many audio clips.

3. There is no obvious way to stop or clear the audio preview after a generation finishes.
After TTS completes, the generated audio remains available in the preview, and I cannot find a Stop/Clear option. My current workaround is to completely close and reopen OmniVoice Studio, which is inconvenient.

These issues make the newer versions less efficient to use than v0.3.5.

### Steps to reproduce

1. Open OmniVoice Studio (latest version).
2. Paste the same text.
3. Click Generate.
4. Compare the generation speed with OmniVoice Studio v0.3.5.
5. Try to find the "Clear History" option.
6. Generate audio and try to stop or clear the preview without restarting the application.

### What did you expect instead?

I expected the latest version to provide performance at least comparable to v0.3.5 when generating speech from the same text using the same TTS engine and hardware.

I also expected:

- A Clear History button to remove previous TTS jobs.
- A Stop or Clear option for the audio preview after generation.

### Operating system

Windows (x64)

### How did you install it?

Desktop app (installer / AppImage)

### Version

v0.3.14

### Compute device

NVIDIA GPU (CUDA)

### Logs / diagnostic bundle

```text

```

debpalash:
Thank you — a three-part report with a clean baseline comparison ("same model, settings, hardware, text — v0.3.5 was faster") is exactly what makes a perf regression findable. Investigating all three now:

[...]

PR #1519, opened 2026-08-13
fix(asr): degrade past a broken engine on every transcription path (#1512)

### LINKED ISSUE #1512: Transcription 500s instead of degrading when the auto-picked ASR engine fails to load
## Summary

Transcription request paths select the ASR engine with `get_active_asr_backend()` — the **pure selector** — instead of `load_active_asr_backend()`, the loader added in #1185 that also runs `ensure_loaded()` and degrades past an engine whose deep import chain is broken. An engine that passes the shallow `is_available()` probe but dies on first real use therefore reaches `.transcribe()` and 500s the request, even with a healthy engine next in line.

## Repro

Any host where `import whisperx` succeeds but its ctranslate2 dependency cannot load. Observed on Linux 7.0 (hardened kernel refusing an exec-stack `.so`):

```
ImportError: libctranslate2-d3638643.so.4.4.0: cannot enable executable stack
             as shared object requires: Invalid argument
```

Auto-detect picks `whisperx` (probe passes), then:

```
POST /v1/audio/transcriptions   → 500, raw dlopen error returned to the client
POST /transcribe  mode=accurate → 500 {"error_class":"NETWORK_ERROR"}
POST /transcribe  (fast)        → 200 — routes via get_capture_asr_backend()
```

`GET /system/asr-backends` reports `pytorch-whisper` and `sherpa-onnx-asr` as available the whole time.

Confirming the loader already handles it correctly:

```
active_backend_id: whisperx
get_active  -> whisperx          # what the failing routes called
load_active -> pytorch-whisper   # degraded OK
```

## Affected call sites

| File | Path |
|---|---|
| `backend/api/routers/openai_compat.py:549` | `POST /v1/audio/transcriptions` |
| `backend/api/routers/capture.py:101` | `POST /transcribe` (accurate mode) |
| `backend/api/routers/dub_export.py:1438` | dub QC / verify pass |
| `backend/api/routers/batch.py:161` | batch transcribe stage |
| `backend/api/routers/dub_core.py:1691` | dub main transcribe |
| `backend/services/asr_backend.py:2691` | `transcribe_reference` (voice-clone reference text) |

## Fix

Swap each to `load_active_asr_backend()` (it accepts the same `asr_pipe=` kwarg), map `ASRModelMissingError` to the typed 409 already used by the preflight, and move the load inside the thread pool so it never blocks the event loop. Plus a CI guard so a seventh call site can't reintroduce it.

ahov520:
## Summary

- Six transcription paths (OpenAI-compat `POST /v1/audio/transcriptions`, dictation accurate mode, dub main transcribe, dub QC, the batch transcribe stage, and voice-clone reference transcripts) selected the ASR engine with `get_active_asr_backend()` — the pure selector — so an engine that passes the shallow `is_available()` probe but dies on first real use (broken deep import chain, the #1185 class) reached `.transcribe()` and 500'd the request even with a healthy engine next in line.
- All six call sites now go through `load_active_asr_backend()` (same `asr_pipe=` kwarg), with the possibly minutes-long load running inside the thread pool so it never blocks the event loop, and `ASRModelMissingError` from a weightless fallback mapped to the same typed 409 + download CTA the preflights already use.
- `transcribe_reference` keeps its construct-only fast path, the pytorch-whisper deferral, and the #1032 cache semantics; it falls into the degrading loader only when `ensure_loaded()` reports a broken import chain.
- A new CI guard (`backend/tests/test_asr_selector_stays_private.py`) keeps a seventh call site from reintroducing the pure selector outside `services/asr_backend.py`, and a fail-before/pass-after test covers the reference-transcript degrade.

Fixes #1512

## Test plan

- [x] `pytest backend/tests/` — 221 passed, 2 skipped
- [x] Affected suites: `tests/test_asr_model_missing.py`, `tests/test_transcribe_reference.py`, `tests/test_dub_transcribe.py`, `tests/test_speaker_hint.py`, `tests/test_funasr_global_speaker_chunking.py`, `tests/test_capture_refine.py`, `tests/test_engines.py`, `tests/test_tts_placement_self_heal.py`, `tests/test_asr_openai_compat_877.py` — all green
- [ ] CI full suite (backend + frontend)

All ASR transcription paths now use `load_active_asr_backend()` with thread-pool loading and structured 409 responses for missing models. This prevents 500 errors when an available engine fails during loading and adds regression coverage for fallback behavior and selector privacy. Review the fallback and error-handling paths for consistent behavior across transcription endpoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ the frozen-backend fallback mirror it for their toolchains.
 ### Fixed
 
 - The Linux app icon is no longer blank: the AppImage shipped `.DirIcon` as a symlink into the machine that built it, so file managers and app menus drew nothing. (#1518)
+- Transcription, dubbing, batch and QC now degrade to the next working speech-recognition engine when the auto-picked one fails to load, instead of failing the request while a healthy engine is installed. (#1512)
 - The Linux desktop entry no longer ships an empty `Categories=`, which `desktop-file-validate` rejects and menu builders skip. (#1518)
 - The Simplified Chinese (zh-CN) translation no longer mistranslates brand names and technical terms — Discord, Tailscale, Hugging Face, IPA, and LLM (Cinematic) were rendered as nonsensical literal translations, and ~250 more awkward machine-translation strings are now natural Chinese. (#1508) — thanks @anyingiit!
 - Worker restart coverage now waits for the registration response to persist its identity instead of racing the client callback in CI. (#1505)
--- a/backend/api/routers/batch.py
+++ b/backend/api/routers/batch.py
@@ -151,14 +151,20 @@ async def _run_batch_pipeline(job_id: str, job: dict):
     # ── 2. Transcribe ─────────────────────────────────────────────────
     _set_progress(job, "transcribe", 0)
 
-    from services.asr_backend import get_active_asr_backend
+    from services.asr_backend import load_active_asr_backend
     from services.model_manager import _gpu_pool, _cpu_pool, run_on_gpu_pool_guarded
     from services.segmentation import (
         segment_transcript, assign_speakers_heuristic,
     )
 
     def _transcribe():
-        backend = get_active_asr_backend()
+        # Loading selector (#1512): degrades past an engine whose deep import
+        # chain is broken instead of failing the job while a healthy engine
+        # works. An ASRModelMissingError from a weightless fallback propagates
+        # to _worker's structured job-failure handling, same as any other
+        # transcribe-stage error (the submit route already preflights the
+        # initial selection).
+        backend = load_active_asr_backend()
         result = backend.transcribe(audio_path, word_timestamps=True)
         detected_lang = result.get("language", "en")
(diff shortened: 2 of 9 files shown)


debpalash:
Thank you for this — and sorry for the collision. The same fix landed a few minutes before this PR opened, in #1515, which is why it now shows as conflicting.

What's on `main` covers the same six call sites (`openai_compat`, dictation accurate mode, dub main + QC, batch, voice-clone reference) via `load_active_asr_backend()`, plus `backend/tests/test_asr_request_path_degrades.py` as a recurrence guard.

**Your version did one thing better**, and I've absorbed it with credit in #1523: your guard scanned the whole backend with `rglob`, where mine scanned `api/routers/` only. That difference is real — a service that transcribes on a request's behalf skips `ensure_loaded()` just as thoroughly, and I confirmed a `get_active_asr_backend()` call in `services/tts_backend.py` passes the router-only scan and fails yours.

Closing as superseded, not as unwanted. The changelog credits you.

PR #1520, opened 2026-08-13
fix(api): answer 422 for malformed requests instead of 500ing and logging the body (#1513)

### LINKED ISSUE #1513: Binary body to a JSON endpoint returns 500 and dumps the whole request into omnivoice.log
## Summary

`backend/main.py` registers an `Exception` handler but none for `RequestValidationError`, so FastAPI's default handler runs. That handler calls `jsonable_encoder(exc.errors())`, and for a body-level failure `errors()[i]["input"]` holds the **raw request body**. `jsonable_encoder` decodes `bytes` as UTF-8, so any binary body raises `UnicodeDecodeError` *inside the error handler* — turning a malformed request into a 500.

Two consequences:

1. **Wrong status.** A malformed request returns 500 instead of 422.
2. **The whole request body is written to `omnivoice.log`.** The escaping exception carries it into the traceback. A 145 KB WAV wrote ~500 KB of log — user audio landing on disk, in the file we invite people to paste into bug reports.

## Repro

Post a binary multipart body to any JSON-body route. A 4-byte file is enough:

```bash
printf '\x80\x81\x82\xff' > bin.dat
for ep in /design/describe /tools/probe /dub/translate; do
  curl -s -o /dev/null -w "$ep -> %{http_code}\n" -X POST localhost:3900$ep -F "f=@bin.dat"
done
```

Before:

```
/design/describe -> 500
/tools/probe     -> 500
/dub/translate   -> 500
```

Backend log:

```
File ".../fastapi/encoders.py", line 318, in jsonable_encoder
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 154: invalid start byte
```

A *text* multipart body to the same routes correctly returns 422 — it's specifically the non-UTF-8 path that breaks. This is easy to hit accidentally: an MCP or OpenAI-compat client posting an audio upload to the wrong path gets a 500 and a log full of the clip.

## Fix

Register a `RequestValidationError` handler that sanitizes each error's `input` before encoding — report binary as `<N bytes of binary data>` rather than decoding it, truncate long strings, and stringify `ctx` — then return 422 through `_cors_headers_for(request)` like every other hand-built error response in that module. Keeping the exception from escaping also closes the log-dump.

ahov520:
## Summary

- `backend/main.py` registered an `Exception` handler but none for `RequestValidationError`, so FastAPI's default handler ran `jsonable_encoder(exc.errors())`. For a body-level failure `errors()[i]["input"]` holds the raw request body, and `jsonable_encoder` decodes bytes as strict UTF-8 — any binary body (e.g. an audio upload posted to a JSON route by an MCP / OpenAI-compat client) raised `UnicodeDecodeError` *inside the error handler*: the client got a 500 instead of a 422, and the escaping traceback carried the whole request body into `omnivoice.log`.
- A new `RequestValidationError` handler sanitizes every error's `input` before encoding — binary reported by size only, long strings truncated, containers walked shallowly, `ctx` values (including the exception objects pydantic parks there) stringified with a bound — and answers 422 through `_cors_headers_for(request)` like every other hand-built error response in the module.
- Tests cover the 422 contract, the "body never reaches the crash log / error journal" guarantee, CORS headers on the handled response, and the sanitizer itself.

Fixes #1513

## Test plan

- [x] `pytest backend/tests/` — 228 passed, 2 skipped (includes the 9 new tests driving main's real handlers via TestClient)
- [x] `tests/test_changelog_style.py`
- [ ] CI full suite (backend + frontend)

Added a FastAPI `RequestValidationError` handler that sanitizes invalid inputs and returns 422 responses with CORS headers. This prevents malformed binary requests from causing `UnicodeDecodeError`, 500 responses, or raw body logging. Review the bounded sanitizer behavior for nested and non-JSONable values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
+- Posting a binary body to a JSON endpoint now returns a proper 422 validation error instead of a 500, and the request body no longer lands in `omnivoice.log`. (#1513)
 - The Linux app icon is no longer blank: the AppImage shipped `.DirIcon` as a symlink into the machine that built it, so file managers and app menus drew nothing. (#1518)
 - The Linux desktop entry no longer ships an empty `Categories=`, which `desktop-file-validate` rejects and menu builders skip. (#1518)
 - The Simplified Chinese (zh-CN) translation no longer mistranslates brand names and technical terms — Discord, Tailscale, Hugging Face, IPA, and LLM (Cinematic) were rendered as nonsensical literal translations, and ~250 more awkward machine-translation strings are now natural Chinese. (#1508) — thanks @anyingiit!
--- a/backend/main.py
+++ b/backend/main.py
@@ -346,6 +346,7 @@ import time
 import threading
 from contextlib import asynccontextmanager
 from fastapi import FastAPI, Request
+from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse, RedirectResponse, Response
 from fastapi.staticfiles import StaticFiles
 from fastapi.middleware.cors import CORSMiddleware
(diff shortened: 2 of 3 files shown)


debpalash:
Thank you — same collision as #1519: this fix landed in #1515 a few minutes before the PR opened, which is why it conflicts now.

`main` has the `RequestValidationError` handler in `backend/main.py` returning 422 and keeping the raw body out of `omnivoice.log`, with `backend/tests/test_validation_error_handler.py` covering it. I compared your version against it and didn't find a case it handles that main misses — if you spot one, please say so and I'll take it.

Closing as superseded, not as unwanted.

PR #1537, opened 2026-08-13
fix(worker): reserve the capacity slot before announcing the accept

### LINKED ISSUE #1536: [Flaky test] test_worker_at_capacity_rejects_without_penalty: second task reaches RUNNING on a capacity-1 worker
CI run https://github.com/debpalash/VoiceStudio/actions/runs/31728310305 (PR #1534, a locale-JSON-only diff) failed:

```
FAILED tests/test_worker_transport.py::test_worker_at_capacity_rejects_without_penalty
AssertionError: assert <TaskState.RUNNING: 'running'> in (QUEUED, ASSIGNED, ACCEPTED)
```

Passes 3/3 locally on the same tree, so it is a timing flake. Reading the test: `harness.client.config.max_concurrent_tasks = 1` is set **after** `connect_worker` and the first dispatch — on a slow runner the second assignment's worker-side accept decision can apparently race the first task being counted against capacity (or the config change propagating), letting the second task be accepted → RUNNING, which the assertion (correctly, per the docstring's 'worker accept is authoritative') treats as a violation.

Worth root-causing in the worker transport rather than papering over in the test: if a capacity-1 worker can genuinely accept a second task in a dispatch race, the production invariant the docstring claims ('the worker's own accept/reject is authoritative') has a real hole; if it cannot, the test's setup order (config set after connect) is the artifact and should move before `connect_worker`.

Seen once tonight (2026-08-13); CI rerun passed-pending. Filing so the class is tracked instead of eaten by reruns.

debpalash:
Closes #1536.

Root-cause analysis of the CI flake (`second.state == RUNNING` on a capacity-1 worker):

- The naive read — an interleaving between the capacity check and the slot insertion — turns out to be **impossible today**: the stream read loop handles messages serially and `_send` enqueues via `put_nowait` without ever suspending. 
- The only path to the observed state is the **first attempt dying environmentally** on a loaded runner (a stream hiccup fails `_run`, whose `finally` frees the slot), after which the worker accepting the second task is *correct* — the test's flat assertion punished legitimate behaviour.

Changes:
1. **Client hardening:** `_running` is reserved *before* the accept-send await — reserve-then-announce is the order that stays correct if `_send` ever gains backpressure (bounded outbox), instead of silently reopening an over-accept window. A failed accept send cancels the reserved task so work the scheduler never saw accepted cannot double-execute. Stream ordering is unaffected (`_send` enqueues synchronously, so ACCEPTED still precedes anything `_run` emits).
2. **Test pins the real invariant:** no over-concurrency — the assertion applies only while the first attempt is still RUNNING, and names the over-accept explicitly when it fires.

`tests/test_worker_transport.py` 43/43, the previously flaky test 5/5 in repetition.

The worker now reserves a running-task slot before sending acceptance and releases it if the send fails or is cancelled. This preserves the capacity invariant during concurrent assignments. The capacity-1 test now uses bounded polling and checks over-concurrency only while the first task remains running.

### DIFF
--- a/backend/worker/transport/client.py
+++ b/backend/worker/transport/client.py
@@ -667,9 +667,11 @@ class WorkerClient:
         self._running[key] = asyncio.create_task(self._run(assignment))
         try:
             await self._send(pb.WorkerMessage(accepted=pb.TaskAccepted(ref=assignment.ref)))
-        except Exception:
-            # The stream is dying; don't run work the scheduler never saw
-            # accepted — it would double-execute after reassignment.
+        except BaseException:
+            # BaseException, not Exception: a handler CANCELLED mid-send must
+            # release the slot too, or the reserved task keeps running work
+            # the scheduler never saw accepted — and double-executes after
+            # reassignment. The stream-death case lands here as well.
             task = self._running.pop(key, None)
             if task is not None:
                 task.cancel()
--- a/backend/worker/transport/client.py
+++ b/backend/worker/transport/client.py
@@ -657,8 +657,23 @@ class WorkerClient:
                 )
             )
             return
-        await self._send(pb.WorkerMessage(accepted=pb.TaskAccepted(ref=assignment.ref)))
+        # Reserve the slot BEFORE the accept-send await: awaiting yields to
+        # the event loop, and a concurrently delivered assignment would read
+        # the un-reserved counter and over-accept past capacity (#1536 — a
+        # capacity-1 worker accepted a second task on a slow runner). Message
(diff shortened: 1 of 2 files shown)


debpalash:
Review triage: the CancelledError leak (Major) is fixed — except BaseException with a fail-before/pass-after test; the fixed sleep is now a bounded state poll; the excluded_workers assertion moved outside the running-state guard. The TaskAccepted.envelope finding is skipped as pre-existing design: the server reads an envelope only at registration (server.py:443), nowhere on TaskAccepted — populating it would be dead wire bytes until a consumer exists.

PR #1548, opened 2026-08-14
fix(shell): gate Ready on the deep health probe; pace crash-loop restarts

debpalash:
## Why

From the backend-reliability root-cause pass (the "can't reach the local backend" class, 103 lifetime issues): two shipped-code gaps — the routine health poll is a shallow 500ms string-sniff of `/system/info` while the stronger `backend_deep_healthy` probe existed but wasn't the poll, and the supervisor respawns crashed backends back-to-back with zero pacing.

## What

- **`backend_ready()`** = identity probe AND deep probe (`/profiles` must return 200). Used at both Ready transitions: the 300s startup poll and the supervisor's 120s respawn wait. A zombie that answers `/system/info` from memory but 500s every DB-backed route can no longer be announced Ready. Death detection is untouched (process-exit-only via `try_wait`) — a busy-but-alive backend is still never killed.
- **Restart backoff** (`restart_backoff_delay`, pure): 0s → 5s → 15s keyed on respawns already inside the 3-in-600s window. First respawn stays immediate; the escalation stops a tight crash loop from burning the whole budget in seconds (back-to-back torch-import storms are exactly what re-OOMs a memory-pressured machine). Chunked sleep honors app-quit and deliberate retry-flow kills within 500ms, behind the already-visible reconnecting banner.

## Tests

- `ready_requires_the_deep_probe_not_just_identity` — loopback stub answering `/system/info` genuinely but `/profiles` with 500: shallow passes, ready fails; 200-both passes; nothing-listening fails.
- `restart_backoff_escalates_but_first_respawn_is_immediate` — immediate first respawn, escalation, cap, monotonicity.
- Full `cargo test`: 146 passed.

The supervisor now requires both `/system/info` and successful `/profiles` checks before declaring the backend ready, and it applies escalating restart delays after repeated crashes. Backoff waits remain interruptible within 500 ms for shutdown, deliberate retry cancellation, and replacement-child tracking. Review the deep probe and restart timing for false unhealthy states or delayed recovery.


debpalash:
Bot-review triage: fixed Greptile's P1 — the backoff now also yields the moment a live replacement child is tracked (the durable signal), not only on the deliberate-kill flag, which a completed Retry can set-and-clear between two 500ms samples. The Highlights-ref finding is noise: Highlights bullets don't carry refs per the deterministic changelog linter; subsection entries do.

debpalash:
Second-pass triage: replaced the liveness check with a spawn-generation counter — `track_backend_child` bumps it and it never un-bumps, so a replacement that has already exited by the next sample is still observed as an ownership transfer. Snapshot at death detection, yield on change.

debpalash:
Third-pass triage: snapshot now taken BEFORE observing the exit (any tracking after it — including one whose child we're about to see die — reads as a transfer), plus a final generation check before the port reclaim so the zero-backoff first respawn is covered.

PR #1547, opened 2026-08-14
fix(report): offer the latest release before filing from an outdated build

debpalash:
## Why

Root-causing the project's #1 lifetime failure class ("Can't reach the local backend", 103 issues) showed **61% of sampled reports were filed from builds that were already obsolete** and got closed with "please update" — triage noise that buries the real signal.

## What

- **`openBugReport()`** — the single gate every Report-bug affordance now goes through (ReportBugButton, ErrorBoundary, error toasts, BackendCrashNotice, BackendStartFailureNotice). On an outdated build it offers the latest release first, with a **File anyway** escape hatch (native dialog labels on desktop).
- **`**Build status:**` line** in every report body — `OUTDATED — vY was already out…` or `current at filing time` — so triage can grep current-version recurrence (the reliability metric) separately from stale-build reports.
- Freshness sources (identical behavior, per-mode implementation):
  - **Desktop:** the Rust updater's channel-aware verdict from the store — no new network path, no CSP widening. `idle` stays silent (can't distinguish up-to-date from not-yet-checked).
  - **Browser/dev/Docker:** one bounded (2.5s) latest-release GET, made only inside the user-initiated report flow whose destination is github.com — local-first preserved.
  - An `unknown` dev build is never nudged and never claims to be current.
- Preview builds (`X.Y.(Z+1)-N`) compare ahead of the stable tag by the versioning convention — never flagged.
- `askConfirm` gains optional `okLabel`/`cancelLabel` (backward compatible).
- 4 new i18n keys across all 21 locales.

## Tests

- `staleBuild.test.js` — version-triple parsing + outdated matrix (numeric compare, preview stamps, unparseable pairs).
- `bugReportGate.test.js` — browser + desktop gate paths: take-the-update, file-anyway (body carries OUTDATED marker), current, GitHub-unreachable, unknown-build silence, session caching, desktop no-network.
- Existing bugReport/component suites updated to the new entry point; locale parity + changelog style green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Centralize bug-report handling in `openBugReport()` and detect outdated builds before filing, while adding build-status details to submitted reports. Offer the latest release first for outdated builds and retain a “File anyway” option across desktop, browser, development, and Docker builds. Review the bounded release lookup, session caching, strict version parsing, and channel-aware desktop updater status for failure or stale-result risks.

PR #1549, opened 2026-08-14
feat(triage): crash-class recurrence report — the reliability metric

debpalash:
## Why

The reliability cycle's definition of done: the "backend died / never came up" report rate, **filtered to the current version** — 61% of the historical class came from obsolete builds, and counting those judges the work against noise it can't fix.

## What

- `scripts/crash_class_report.py` — buckets crash-class issues (unreachable / never-started / backend-died, matched on the auto-reporter's stable title shapes) by the `**Build status:**` stamp from #1547: `current` / `outdated` / `unknown` (pre-deflection builds). Windows default to the latest release; `--since`/`--version` override. Validated against August's real corpus (14 crash-class issues, all bucketed correctly).
- `tests/scripts/test_crash_class_report.py` — pins title mapping against real historical titles and locks the stamp literals to `bugReport.js`, so a reworded marker fails CI instead of silently zeroing the metric.

Maintainer-side tooling only — no app behavior change, no changelog entry.

Adds a CLI to measure crash-class issue recurrence by title, build freshness, and optional application version. Adds tests that lock historical title mappings and `bugReport.js` build-status markers. Review the hard `--limit 500` issue query because repositories with more matching issues can produce incomplete metrics.

### DIFF
--- a/scripts/crash_class_report.py
+++ b/scripts/crash_class_report.py
@@ -53,13 +53,44 @@ def latest_release_date() -> str:
     return data["publishedAt"][:10]
 
 
+_FETCH_LIMIT = 500
+
+
 def fetch_issues(since: str) -> list[dict]:
     out = sh(
         "gh", "issue", "list", "--repo", REPO, "--state", "all",
-        "--search", f"created:>={since}", "--limit", "500",
+        "--search", f"created:>={since}", "--limit", str(_FETCH_LIMIT),
         "--json", "number,title,body,createdAt",
     )
-    return json.loads(out)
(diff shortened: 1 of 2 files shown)


debpalash:
Bot-review triage: fixed both majors — `--version` now makes the Environment Version line authoritative (pure `classify_build` + parametrized tests), and the 500-issue fetch cap warns loudly instead of silently understating. Also asserting the full Build-status prefix in the lockstep test. The "v0.5.x labels unapproved" finding is noise — 0.5.0 is the current released version.

Issue #1164, opened 2026-07-16
[Bug] Can't reach the local OmniVoice backend — it may still be starting up, or it sto


OrcaGIS:
## Error

```
Can't reach the local OmniVoice backend — it may still be starting up, or it stopped. Wait a few seconds and try again; if it persists, restart the app (or check Settings → Logs → Backend).
ApiError@http://localhost:3901/src/api/client.ts:134:3
apiFetch@http://localhost:3901/src/api/client.ts:286:10

```

## Environment

**Version:** `0.3.22`
**Platform:** `Mozilla/5.0 (X11; Linux x86_64; rv:140.0) Gecko/20100101 Firefox/140.0`

## Recent actions

```
08:55:36 view:settings
08:58:05 view:studio
08:58:08 view:launchpad
08:58:10 view:studio
09:00:18 view:settings
09:01:09 view:studio
09:02:43 generate:start (audio)
```

debpalash:
Thanks for the report. The report itself tells me two useful things:

The stack trace shows unbundled source paths (`/src/api/client.ts`) served from port 3901 in Firefox — so this is a **source checkout running `bun run dev`**, viewed in a browser, not the desktop app. That matters because in browser/dev mode there's no desktop supervisor: no auto-restart, and none of the crash forensics v0.3.22 added (the crash notice with the exact exit reason is desktop-only). The auto-report can only carry the generic "can't reach" message — the *why* lives in your backend process.

Your timeline shows the failure at `generate:start`, the first generation of the session — so either the backend process (port 3900) had already exited earlier, or it died the instant the engine loaded. On Linux the usual culprit for the latter is the OOM killer.

To dig in, could you paste:

1. The last ~30 lines of backend output from the `bun run dev` terminal — or, if that's gone, the tail of `omnivoice_data/omnivoice.log` (plus `omnivoice_data/crash_log.txt` if it exists; it's only written on a hard crash).
2. Which engine/model the generation at 09:02:43 used, and your hardware (GPU, RAM).
3. If the process just vanished with no traceback: `journalctl -k --since today | grep -i oom` — an OOM kill will show up there.

FWIW, if you want the self-healing supervisor + crash diagnostics, that's the desktop build; dev mode intentionally leaves process management to you.

debpalash:
[...]

Issue #314, opened 2026-06-10
[Bug] Backend exits with code 106 — "No pyvenv.cfg file"; bootstrap never self-heals the broken venv


debpalash:
## Describe the bug
On launch, the backend immediately exits with:

```
Backend process exited (exit code: 106): No pyvenv.cfg file
```

The UI shows a setup-failure message. **Retry and app restart do not help** — the bootstrap keeps trying to use the same half-created/corrupted venv instead of detecting the missing `pyvenv.cfg` and rebuilding it.

Reported on Discord (v0.3.0). This is the same first-run-reliability family as #307: a broken venv state that the app can detect but currently can't recover from.

## To reproduce
1. End up with a venv directory whose `pyvenv.cfg` is missing (interrupted first-run install is the likely path).
2. Launch the app — backend exits 106 immediately.
3. Click Retry / restart the app — same failure forever.

## Expected behavior
Bootstrap validates the venv before spawning the backend (at minimum: `pyvenv.cfg` exists, interpreter runs). If validation fails, it rebuilds the venv automatically (or offers a one-click "Repair installation") instead of looping on the dead one.

Related: the in-flight setuptools-repair hardening work (follow-up to #253) covers package-level repair but not this venv-level corruption.

## Environment
- App version: v0.3.0
- Source: Discord report, 2026-06-02 ("OmniVoice Studio Backend Fails to Start (Exit Code 106)")

Issue #1375, opened 2026-08-05
[Bug] The crash notice fires with no evidence, and invites a bug report that cannot be diagnosed


debpalash:
Meta-issue, filed from a triage sweep. Several user reports are unactionable in principle, and the cause is on our side.

## The pattern

The auto-captured block in these reports reads:

```
Exit: process ended uncleanly (previous run)
Uptime before crash: 0 s
(log tail: empty)
```

Seen in #1243, #1336, #1345 (and #1282 is adjacent). Different OSes, different compute devices, different engines — the common factor is not a crash, it is the **run sentinel**.

## What is happening

On startup, the backend notices the previous run's marker was never cleared and synthesizes a crash record. `frontend/src/utils/backendCrash.ts::_adaptLastRunCrash` turns it into the same `BackendCrashMarker` the real shell-side crash path uses, with `exit_code: null`, `signal: null`, and `exit_desc: 'process ended uncleanly (previous run)'`.

`fetchLastRunCrash()` then returns that marker for **any** non-null record. There is no check that the record carries evidence. So a sentinel with `uptime_hint_s: 0` and an empty `log_tail` produces a full crash notice and a prefilled GitHub issue containing an empty code block.

## Why this matters

An unclean previous run is often completely benign — the machine slept, the user force-quit, a WSL VM stopped, Docker restarted. We show those users a crash notice and hand them a one-click bug report that contains nothing anyone can act on. They file it in good faith; it cannot be answered; it sits open.

That is a bad trade in both directions: noise in the tracker, and a user who reported a bug and got nowhere.

## What I am NOT proposing

**Do not simply suppress low-evidence records.** `uptime 0 s` with an empty tail can also be a genuine and serious failure — a backend that dies instantly before it can log anything. Hiding that would be worse than the noise.

## Options worth weighing

1. **Make the prefill honest.** When there is no log tail, the report template should say so and ask for `Settings → Logs → Backend` explicitly, rather than emitting an empty code block that reads as "there was nothing to report".
2. **Distinguish "unclean" from "crashed" in the notice wording.** The sentinel genuinely does not know which; saying so — and naming the benign causes (sleep, force-quit, VM stop) — would let most users self-resolve without filing.
3. **Gate the one-click report, not the notice.** Still tell the user what happened; only offer the prefilled issue once there is something in it.

(3) plus (2) is my inclination: it keeps every real crash visible while ending the unanswerable reports.

## Acceptance

[...]

Issue #1000, opened 2026-07-08
[Bug] Can't reach the local OmniVoice backend — it may still be starting up, or it sto


SupaNotch:
## Error

```
Can't reach the local OmniVoice backend — it may still be starting up, or it stopped. Wait a few seconds and try again; if it persists, restart the app (or check Settings → Logs → Backend).
v@tauri://localhost/assets/client-DU5MrUgu.js:2:3156
```

## Environment

**Version:** `0.3.11`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.4-arm64-arm-64bit`
**Python:** `3.11.14`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M4 Max`
**RAM:** `36 GB`
**Disk free:** `56 GB`
**Active TTS engine:** `mlx-audio`

## Recent actions

```
20:10:33 view:studio
20:11:40 generate:start (design)
20:19:50 view:dub
20:19:52 view:stories
20:19:53 view:audiobook
20:19:54 view:gallery
20:20:02 view:studio
20:20:12 generate:start (audio)
```

debpalash:
Thanks — this one's interesting and I'm actively digging into it. A few things already stand out from your own auto-captured log:

- You're on **v0.3.11**, which already ships crash forensics (#969) — and no crash was recorded. That suggests the backend process didn't die; it likely became unresponsive while still running.
- Your action log shows you navigating between tabs (Dub, Stories, Audiobook, Gallery) in the minutes after the stuck `generate:start (design)` — that's frontend-only, so the app window itself stayed responsive. Only the backend connection was dead.

That combination points at a specific theory: the 300s timeout guard that's supposed to catch a wedged generation may not actually apply cleanly to **mlx-audio** specifically, if MLX's Metal/Neural-Engine computation doesn't yield control the way it needs to for the guard's timer to fire — which would explain a full "can't reach backend" instead of a clean timeout error. I'm verifying this now.

One thing that would help either way: **Settings → Logs → Backend**, the ~30 lines around 20:11–20:20 (right when the design generation started through when you saw the error). That'll tell me whether the process was truly wedged or something else was happening. And: was this your **first** design generation with mlx-audio, or does it happen reliably on that engine?

debpalash:
[...]

PR #1550, opened 2026-08-14
feat(startup): bind the socket in ~1s and narrate startup step by step

debpalash:
## Why

The structural fix for the project's #1 failure class: nothing listened until every heavy import and migration finished, so "starting at step X", "stuck", and "dead" were all indistinguishable — and all filed as "Can't reach the local backend" (103 lifetime issues).

## What

**Backend** (`backend/main.py`, new `backend/core/startup_progress.py`):
- Module scope stays fast; heavy work splits into `_phase_a_build` (executor thread: prefs/env restore + #963 migration, yt-dlp overlay, cuDNN preload, torchaudio, model_manager, 30-router import fan-out — order preserved, literal imports so PyInstaller traces), `_phase_a_finalize` (event loop, no awaits → atomic route registration), `_phase_b` (old lifespan body; handles on `app.state`).
- **Eager mode** (auto under pytest, or `OMNIVOICE_EAGER_INIT=1`) runs everything at import — the ~100 lifespan-less `TestClient(app)` sites see today's exact app. Server runs get the deferred path.
- While starting: `/health` → 503 + current step; new `GET /startup/progress` → full step ledger (always 200); `StartupGateMiddleware` 503s all else with the `[starting]` marker (skip-Report convention, like `[shutting_down]`) and doubles as route-mutation safety.
- Deferred failure keeps import-crash semantics: traceback → `backend_err.log` (shell forensics + #314 venv-heal signatures), sentinel uncleared (#1164), exit 1 naming the failed step.
- `--health-check` 60→180s (now measures the full init); `--diagnose` runs Phase A up front (restored prefs, HF token, FFMPEG_PATH).

**Shell** (`backend.rs`, `bootstrap.rs`):
- `startup_progress()` probe — requires the `x-omnivoice-backend` marker header so a foreign responder can't narrate the splash; old backends → `None` → legacy behavior.
- Launch poll + supervisor reconnect wait emit a log line per step change ("Startup: Loading ML runtime (PyTorch)…") into the splash's existing log stream — no stage-enum or splash changes needed.

**Compat verified**: old shell → new backend (503 fails the string-sniff and deep probe → keeps polling, 300s budget unchanged); new shell → old backend (no endpoint → silent fallback); Docker `HEALTHCHECK` (`curl -f` fails on 503 as it did on refused); `uvicorn main:app` third parties unchanged.

## Tests

- `tests/test_startup_progress_early_bind.py`: **integration** — real uvicorn spawn must answer marker-stamped `/startup/progress` within 8s (fails before this change); gate 503-with-marker then inert; ledger state machine; AST guard pinning the #963 migration→prefs→ytdlp and cudnn-before-torch ordering inside `_phase_a_build`.
- Rust: `startup_progress_parses_fields_and_requires_the_marker` (loopback stub).
- Full backend suite green locally (5698 passed; the 35 pre-fix failures were the two subprocess embedders needing `OMNIVOICE_EAGER_INIT=1`, now set; route snapshot regenerated: +`GET /startup/progress`).
- `test_shutdown_preload_race_1000` source-guard updated to the app.state call shape (bound unchanged, ≥15s).
- Docs: docker.md "watching startup" section.

Backend startup now binds the socket early, defers heavy initialization, and exposes progress through `/health` and `/startup/progress`, while preserving eager initialization for tests and embedders. The shell logs startup steps, health checks allow up to 180 seconds, and diagnostics run Phase A before collection. Review deferred failure handling and bounded Phase A shutdown waits because timing and cancellation errors could affect crash forensics or process termination.


debpalash:
Bot-review triage: all four technical findings fixed — shutdown now joins the Phase A executor thread via a bounded event wait (cancelling the task can't stop the thread; #1000 class), the fail-path beat is `await asyncio.sleep` so the loop can actually serve the last /startup/progress poll, dump_api_routes assigns EAGER_INIT outright, and the integration test's child uses DEVNULL instead of an undrained pipe. Highlights-ref finding is linter-sanctioned noise.

debpalash:
On the shutdown-deadline P1: the desktop shell has always SIGKILL'd the backend 2s after SIGTERM (lib.rs ExitRequested), so every graceful-shutdown wait — including the pre-existing 20s `_cancel_and_await_tasks` bound from PR #1002 — is already cut short on desktop quit, and the uncleared sentinel is reconciled by the shell's quitting/kill-intended flags plus #1375's unclean-vs-crash wording, so no phantom crash reaches the user. The Phase A join matters where the grace is real (Docker's 10s, plain SIGTERM server runs) and is bounded + conditional there. No regression; keeping it.

debpalash:
Second-pass triage: submission race closed — `_phase_a_started` set before `run_in_executor`, submission wrapped in `asyncio.shield` (a cancel on a queued callable would otherwise strand `_phase_a_finished` forever and stall the join), and the build wrapper sets finished on every exit path. Pinned by `test_phase_a_thread_join_contract`. CodeQL: comments added to the new bare excepts, websocket branch return made explicit; the BaseException handler is deliberate (converts deferred-startup failure to process death) and documented.

PR #1551, opened 2026-08-14
test(shell): backend-lifecycle fault-injection harness

debpalash:
## Why

Nothing anywhere ran the supervisor/startup path against a real dying child and asserted the *named diagnosis* that comes out — the wiring between a death shape and the branch that describes it, which is where 61% of the historical "can't reach the backend" class died undiagnosed. This is the permanent regression rig for lifecycle fixes (reliability-cycle task, follows #1547/#1548/#1550).

## Seam (behavior-neutral when unset, unit-pinned)

- **`OMNIVOICE_BACKEND_CMD`** (JSON argv / whitespace form) runs any command as "the backend" inside the real `spawn_backend`: venv bootstrap + ffmpeg resolution skipped, everything else real — err-log run offsets, drainers, env pinning, OS pipes, spawn-failure diagnostics. Doubles as a manual QA tool.
- **`OMNIVOICE_LOG_DIR`** — explicit log + crash-marker dir (per-test isolation; support-friendly).
- **`OMNIVOICE_STARTUP_BUDGET_S` / `OMNIVOICE_SUPERVISOR_POLL_MS`** — harness-only timing; production defaults (300s / 2s) pinned by `timing_overrides_default_to_production_values`.
- Lifecycle fns genericized over `tauri::Runtime` (MockRuntime app); `tauri = { features = ["test"] }` dev-dep.

## Scenarios (`tests/backend_lifecycle.rs`, 9 tests, ~63s serial)

Children are this test binary re-invoking itself (`scenario_child`) — genuine exit codes, signals, and pipe ordering on all 3 OSes, no system python. Each asserts the exact user-visible outcome: port conflict speaks the `detectHints`-localizable phrasing; a chained traceback's root cause survives into the Failed message *and* the marker; spawn failure writes **no** bogus marker; slow start names the budget + last stderr; a crash loop announces 3 restarts with markers recorded before each and gives up naming the last exit; SIGKILL is named signal 9; a deliberate kill yields silently; a deferred-startup FATAL (#1550) carries the failed step into the diagnosis, forensics, and splash narration.

## CI

Added to the 3-OS `tauri-cross-platform` matrix with `--test-threads=1`. Local: 149 lib + 9 harness green.

Adds a cross-platform backend lifecycle fault-injection harness that exercises real child-process failures through supervisor and startup paths. It supports command and log overrides, configurable timing, nine serialized MockRuntime scenarios, and CI coverage to verify user-facing diagnostics and restart behavior without system Python dependencies. Review the Windows test manifest and environment-controlled timing and command paths for platform-specific or test-isolation risks.


debpalash:
Triage: the self-injection major is fixed — the scenario gate now carries the arming process's PID, so the parent's own `scenario_child` stays inert under any thread count (verified with a default parallel `cargo test` locally). Keeping the absolute manifest path in the linker args: it's a local, non-shipped build artifact, and a cwd-relative path would silently depend on cargo's rustc working directory — the absolute form is verified working on the Windows runner.

PR #1555, opened 2026-08-14
docs(readme): lead with download + first clone; seed benchmarks page

debpalash:
Applies the conversion lessons from the unsloth teardown:

- **Quickstart first**: per-OS download buttons, install guides, and a three-step first-clone walkthrough now sit directly under the hero in both `README.md` and `README_CN.md` — before What's-new and Features.
- **docs/benchmarks.md** (new): measured per-engine/per-device numbers (RTF, VRAM), produced only by `scripts/bench_pipeline.py` on named hardware — community-contributed rows, no estimates. Linked from both READMEs and `docs/performance.md`.

No app code touched. Changelog entry follows in a fixup commit with this PR's number.

Moved English and Chinese README quickstart content below the hero section and added measured engine/device benchmark documentation with links from both READMEs and the performance guide. Updated the benchmark harness to report warm-run RTF and CUDA peak VRAM, reject invalid RAM readings, and leave VRAM blank on MPS/CPU. Review the benchmark methodology and hardware results for accuracy before merging.


debpalash:
Triage of the 4 bot findings — all addressed:
- **Greptile P1 (harness can't produce RTF/VRAM)**: real — `bench_pipeline.py`'s tts stage now emits RTF next to each warm measurement (from the returned tensor length / backend sample rate) and CUDA peak VRAM; the doc now collects exactly what the harness prints, VRAM blank on MPS/CPU by design.
- **CodeRabbit (NaN RAM floor)**: fixed — the stage guard now refuses non-finite free-RAM readings (`OMNIVOICE_BENCH_FLOOR_GB=0` to force).
- **CodeRabbit (submission schema)**: fixed — dropped the redundant Hardware column, defined every column, Source = the contributing PR.
- **CodeRabbit (badges all point at the release page)**: per-asset URLs aren't stable (filenames embed the version), so the badges keep the latest-release link and a caption now says exactly that.

debpalash:
Second harvest (4 new):
- **CodeQL empty-except ×2**: comments added — VRAM counters are bonus metrics that degrade to absent, never abort the run.
- **Greptile (engine identity unprinted)**: real — the tts stage now prints `engine: <id> (<class>)` at start, matching what the doc promised.
- **CodeRabbit Major (cold row excludes model load)**: mostly off — weights load lazily on first `generate()` (the resolver only picks/validates the backend), so 'model load + first synth (cold)' is honest; the line-152 `get_model()` evidence is from the clone stage, not tts. Hardened anyway: VRAM tracking now starts before resolution so resolver allocations are inside the peak. Timing the resolver inside the synth number would pollute it — declined.

debpalash:
Third harvest (2 new, both Greptile, both real, both fixed): peak VRAM now reads `max_memory_reserved` (the capacity a card actually needs, not just live tensors), and adapter engines print their resolved model id (`engine: mlx-audio [mlx-community/Kokoro-82M-bf16] (...)`) so rows can't blur distinct models behind one backend id.

debpalash:
Fourth harvest (1 new, Greptile, real, fixed): subprocess-isolated engines (SubprocessBackend subclasses) now print `peak VRAM: n/a` instead of the parent allocator's misleading near-zero; doc updated to say the column stays blank for them.

debpalash:
Fifth harvest (2 new, both Greptile, both real, both fixed): (1) omnivoice-gguf spawns a binary per generate but inherits TTSBackend directly, so the isinstance gate missed it — replaced with a declarative `runs_out_of_process` attribute on TTSBackend, set by SubprocessBackend and omnivoice-gguf (duck-typed for the same module-purge reason as `_is_subprocess_isolated`; fixes the class, not the instance). (2) sherpa-onnx rows now derive model identity from `_model_dir`'s basename.

debpalash:
Sixth harvest (1 new, Greptile, real — CosyVoice stores neither sniffed attribute): fixed the class this time. `TTSBackend.model_identity()` (default None) is now the contract — mlx-audio returns its repo id, sherpa-onnx its model-dir basename, cosyvoice its resolved dir basename — and the harness just asks. No more per-engine attribute knowledge in the profiler.

PR #1556, opened 2026-08-15
docs(engines): a guide for every engine + index; fix two engine-metadata bugs

debpalash:
Task from the unsloth teardown: per-model guide pages are their docs template, and most of our issue volume is engine-specific.

- **21 new pages** under `docs/engines/`: all 10 TTS engines that had none (including the default `omnivoice`), all 10 local ASR engines, and an index `README.md` mapping every engine → guide → platform → enablement. Existing 8 pages untouched; filenames match the ids backend error strings already point at.
- **READMEs** (EN+CN) link the index from both engine-matrix sections.
- **Fact-check fallout** (every page was verified against the registries): fixed the KittenTTS docstring ('Jasper' vs `expr-voice-2-f`), unified the isolated-ASR sidecar/preflight model resolution (`ASR_MODEL_FW` override → `ASR_MODEL_FASTER`), corrected moonshine's install hint to the packages the backend actually imports.

Targeted tests green: 462 ASR/sidecar/sherpa-related + CJK guard.

Added 21 engine guides and an index covering engine selection, platform support, configuration, installation, and limitations, with updated English and Chinese README matrices. Aligned isolated Faster-Whisper model resolution across the sidecar and download preflight, and corrected KittenTTS and Moonshine guidance. Review `ASR_MODEL_FW` precedence and engine-specific installation instructions because incorrect values can select the wrong model or block setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,10 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Docs
 - The READMEs now lead with download buttons and a three-step first-clone walkthrough, and a new benchmarks page anchors measured per-engine/per-device numbers on the in-repo harness (#1555)
+- Every engine now has its own guide — 21 new pages under docs/engines plus an index covering all 16 TTS and 11 ASR engines, linked from both READMEs (#1556)
 
 ### Fixed
+- The crash-isolated ASR sidecar and its download preflight now agree on which model to load — setting the shared faster-whisper model variable applies to both variants instead of the sidecar quietly using a different one (#1556)
 - "Ready" now requires the deep health probe (a working database-backed route), not just the identity probe — a backend whose install broke underneath can no longer be announced up while every real request fails (#1548)
 - Supervisor restarts after repeat crashes now back off (immediate, then 5s, then 15s) instead of respawning back-to-back, so a tight crash loop can't burn the whole restart budget in seconds (#1548)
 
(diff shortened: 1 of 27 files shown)


debpalash:
Harvest triage (12 findings):

**Fixed (8):** empty-env preflight fallback now matches the sidecar; gpt-sovits guide requires TLS-or-tunnel guidance for non-loopback servers with a plaintext warning; kitten log-level wording; first-use download guidance on both OmniVoice pages (install ahead via Model Catalogue); MOSS-Nano installs via `uv pip install -e .`; index links the install guides incl. the Gatekeeper right-click → Open step; README matrix aligned to the code's `gpu_compat` (CosyVoice macOS = CPU not MPS; IndexTTS = CUDA/CPU incl. macOS CPU; GGUF = CUDA/MPS/CPU — the registry drives routing, so it is the source of truth); README_CN engine counts synced to 16 TTS / 11 ASR.

**Rejected with rationale (4):**
- *moss-tts-nano env var*: the cited runtime is `moss_tts_v15` — a different engine. MOSS-TTS-**Nano** is in-process and reads `OMNIVOICE_MOSS_TTS_MODEL` (tts_backend.py:1051). Doc is correct.
- *VoxCPM2 OpenRAIL-M warning*: VoxCPM2 is Apache-2.0 (README matrix row); the OpenRAIL-M engine is Supertonic-3, whose page and license gate already cover it.
- *whisperx 'platform-divergent default' (Critical)*: the auto-detect ladder IS platform-divergent by design — hardware-dependent engine selection is sanctioned (CLAUDE.md behaviour-vs-performance clarification) and the two pages describe the same ladder consistently. Documenting a fictional single default would make the docs lie.
- *lowercase README.md rename*: GitHub auto-renders README.md as the directory index — that behavior is the whole point of the file; `index.md` gets no such treatment.

debpalash:
Round 3 (1 new, valid): the CN blanket "every engine runs locally" note now excludes the optional OpenAI-compatible remote client, matching the EN wording.

Issue #981, opened 2026-07-06
[Bug] There is no way to load an other model than OmniVoice TTS


bistory:
### Before filing

- [x] I searched [existing issues](https://github.com/debpalash/OmniVoice-Studio/issues?q=is%3Aissue) and this isn't a duplicate.
- [x] I'm on the latest release (or `main`) — older builds may already be fixed.

### What happened?

OmniVoice TTS is the base model, in the settings, you can download other models, however: there is no way in the UI to switch models.
I'm trying to test video dubbing on a Mac M1 (because I'm not able to make the Linux version work at the moment); and it's stuck with a heavy model while I'd like to use Llama-OuteTTS 1.0 1B.

### Steps to reproduce

1. Install OmniVoice
2. Download Llama-OuteTTS 1.0 1B
3. Make dubbing for a video
4. Check on the top right-corner, the loaded model is always OmniVoice TTS

### What did you expect instead?

I would like to select an other model.

### Operating system

macOS (Apple Silicon)

### How did you install it?

Desktop app (installer / AppImage)

### Version

v0.3.11

### Compute device

Apple Silicon (MPS)

### Active TTS/ASR engine

OmniVoice TTS

### Logs / diagnostic bundle

```text

```

debpalash:
Fixed on `main`. mlx-audio multiplexes 7+ curated models (Kokoro, CSM, Qwen3-TTS, Dia, Chatterbox, MeloTTS, and yes — OuteTTS) behind one "mlx-audio" engine, but there was genuinely no way anywhere in the UI or API to pick which one loads; it always defaulted to Kokoro regardless of what you downloaded. Settings → Engines now shows a model picker on the mlx-audio row once you select it — pick OuteTTS there and it'll actually load and use it, no restart needed to switch between curated models. Ships in the next release.

debpalash:
Closing — fixed on main, ships next release.

Issue #1529, opened 2026-08-13
[Bug] ASR backend initialization failed: CUDA failed with error CUDA driver version is


patmansk:
## Error

```
ASR backend initialization failed: CUDA failed with error CUDA driver version is insufficient for CUDA runtime version
Error: ASR backend initialization failed: CUDA failed with error CUDA driver version is insufficient for CUDA runtime version
    at EventSource.<anonymous> (http://192.168.178.104:3900/assets/main-app-SORYwEQi.js:17:10555)
```

## Environment

**Version:** `0.4.2`
**Platform:** `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36`
**OS:** `Linux-7.1.4-200.nobara.fc44.x86_64-x86_64-with-glibc2.39`
**Python:** `3.12.3`
**Compute device:** `cuda`
**GPU:** `AMD Radeon RX 7900 XTX (24 GB VRAM)`
**CPU:** `AMD Ryzen 7 8845HS w/ Radeon 780M Graphics`
**RAM:** `28.2 GB`
**Disk free:** `367.6 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Deployment mode:** `server`
**Last backend response:** 3 s before this report

## Recent actions

```
08:24:35 view:settings
08:24:55 view:launchpad
08:24:57 view:studio
10:24:22 view:dub
10:25:50 dub:upload
```

## What I was doing

1. Opened OmniVoice Studio web UI at http://192.168.178.104:3900
2. Navigated to Voices → New → Clone
3. Uploaded a German audio clip for voice cloning
4. ASR backend failed to initialize with CUDA driver version error

## Additional context

Running inside a Podman Quadlet container (rootless pod) on Nobara Linux 44.
ROCm version on host: 7.8.0 (rocm-smi-lib 7.8.0)
The ROCm :rocm image ships ROCm 7.2.4 internally.

TTS model loads successfully on GPU (cuda/ROCm):
  "Loading VoiceStudio model on device: cuda"
  "GPU pool sized to 4 worker(s) — 23.9 GB free"

But ASR (Whisper/faster-whisper) fails when trying to use CUDA:
  "CUDA driver version is insufficient for CUDA runtime version"

This suggests the ASR backend uses a different CUDA/ROCm runtime 
than the TTS backend, possibly a version mismatch between the 
[...]

PR #1557, opened 2026-08-15
feat(settings): compute-device override (auto / CUDA / ROCm / XPU / MPS / CPU)

debpalash:
Task 4 of the unsloth teardown: an explicit device override ends the "auto-detect picked wrong" issue class.

- **One choke point**: the override is applied inside `_probe()`'s family selection, so `resolve_routing`, `get_best_device()` (which delegates its family decision to the probe), and every Model Catalogue badge inherit it with zero per-consumer changes.
- **Resolution** `OMNIVOICE_DEVICE` > Settings pick > auto — the engine-selection (#981) pattern. The env pin disables the UI control and says so.
- **Safety**: only detected families are selectable (UI + 400 from the API); CPU is always honorable; an honored pin and an ignored request both leave a note in `HostCaps.notes` so diagnostics explain why a CUDA host shows CPU chips.
- **Restart contract**: caps are immutable per process; the endpoint reports `applied` vs `value` so restart-required is shown truthfully (Performance tab is already `restart: true`).
- Parity: identical behavior on all 3 OSes; options varying by host hardware is the sanctioned performance/hardware variance.

Tests: 7 probe-contract tests (steer-not-invent, env-beats-prefs, degraded-probe), 4 API tests (validation, env pin, restart flag), 4 panel tests (families-only options, PUT + restart note, env-pin disable, failed-PUT re-sync), restart-badge lockstep updated. 121 backend device/routing/prefs tests + 241 settings-component tests green. Docs: performance.md knob row + engines-index pointer.

Adds compute-device override support for Auto, CUDA, ROCm, XPU, MPS, and CPU through settings or `OMNIVOICE_DEVICE`, with API and UI state reporting, detection filtering, persistence, localization, and documentation. The selected family now controls routing, `get_best_device()`, Model Catalogue badges, CTranslate2 ASR, and the ASR sidecar, with changes applied after restart. Review fail-safe CPU behavior for broken probes and environment-variable precedence, because unavailable or pinned requests are ignored, reported, and recorded in `HostCaps.notes`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Added
 - The bug reporter notices when you're on an outdated build and offers the latest release before filing — with a "File anyway" escape hatch — and stamps a `Build status` line into every report so up-to-date reports are tellable from stale ones (#1547)
-- Settings → Performance & Device gains a compute-device override (Auto / CUDA / ROCm / MPS / CPU, or `OMNIVOICE_DEVICE`) — pin the device when auto-detect picks wrong; only devices your machine actually has are offered (#1557)
+- Settings → Performance & Device gains a compute-device override (Auto / CUDA / ROCm / XPU / MPS / CPU, or `OMNIVOICE_DEVICE`) — pin the device when auto-detect picks wrong; only devices your machine actually has are offered (#1557)
 
 ### Docs
 - The READMEs now lead with download buttons and a three-step first-clone walkthrough, and a new benchmarks page anchors measured per-engine/per-device numbers on the in-repo harness (#1555)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Added
 - The bug reporter notices when you're on an outdated build and offers the latest release before filing — with a "File anyway" escape hatch — and stamps a `Build status` line into every report so up-to-date reports are tellable from stale ones (#1547)
+- Settings → Performance & Device gains a compute-device override (Auto / CUDA / ROCm / MPS / CPU, or `OMNIVOICE_DEVICE`) — pin the device when auto-detect picks wrong; only devices your machine actually has are offered (#1557)
 
 ### Docs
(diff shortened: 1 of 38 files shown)


debpalash:
Harvest triage (7 findings + the oxfmt CI failure):

**Fixed (6):**
- *CT2 ASR ignores the override* (Greptile — the real catch): `_ctranslate2_cuda_ok()` now gates on the probe's family first, covering whisperx + faster-whisper through the one shared gate; the isolated-ASR sidecar asks `detect_host_caps()` instead of re-deriving its own device. Bonus: the sidecar now also honors the #1529 ROCm/CT2 rule it previously bypassed.
- *applied/ignored semantics*: `override_ignored` exposed and shown — an env pin naming absent hardware now says 'Auto is in effect' instead of implying a restart would help.
- *i18n*: all 8 strings + 5 family labels added to all 21 locales (translated, not English-copied); labels render through `t()`.
- *save error cleared by re-sync*: the failure path re-syncs state without touching the error.
- *panel tests raced the GET*: waits now target the loaded 3-option state.
- *test isolation*: cleanup drops OMNIVOICE_DEVICE before the cache-restoring re-probe.

Plus: `xpu`/`intel` search keywords, XPU named in the changelog, oxfmt formatting fixed (the CI failure).

debpalash:
Round 2 (4 findings + the CI failure, all fixed): the CI failure was `test_ctranslate2_never_gets_cuda_on_a_rocm_build` mocking the low-level helpers but not the probe the new gate consults — it now pins the probe family (and was silently order-dependent on the lru_cache before). CodeRabbit's fail-safe point taken fully: a broken probe now means CPU in both the CT2 gate and the sidecar, never a torch guess (regression test added). Greptile's two: env-pinned+not-detected renders both facts in one subtitle, and the two error strings are now translated in all 21 locales (plus literal 'Auto' localized in the 7 locales that translate the option label).

debpalash:
Round 3 (CI failure, same class as round 2): `test_faster_whisper_float16_unsupported_falls_back_to_int8` faked `torch.cuda` but not the probe the override gate consults first — probe family now pinned to cuda so the fallback chain under test is reachable on cpu-only CI. Full isolated suite (288) green.

debpalash:
Round 4 (1 new, rejected): `settings.perf_save_failed` IS defined in en.json — and in all 21 locales — since the round-2 commit (verified: every locale has both `perf_save_failed` and `device_load_failed`). The finding reviewed a stale hunk.

PR #1562, opened 2026-08-15
fix(events): sync endpoints dropped their WS events — rename/delete left every tab stale

### LINKED ISSUE #1561: [Bug] Renaming a voice makes the list look stale or empty in other tabs (WS event dropped from sync endpoints)
### Problem

After renaming a cloned voice from the profile page, the voices panel does not refresh — the old name keeps showing until a full reload, and a reload landing in the wrong moment can present an **empty** voices panel, which looks like all voices were deleted (they are not: the rows and reference audio are intact; a fresh load shows everything again).

### Root cause (verified on main @ e4c1ef0d)

- `PUT /profiles/{id}` (`update_profile`), `DELETE /profiles/{id}`, `DELETE /profiles/{id}/consent`, the history star/clear/delete routes and the export recorders are **sync** FastAPI endpoints → Starlette runs their bodies in **threadpool workers**, where `asyncio.get_running_loop()` raises `RuntimeError`.
- `event_bus.emit()` (backend/core/event_bus.py) treated that exception as "no event loop — event dropped" (DEBUG log only), so the `profiles` WebSocket event was silently dropped for exactly the mutations users notice most.
- The frontend only refetches the voice list on that WS event (`useRealtimeEvents` → `loadProfiles`), so every open tab kept stale state.
- Compounding it: the initial list load ran exactly once with no retry (`useAppData`), so a transient failure on first fetch left an empty panel until some later WS event or reload — the "my voices are gone" presentation.

Evidence: live probe against a running backend — WS client connected, PUT rename returned 200 and the row updated in SQLite, but no `{"kind":"profiles"}` frame ever arrived on the socket (3s window). Async endpoints (lock/unlock/consent-record) emit fine; only sync ones drop.

### Repro

1. Open the app in two tabs (or rename and watch the same tab's sidebar).
2. Rename a cloned voice (profile page → Edit → rename → Save).
3. The list still shows the old name; the other tab never updates.

### Fix

`emit()` captures the serving loop in `subscribe()` and hands off from foreign threads via `call_soon_threadsafe`; async callers unchanged. Plus the initial loads retry until first success (WS-triggered reloads keep the keep-previous-list behavior). Regression test included (fails on old emit, passes on fix); PR to follow.

— happy to link the PR: paoloantinori

paoloantinori:
Fixes #1561.

## What

- **backend/core/event_bus.py** — `emit()` called from a threadpool worker (sync FastAPI endpoint: profile rename/delete/revoke-consent, history star/clear/delete, export recorders) used to hit the `RuntimeError` branch of `asyncio.get_running_loop()` and silently drop the event. `subscribe()` now captures the serving loop and foreign threads hand off via `call_soon_threadsafe`; async callers take the same `create_task` path as before.
- **frontend/src/hooks/useAppData.js + utils/initialLoadRetry.js** — the initial list loads retry (exponential backoff) until **first** success; WS-triggered reloads keep the existing keep-previous-list-on-failure behavior. Loaders gained `{ rethrow: true }` for the initial path so the retry actually engages — they swallow errors by design everywhere else, and without this the retry resolves on attempt one (caught during self-review).

## Why

The UI refetches the voice list only on the WS `profiles` event, so a rename left every open tab stale; a reload in that window with a transient first-fetch failure presented an empty voices panel — indistinguishable from data loss to a user, though nothing was deleted (#1561 has the full repro and the live-socket evidence).

## Verification

- `tests/test_event_bus_thread_emit.py`: fails on the original emit (re-verified by stashing the fix), passes with it. Drives the real shape: serving loop in its own thread, `emit()` from a loop-less worker.
- Live two-instance probe (isolated `OMNIVOICE_DATA_DIR`, throwaway ports): PUT rename → `{"kind":"profiles","action":"updated"}` arrives on the WS with the fix; nothing arrives within 3s on the original build.
- `frontend` vitest suite (4 tests incl. a wiring pin that fails if `{ rethrow: true }` ever disappears from the call site) and `bun run build` both green.
- `tests/test_api.py`: 4 failures that pre-date this branch (non-loopback 401-vs-403 class; identical with/without this diff).

## Notes for reviewers

- Events emitted before any WS client ever subscribed are still dropped (no serving loop captured) — same as before; there is nobody to notify. MCP-side profile mutations go through async paths that already worked.
- The altitude call (loop capture in `subscribe()` vs lifespan vs per-endpoint asyncification) was reviewed and `subscribe()` chosen: it is the first point the serving loop provably exists, and converting endpoints to async would move their DB I/O onto the event loop for no gain.

The event bus now forwards WebSocket events from synchronous threadpool endpoints, and initial frontend list loads retry with exponential backoff. This prevents stale cross-tab data and empty lists after transient failures while preserving existing WebSocket reload behavior. Review event-loop lifecycle handling and retry cancellation paths; four `tests/test_api.py` failures are pre-existing.


paoloantinori:
Heads-up: the `Tests (backend + frontend)` failure on the latest run is **pre-existing on `main`** — upstream's own CI run for `main` @ `31886909245` (commit `fix: require remote voice render parity`, 13:19Z today) fails the identical test, before any of my commits:

```
FAILED tests/test_watermark_route_coverage.py::test_every_synthesis_module_routes_through_mark_synthetic
AssertionError: Modules synthesize audio but never reference the mark_synthetic chokepoint (EU AI Act Art. 50(2), #1169): ['worker/transport/server.py']
```

My branch contains no changes to `worker/transport/server.py`, watermark code, or that test — the failure arrived with the merge of upstream `main` (the branch was conflict-free against `e4c1ef0d` and is kept current with `main` per repo convention). Nothing in this PR touches that path; every check that this PR *can* affect (lint, format, build, both new suites, CodeQL, Bandit, gitleaks, Greptile, CodeRabbit) is green.

PR #1564, opened 2026-08-15
fix(test): watermark producer scan must match code, not prose

debpalash:
Main's CI is red since ee35d238: its added *comment* mentions `backend.generate()`, and the watermark coverage guard (`tests/test_watermark_route_coverage.py`) greps raw source — so `worker/transport/server.py` became a "producer that never marks" without synthesizing anything.

**Fix (class, not instance):** blank COMMENT/STRING token spans (via `tokenize`, layout preserved; unparseable source falls back to a raw scan so the guard stays fail-closed) before matching `_SYNTH_CALL`, and apply the same rule to the allowlist staleness check. A new self-test pins the regression: prose mentions in comments/strings/docstrings never make a module a producer; a real call still does.

Fail-before/pass-after: the guard fails on current main, passes with this change; all 16 tests in the module green.

The coverage test now scans code-only content, with version-aware f-string handling, to prevent comments, strings, and docstrings from falsely identifying producers. It applies the same filtering to `mark_synthetic` checks and allowlist validation, while retaining a raw-scan fallback for unparseable source. The main risk is tokenization behavior for malformed or version-specific source.

### DIFF
--- a/tests/test_watermark_route_coverage.py
+++ b/tests/test_watermark_route_coverage.py
@@ -86,14 +86,27 @@ def _code_only(src: str) -> str:
     not deletion, to keep the layout the regexes were written against).
     Unparseable source falls back to the raw text — fail closed, a module we
     can't tokenize still gets scanned.
+
+    f-strings stay conservative (Greptile P1 on #1564): on Python ≤3.11 the
+    whole f-string — replacement expressions included — is ONE STRING token,
+    so blanking it would let ``f"{backend.generate(t)}"`` evade the guard.
+    f-prefixed strings are therefore kept raw there (a literal f-string
+    *mentioning* a primitive false-positives toward the allowlist — fail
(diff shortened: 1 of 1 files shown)


debpalash:
Both findings addressed in 7ecabba6:

- **Greptile P1 (f-string evasion)**: confirmed — on ≤3.11 the whole f-string is one STRING token. f-prefixed strings are now kept raw there (conservative: literal f-string prose false-positives toward the allowlist, fail closed), and on 3.12+ only literal `FSTRING_MIDDLE` text is blanked so replacement-field code stays scannable. Self-test now asserts `f"{backend.generate(...)}"` is caught.
- **CodeRabbit (prose could certify a module)**: the `mark_synthetic`-reference check is now code-only too — a comment can't satisfy the guard. Also self-tested.

Issue #315, opened 2026-06-10
[Bug] Voice clone: first render is perfect, second render onward has static noise and slow playback (Windows)


debpalash:
## Describe the bug
On Windows, the **first** voice-clone render sounds perfect. Every render **from the second one onward** in the same session comes out with static noise and slowed-down speed.

Reported on Discord with a screenshot of the Voice Clone tab (Japanese text, v0.3.5, Windows; backend log badge showed 24 errors at the time).

This is distinct from #284 (noise from the very first render, Win11 + RTX 5060) but is likely the same family — possibly state not being reset between generations, or the same RTX 50xx / `torch.compile` path as #278. The "works once, then degrades" repro is the useful new signal: whatever breaks is introduced by the first generation run.

## To reproduce
1. Windows, v0.3.5, Voice Clone tab with a reference WAV.
2. Synthesize — output is clean.
3. Synthesize again (same or different text) — output has static noise and plays slow.

## Expected behavior
Every render in a session has the same quality as the first.

## Related
- #284 (noise-only TTS on Win11/RTX 5060)
- #278 (torch.compile failures on RTX 5060)

## Environment
- OS: Windows
- App version: v0.3.5
- Source: Discord thread "Voice clone issue", 2026-06-08

Issue #278, opened 2026-06-04
[Bug] Archetype Preview/TTS Fails with torch.compile Enabled (RTX 5060, OmniVoice Studio 0.3.5)


Thomas27-0379:
Bug: Archetype Preview/TTS Fails with torch.compile Enabled (RTX 5060, OmniVoice Studio 0.3.5)

**Environment**
OmniVoice Studio: 0.3.5 Appimage
GPU: NVIDIA RTX 5060
OS: Linux
Backend: CUDA enabled
Model successfully loads on GPU

**Summary**

The OmniVoice model loads successfully and reports approximately 15 GB of available VRAM. However, generating Archetype previews (e.g. German Companion, German Narrator) intermittently fails with HTTP 503 errors and backend exceptions.

The backend reports the failure as a possible out-of-memory condition, but the actual exception appears to be related to an interaction between torch.compile(), TorchDynamo, FX tracing, and CUDA graph execution.

**Relevant Logs**

Model initialization succeeds:

GPU pool sized to 4 worker(s) — 15.0 GB free / 2.5 GB per job
OmniVoice model loaded successfully.
torch.compile applied.
Preload complete — model ready.

Generation subsequently fails with:

RuntimeError:
Detected that you are using FX to symbolically trace a dynamo-optimized function.
This is not supported at the moment.

Later failures originate from:

torch/_inductor/cudagraph_trees.py
AssertionError

The UI then reports:

RuntimeError:
TTS engine stopped mid-generation.
This usually means it ran out of memory.

However, the underlying exception does not appear to be VRAM exhaustion.

**Additional Observations**

The UI queries:

GET /api/settings/perf/torch-compile-disabled

which suggests OmniVoice already exposes a torch.compile-related setting.

Several Archetype requests fail:

POST /archetypes/ml_german_companion/use -> 503 Service Unavailable
POST /archetypes/ml_german_narrator/use -> 503 Service Unavailable

After some retries, one request eventually succeeds:

POST /archetypes/ml_german_companion/use -> 200 OK

suggesting the model itself is functional and the failure may be related to compilation/runtime state.

Separate Issue: pyannote Download Failure

During setup, the recommended model fails to install:

pyannote/speaker-diarization-3.1

Error:

An error happened while trying to locate the file on the Hub and we cannot find the requested files in the local cache.

The installer retries several times and then aborts.

Subsequent installation attempts often return:

POST /models/install -> 429 Too Many Requests

[...]

PR #1565, opened 2026-08-15
feat(omnivoice): voice prompts that survive restarts + opt-in FlashInfer (~2.2x)

debpalash:
Ports from the upstream k2-fsa/OmniVoice teardown, every change verified with real generated voice samples (design, clone, clone-from-disk-prompt, eager-vs-FlashInfer, ref+instruct+`[laughter]`) plus a Whisper round-trip transcript check on each.

## 1. Voice-clone prompts persist across restarts
- Ports `VoiceClonePrompt.save()/.load()` (upstream format v1, `weights_only=True`-safe) onto the vendored model.
- New disk layer under the in-memory prompt LRU: `DATA_DIR/prompt_cache/`, keyed identically (ref path + mtime + ref_text + preprocess flag), ~10 KB per voice, 32 newest kept, atomic writes, corrupt files dropped-and-reencoded. `OMNIVOICE_PROMPT_DISK_CACHE=0` opts out.
- Effect: the first generation of a session with a known voice skips the reference re-encode **and** the auto-transcription pass. Verified across two real processes: run 1 `encodes=1`, run 2 `encodes=0`, same voice.
- Best-effort by contract — single-use dub refs never touch disk (same scan-resistance as the LRU), any disk failure degrades to the old re-encode. 9 new tests.

## 2. Opt-in FlashInfer acceleration (CUDA)
- Ports upstream's `omnivoice_flashinfer.py` (packed CFG attention, fused RMSNorm/RoPE/GEMM, optional CUDA graphs), with the unmasking schedule adapted to our `num_step + 1` divergence so packed and eager decode identically.
- `OMNIVOICE_FLASHINFER=1|graph`, applied instead of `torch.compile` for that session. Same #278 contract as compile: missing package / non-CUDA / apply failure / runtime failure each log a named reason and continue on the standard path (runtime failures classify → unapply → retry once → session latch). `graph` mode reuses the #315 single-thread affinity (CUDA-graph state is thread-local).
- Measured on this branch: **2.20x** at batch=1 (RTF 0.041 → 0.019, RTX 4090, fp16, num_step=16), transcript-identical output. `flashinfer` stays an optional, undeclared dependency — install instructions live in docs/performance.md. 18 new tests.

## 3. Docs corrections from the teardown
- OmniVoice guide: instruct + reference **combine** (consistent instruct stabilizes cloning — upstream's dialect-cloning guidance; the reference wins conflicts); inline pronunciation control (pinyin / CMU phonemes); prompt persistence; and corrects the false "no voice design on the default engine" claim (attribute-based design is supported and shipped).
- performance.md documents both new env knobs.
- CJK guard: `docs/engines/omnivoice.md` allowlisted for the functional pinyin example.

Not ported (already present, found during the audit): auto-transcription of missing `ref_text` (generate-time with persist-back, plus the model's own on-the-fly ASR fallback) and instruct+ref plumbing (flows end-to-end from the UI already). Sidecar prompt reuse (omnivoice-subprocess re-encodes per synthesize) noted as a follow-up.

Note: base currently includes ee35d238, so the pre-existing watermark-scan failure shows here until #1564 lands — will merge main after.

Adds persistent `VoiceClonePrompt` disk caching and opt-in CUDA FlashInfer acceleration with sanitized fallback handling, tests, and documentation updates. This reduces repeated reference processing and improves batch-one decoding performance while preserving output. Review CUDA patching, thread-affine teardown, and fail-closed sanitizer handling for compatibility and recovery risks.


debpalash:
Review round 1 addressed in 5c48a858:

- **Greptile (apply failure leaves partial patches)**: accepted — the except branch now calls `_unapply_flashinfer` so a failed apply restores the standard path completely.
- **CodeRabbit Critical + Greptile (`_CTX` concurrency)**: accepted — eager mode now gets the same single-thread pin as graph mode (`_install_compile_thread_affinity`); the per-generation plan/pos_ids state is never touched by two workers. Documented in performance.md.
- **CodeRabbit (PID-only tmp file)**: accepted — tmp names now include a uuid per write.
- **CodeRabbit (attn impl restore)**: accepted and extended — unapply restores the *captured* pre-apply implementation (could be flash_attention_2), unit-tested.
- **CodeRabbit (docs: Known limits states a feature / OOM overpromise)**: accepted — support sentence moved to Behaviour notes; performance.md now states the fused-weight VRAM cost and scopes the fallback claim to classified kernel failures.
- **CodeRabbit (weight-release after fusing)**: rejected — freeing the original q/k/v/gate/up projections would break both `_unapply_flashinfer` and the runtime fallback path, which restore the class forwards that use those weights. The honest trade-off is documenting the VRAM cost (done) and keeping the opt-in off by default.
- **CodeRabbit (docstring layout, stale test import)**: accepted.

debpalash:
Round 2 addressed in 7a361a54:

- **Greptile (fallback publishes eager path before teardown completes)**: accepted — reordered so `_unapply_flashinfer` runs first; until it returns, `_model.generate` still routes through the thread-affinity wrapper, so concurrent renders queue instead of racing the half-restored model.
- **CodeRabbit (latch reason can leak home paths)**: accepted — `mark_flashinfer_runtime_failure` now runs the reason through `core.failure.sanitize` (the existing redactor) before latching, covering the log line and every later skip re-log.
- **CodeRabbit (CPU-persistence test too weak)**: accepted — tokens created on CUDA when available, and the test asserts the *persisted payload* is CPU-resident (save-time move), not just the loaded object.

debpalash:
Round 3 addressed in 5b3947f7: the sanitize fallback no longer keeps the raw reason — on redactor failure only the exception class is latched with a fixed "details redacted" note (also resolves the CodeQL empty-except alert). Regression tests cover both the normal redaction ($HOME → ~) and the sanitizer-failure path.

Issue #1169, opened 2026-07-16
/v1/audio/speech returns unwatermarked audio (EU AI Act Art. 50(2), applies Aug 2 2026)


debpalash:
## Summary

The OpenAI-compatible endpoint `POST /v1/audio/speech` returns synthetic audio **without the AudioSeal invisible watermark**, while `/generate` watermarks the same text. The provenance mark depends on which door the audio came out of.

This matters beyond consistency: **EU AI Act Article 50(2) becomes applicable on 2 August 2026.**

## Evidence

Watermarking lives in the shared tail of `generation.py`:

```python
# backend/api/routers/generation.py:637
async def _finalize_generation(audio_tensor, sample_rate, *, text, ...):
    """Shared tail of a successful generation: watermark → save WAV →
    history row (self-healing) → retention prune → event emit.
    ...
```

`_finalize_generation` is **private to `generation.py`** — nothing outside that module calls it. `embed_watermark` is reached from exactly three places:

```
backend/api/routers/generation.py     ✅
backend/api/routers/dub_generate.py   ✅
backend/services/persona_bundle.py    ✅
backend/api/routers/openai_compat.py  ❌  ← 0 references
```

`openai_compat.py` resolves an engine (`_resolve_engine`), synthesizes, calls `_encode_audio(...)`, and returns the bytes — `embed_watermark` never enters the path. Its module docstring says it reuses "the same protocol used by the rest of OmniVoice, so engine selection, GPU offloading, [...]" — watermarking is absent from that list.

Note the default is **on**, so this is a coverage gap, not a settings issue:

```python
# backend/services/watermark.py
def is_enabled() -> bool:
    return resolve("watermark.invisible", default=True) is not False
```

There's already a comment in `generation.py` recording that this exact class of bug bit once before — "plain TTS came out unmarked despite the setting being on."

## Not yet audited

These routers may produce synthetic audio and contain no `watermark` reference. They may well route through a covered path — I did not verify:

- `tts_stream.py`
- `audiobook.py`
- `batch.py`
- `longform_jobs.py`
- `stories.py`

(`dictation.py` is ASR — audio → text — so out of scope.)

## Why the deadline is real

**Article 50(2)**, applicable **2 August 2026**:

> "Providers of AI systems, including general-purpose AI systems, generating synthetic audio, image, video or text content, shall ensure that the outputs of the AI system are marked in a machine-readable format and detectable as artificially generated or manipulated."

[...]

PR #1569, opened 2026-08-15
fix(auth): offer the API-key login on server-mode admin 403s

### LINKED ISSUE #1568: Remote browser UI spams 403 on /sysinfo, /model/status, /system/notifications and never offers the API-key login
## Environment

- Build: main (post-#1525/#1528; first visible in v0.5.0-era previews)
- Deployment: headless/Docker with `OMNIVOICE_SERVER_MODE=1` + `OMNIVOICE_API_KEY` set (the documented remote setup), UI opened from a LAN browser at `http://<host>:3900`

## What happens

The UI loads, but every status poll fails forever with 403 and no login is ever offered:

```
GET /sysinfo                → 403 {"detail":"loopback origin or admin API key required"}
GET /model/status           → 403 (same)
GET /system/notifications   → 403 (same)
```

The console fills with these (the polling hooks use `retry: Infinity` at 5s/10s/30s intervals) and the affected panels show error states. The only way in is knowing to paste the API key into Settings → Remote backend; nothing in the UI suggests it.

## Root cause

Since #1525 the whole `/system` router sits behind `require_admin`, so an unauthenticated non-loopback client gets **403** where it used to get **401**. The SPA routes only 401s into the `ov:auth-required` flow (`RemoteAuthGate` shows the API-key form), so the 403 falls through as a plain error and the auth bootstrap (master → short-lived session) is never offered.

Secondary facet: `require_admin` answers `"loopback origin or admin API key required"` **even in desktop mode**, where the credential checks only run under server mode — no presented key can ever satisfy the desktop gate. A client that routed on this message generically (as I initially did) would trap a desktop PIN-share guest in a login form that can never succeed.

## Proposed fix (PR follows)

1. `frontend/src/api/client.ts`: route a 403 whose detail contains "admin api key" into the existing `ov:auth-required` flow (`mode: 'apikey'`) — same handling as the 401; other 403s (CSRF, desktop-only) stay plain errors.
2. `backend/api/dependencies.py`: `_admin_gate_403()` emits a **mode-distinct** detail — server mode keeps "loopback origin or admin API key required" (a key genuinely satisfies it), desktop mode answers "loopback origin required" (truthful: loopback-only).
3. `tests/test_auth_gate_detail_lockstep.py`: cross-layer lock so a backend reword of either literal fails CI unless `client.ts` is updated in the same change.

Verified end-to-end on a live LAN deployment: the gate renders after 3 initial 403s (polling stops, no spam), key entry exchanges for a session, and all polls return 200.

paoloantinori:
Fixes #1568

## What

- **Frontend** (`client.ts`): a 403 whose detail names the admin API key now routes into the existing `ov:auth-required` flow (`mode: 'apikey'`), exactly like the 401 paths — `RemoteAuthGate` shows the API-key form, the app tree unmounts, and the infinite-retry polling stops instead of spamming the console. Other 403s (CSRF "browser origin rejected", desktop-only routes) deliberately stay plain errors, because presenting a key cannot satisfy them.
- **Backend** (`dependencies.py`): new `_admin_gate_403()` emits a **mode-distinct** detail. Server mode keeps `"loopback origin or admin API key required"` (a key genuinely satisfies the gate there); desktop mode now answers `"loopback origin required"` — truthful, since the admin-credential checks only run under `_server_mode()`, so no presented key can ever satisfy the desktop gate. Without this split, routing on the message alone would trap a desktop PIN-share guest in a login form that can never succeed.
- **Lockstep test** (`tests/test_auth_gate_detail_lockstep.py`): pins the literal contract across layers — the frontend matcher ("admin api key") must keep matching the backend's server-mode detail and must NOT match the details a key cannot fix. A backend reword without updating `client.ts` fails CI instead of silently regressing to 403 spam.

## Tests

- New frontend cases in `client.test.ts`: admin-gate 403 dispatches `{mode:"apikey"}`; CSRF 403 does not dispatch (fail-before/pass-after).
- New backend cases in `test_loopback_server_mode.py`: per-mode detail for `require_admin` + `require_admin_action` (desktop → plain loopback, server+key → key-naming).
- Updated the two desktop-arm assertions in `test_engines_route_shape.py` to the new truthful detail.
- Suites run on top of current `main`: backend 208 passed, frontend api 95 passed; `oxfmt --check` clean.

## Verification

On a live LAN deployment (server mode + `OMNIVOICE_API_KEY`): before, a remote browser logged `/sysinfo`, `/model/status`, `/system/notifications` 403s every few seconds indefinitely. After: 3 initial 403s, the API-key gate renders, the exchange produces a short-lived session, and all polls return 200.

An adversarial review pass caught the first version of the frontend-only fix breaking desktop LAN-share guests (unsatisfiable key form); the mode-distinct backend detail is the root-cause fix for that class.

Server-mode admin 403 responses now trigger API-key authentication when an API key is configured, while unrelated 403 responses remain errors and stale responses cannot clear newer sessions. Backend details now reflect deployment mode and credential availability, with cross-layer tests covering routing and recovery. Review the pre-existing watermark route coverage failure separately; mismatched error details could still break authentication routing.

## Review follow-ups (commit e2bfeaa0)

Greptile's review on 014e18ce flagged two P1s, both fixed with fail-before/pass-after tests:

1. **Unsatisfiable PIN-only login gate** (`dependencies.py`): server mode + share PIN + no `OMNIVOICE_API_KEY` closed read-only bootstrap while no key existed to present, so the key-naming detail gated the whole UI behind a form whose exchange can never succeed. `_admin_gate_403()` now names the key only when server mode **and** an API key is configured; PIN-only and bare servers answer the plain loopback detail (truthful: only loopback can use admin when no key exists), which the SPA leaves a plain error. New test: `test_require_admin_pin_only_server_mode_detail_is_plain_loopback`.
2. **Stale 403 clears new session** (`client.ts`): a late 403/401 landing after a key exchange wiped the fresh session (`clearAdminSession()` was unconditional), reloading a successful login straight back into the gate. A failed response may now only invalidate the credentials it actually carried (send-time session capture + token compare). New test: `a stale 403 does not clear a session stored during its flight`.

CodeRabbit's docstring-coverage warning on the new test functions is also addressed.


paoloantinori:
The red **Tests (backend + frontend)** check is pre-existing on `main`, not introduced by this PR: `tests/test_watermark_route_coverage.py::test_every_synthesis_module_routes_through_mark_synthetic` fails identically on pristine `upstream/main` (`ee35d238`) — offender `worker/transport/server.py` synthesizes audio without the `mark_synthetic` chokepoint (#1169). Everything else in the matrix is green on `e2bfeaa0`.

PR #1573, opened 2026-08-16
fix(tests): read code, not prose — AST-scan the watermark coverage guard

paoloantinori:
Fixes the red **Tests (backend + frontend)** check on `main`.

## What happened

`tests/test_watermark_route_coverage.py::test_every_synthesis_module_routes_through_mark_synthetic` scans `backend/{api,services,worker}` for modules that call a synthesis primitive and requires each to route through the `mark_synthetic` chokepoint (#1169, EU AI Act Art. 50(2)). The scan was a regex over **raw source text**, so the comment in `worker/transport/server.py` explaining the `remote_tts_render_v1` feature flag — "A generic backend.generate() call accepts the same wire shape…" — matched, and the gRPC control plane (which never synthesizes anything; it receives digest-verified artifacts) was flagged as an unmarked producer. Main went red.

## The fix: the guard reads code, not prose

The producer scan, the excusal checks, and the `embed_watermark` call ban all moved from substring/regex matching to AST inspection:

- `_synthesizes(src)` — a module is a producer only if it really CALLS a synthesis primitive (`_run_inference`, `_run_backend_inference`, `generate_with_cached_ref`, `synthesize_chapter`, or the adapter call on any `*backend` receiver). Comments and docstrings cannot trip it. An unparseable source is flagged, never excused (strict degrade).
- The adapter rule matches any receiver whose terminal name ends with `backend` (`backend.`, `self.backend.`, `_backend.`), which preserves parity with the old regex **and** closes a hole it missed: `generation.py`'s `_backend.generate(...)` now scans as the producer it is.
- `_references(src, …)` — an excusal (or the `test_known_producer_still_marks` tripwire) requires a real `Name`/`Attribute` node: direct call, dotted call, or `functools.partial` callback. A prose mention of `mark_synthetic` no longer excuses an unmarked producer.
- `_calls(src, …)` — the `embed_watermark`-outside-the-chokepoint ban scans call sites, so a comment like "do not call embed_watermark() directly" can no longer redden main (the exact sibling of the incident being fixed).

Verified over the whole scanned tree: exactly two files change classification, both prose-only (the false positive `worker/transport/server.py`, and `services/audiobook.py` whose allowlist entry rested on a docstring mention and is removed, sharpening the list). No real producer is declassified.

## Known gap documented (review finding, not fixed here)

`services/gpu_sandbox.py` renders via `model.generate(...)` — a real synthesis site no name rule matches (the old regex missed it too) — currently unwired. Documented next to the sonitranslate precedent in the guard instead of silently ignored; wiring that module in should mark the returned audio or adopt the `*backend` adapter convention.

## Review

`/simplify` (4 angles) and a high-effort code review were both run on the branch; all findings — the substring excusal, the unfixed sibling regex, and the gpu_sandbox blind spot — are addressed in 4922b590 (the blind spot as documentation, per the file's own convention). Watermark suites: 33 passed.

Replaced raw source-text scans with AST-based watermark coverage checks that detect real synthesis calls, excusals, and `embed_watermark` call sites while rejecting unparseable source. This fixes the failing backend/frontend test check and removes prose-only classifications without declassifying real producers. The existing undocumented `model.generate(...)` site in `services/gpu_sandbox.py` remains a known coverage gap; 33 watermark tests pass.

## Review follow-up (fb6e1ebf)

CodeRabbit's Major finding fixed: an `_ALLOWED` module that fails to parse is now flagged with the parse error instead of being silently skipped by its exemption (regression test verified fail-before against the old allowlist-first order, pass-after here); the staleness test parses allowlisted sources explicitly. The changelog entry moved from the unsupported `CI` category into `Fixed` with the contributor credit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,12 +26,11 @@ the frozen-backend fallback mirror it for their toolchains.
 - Every engine now has its own guide — 21 new pages under docs/engines plus an index covering all 16 TTS and 11 ASR engines, linked from both READMEs (#1556)
 
 ### Fixed
+- The watermark-coverage guard scans the AST instead of raw source text, so a comment naming a synthesis primitive (the worker transport's feature-flag rationale) can no longer turn main red — and prose mentions can no longer excuse an unmarked producer either (#1169) — thanks @paoloantinoro!
 - The crash-isolated ASR sidecar and its download preflight now agree on which model to load — setting the shared faster-whisper model variable applies to both variants instead of the sidecar quietly using a different one (#1556)
 - "Ready" now requires the deep health probe (a working database-backed route), not just the identity probe — a backend whose install broke underneath can no longer be announced up while every real request fails (#1548)
 - Supervisor restarts after repeat crashes now back off (immediate, then 5s, then 15s) instead of respawning back-to-back, so a tight crash loop can't burn the whole restart budget in seconds (#1548)
 
-### CI
-- The watermark-coverage guard scans the AST instead of raw source text, so a comment naming a synthesis primitive (the worker transport's feature-flag rationale) can no longer turn main red — and prose mentions can no longer excuse an unmarked producer either (#1169)
 
 ## [0.5.0] — 2026-08-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ the frozen-backend fallback mirror it for their toolchains.
 - "Ready" now requires the deep health probe (a working database-backed route), not just the identity probe — a backend whose install broke underneath can no longer be announced up while every real request fails (#1548)
 - Supervisor restarts after repeat crashes now back off (immediate, then 5s, then 15s) instead of respawning back-to-back, so a tight crash loop can't burn the whole restart budget in seconds (#1548)
 
+### CI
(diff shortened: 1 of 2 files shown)


paoloantinori:
Superseded by #1564 (merged) — same root cause, maintainer's tokenize-and-blank approach, which I'm happy to defer to. Closing to avoid two competing guard implementations.

For the record, four deltas from this PR's review rounds survive #1564 and may be worth a small follow-up:

1. `test_embed_watermark_not_called_outside_the_chokepoint` still greps raw source, so a prose mention of `embed_watermark(` (e.g. a "do not call this directly" comment) reddens main — the exact sibling of the incident #1564 fixes. Applying `_code_only` there closes it.
2. The producer regex still can't see `_backend.generate(...)` (real synthesis, `api/routers/generation.py`) — `\b` doesn't match inside the identifier, before or after this fix. An alias rule or an AST suffix check would catch it.
3. An allowlisted module that fails to parse passes the staleness check vacuously via the raw-scan fallback; parsing allowlisted sources explicitly would fail loudly.
4. `services/gpu_sandbox.py` renders via `model.generate(...)` with no mark — unwired today, invisible to any rule (documented in this PR's guard comment next to the sonitranslate precedent).

Happy to send a follow-up PR with any of these on top of #1564's architecture.

PR #1577, opened 2026-08-17
perf(watermark): background-prefetch the AudioSeal generator at startup

### LINKED ISSUE #1576: First synthesis serializes the AudioSeal watermark load (~42s inline); intermittent 'No module named getopt' skips watermarking
## Environment

- macOS (Apple Silicon), MPS, source run with the `omnivoice-subprocess` TTS backend
- Watermarking enabled (AudioSeal); observed on main@2d5f2e80-era and earlier

## 1) First synthesis serializes the AudioSeal load (~42s inline)

Measured cold start (model idle-unloaded, no GPU contention, cold page cache after overnight eviction):

| t | event |
|---|---|
| 0s | sidecar spawn (1s) |
| ~36s | torch import + OmniVoice weights loaded (313 + 527) |
| ~50s | PCM ready back in the parent |
| **~87s** | `AudioSeal generator loaded` — the audioseal import + generator load ran **inline inside the first `mark_synthetic`**, in the parent, after the audio was already synthesized |
| 90s | client timeout fired, 3s before completion; everything cascaded to a fallback TTS |

The watermark generator load is lazy (by design), but "lazy" here means "serialized into the first synthesis". The span is dominated by the audioseal dependency import on a cold filesystem (~42s measured; ~9s warm) — the checkpoint itself loads in 0.1s from the local torch.hub cache, no network involved (measured with `HF_HUB_OFFLINE=1` and without).

**Proposed fix (PR follows)**: warm the generator on a background thread ~35s after boot, mirroring the capture-ASR warm-up pattern (`OMNIVOICE_PRELOAD_WATERMARK=0` to opt out), on the watermark pool. First-synthesis cold start drops by the whole inline span. The lazy builds also need locks — the prefetch thread races the first embed (per-model locks, so a detector load never queues behind a generator build).

## 2) `ModuleNotFoundError: No module named 'getopt'` — watermark silently skipped

```
WARNING [omnivoice.watermark] Watermark embedding failed (passing through original): ModuleNotFoundError: No module named 'getopt'
```

Observed twice (cold + warm synthesis) in a long-lived backend process; **not reproducible in a fresh process** (embedding works; `tabulate`/`audioseal` import cleanly; the only `import getopt` in the venv is tabulate's CLI entrypoint, lazy and unreached). The failure disappeared after a backend restart, so it is state-dependent in a long-lived process. The catch-all in `embed_watermark` logged only the message, no traceback — the linked PR makes it log `exc_info`, so the next occurrence is diagnosable from the log. Recording this here so anyone who sees it again knows what to look for.

paoloantinori:
Fixes #1576

## What

The first `mark_synthetic` serialized the audioseal import + generator load (~42s measured inline on a cold filesystem) INSIDE the first synthesis, pushing a cold first synthesis to ~87s — 3s past the measured client timeout. The generator now warms on a background task ~35s after boot (+5s past the capture-ASR warm so the two cold imports don't contend for the same disk), on the watermark pool, cancellable at shutdown. `OMNIVOICE_PRELOAD_WATERMARK=0` opts out.

Details that the review rounds (in-repo `/simplify` + a high-effort code review) tightened:

- **Per-model locks** on the lazy builds: the prefetch thread races the first embed (one build per model, verified by a 4-thread test with a deliberately slow load), and a detector load no longer queues behind a ~42s generator build. `release_idle_models` takes both locks in a fixed order.
- **One extra idle window** for a prefetch-warmed, never-used generator: the reaper would otherwise drop it 15 min after boot and re-impose the cold start the PR exists to hide. Real embed/detect use clears the grace.
- **`will_mark()` gates before the pool is created**, preserving `get_watermark_pool()`'s lazy invariant (hosts with watermarking off never spawn its thread); setup-half failures of the warm-up task log immediately instead of surfacing as an unretrieved exception at shutdown.
- **`exc_info=True`** on the embed and detect catch-alls: today an intermittent `ModuleNotFoundError('getopt')` inside AudioSeal's forward (issue #1576 §2, not yet root-caused — state-dependent in a long-lived process, gone after restart) ships unmarked audio with only a one-line hint in the log.

The checkpoint itself is NOT the cost: it loads in 0.1s from the local torch.hub cache with no network (measured with and without `HF_HUB_OFFLINE=1`); the 42s is the audioseal dependency import on a cold page cache.

## Tests

`tests/test_watermark_prefetch_coldstart.py` (new): concurrent lazy-load builds exactly once; detector load not blocked by an in-flight generator build; prefetch loads when watermarking is on, no-ops when disabled/absent, degrades silently on failure; the idle-reaper grace (kept once, released after use). Watermark suites: 60 passed.

AudioSeal generator loading now runs as a cancellable background task with configurable prefetching, per-model locks, detector independence, idle-reaper grace, and traceback logging. Shutdown resets and drains the watermark pool so later operations can create a replacement pool. Review startup timing, shutdown races, and lock coordination for delayed failures.

## Review follow-up (6837ba25)

Both CI failures fixed: the `#1000` shutdown guard now pins all five task handles (regex updated — the guarded property is unchanged), and the idle-window test's full-suite flake is closed at the root — tests that boot the app lifespan could leak the deferred preload task, so `conftest` defaults `OMNIVOICE_PRELOAD_WATERMARK=0` for the session (opt-in possible) and the test is neutralized against leaked warm-ups.

Greptile P1 + CodeRabbit: shutdown now drains the watermark pool (`shutdown(wait=False, cancel_futures=True)`) so a queued warm-up can't outlive the app; a thread already inside the ~42s cold import is the same unkillable-thread reality documented for the GPU pool. The warm-up reads its own delay knob `OMNIVOICE_PRELOAD_WATERMARK_DELAY` (default 35s). `_prefetched_unused` transitions are under `_generator_lock`.

Skipped with reason: gating the prefetch on local-checkpoint presence — the warm-up downloads only what the first embed would have downloaded anyway; time-shifting that is the feature.


paoloantinori:
### Code review

Found 1 issue:

1. The shutdown drain closes the watermark pool singleton without resetting it, so a process that keeps running after a lifespan shutdown dead-submits on the next watermark operation with `RuntimeError: cannot schedule new futures after shutdown` (the CI "Tests (backend + frontend)" check is currently failing with exactly this). This is the same bug class PR #1002 fixed for the GPU pool: the fix is to clear `_watermark_pool_singleton` under its build lock before draining, so the next `get_watermark_pool()` builds a live replacement.

https://github.com/debpalash/VoiceStudio/blob/6837ba25ac5f198ad9e67401776e1d00bbb24868/backend/main.py#L1140-L1144

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

paoloantinori:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1595, opened 2026-08-20
test(desktop): isolate cleanup script artifacts

### LINKED ISSUE #1566: test_linux_process_stop_executes_before_data_wipe fails on dev machines with a desktop build present
`tests/test_desktop_prod_scripts.py::test_linux_process_stop_executes_before_data_wipe` isolates `HOME`/`XDG_DATA_HOME` into a tmp fixture but executes `scripts/desktop-prod.sh --skip-build` with `cwd=<repo root>`. Its final assertion (`returncode != 0`) relies on "no build artifact exists in this fixture" — but the launch step resolves the artifact from the **repo**, not the fixture, so on any machine where the desktop app has been built (src-tauri target present) the script exits 0 and the test fails. CI passes only because runners never have a build.

Fix direction: make the artifact lookup part of the fixture (e.g. run the script from a tmp copy with a stubbed target dir, or stub the launch binary path the same way python3/pgrep/lsof are stubbed) so the test is deterministic on dev boxes too.

Found while verifying #1564/#1565 locally — unrelated to either change.

debpalash:
## Summary

Makes the Linux desktop cleanup regression test independent of build artifacts already present in a developer checkout. Closes #1566.

## Changes

- Run `desktop-prod.sh --skip-build` from a temporary checkout fixture.
- Stub Linux platform detection so the public script seam runs on every CI host.
- Preserve the stop-before-wipe and no-artifact launch assertions.

## Type

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] ♻️ Refactor
- [ ] 📝 Documentation
- [x] 🧪 Tests
- [x] 🔧 CI / Build
- [ ] 🚀 Release prep

## Testing

- `uv run pytest -q tests/test_desktop_prod_scripts.py tests/test_changelog_style.py` (17 passed)

## Checklist

- [x] I've updated relevant documentation (if applicable)
- [x] Version files are in sync (no version bump)
- [x] Runtime behavior is unchanged; the smoke fixture is unaffected

## Release cadence

VoiceStudio ships continuous-to-main; this test-only fix is ready for the rolling preview after review and green CI.

The regression test now runs `desktop-prod.sh --skip-build` from an isolated temporary checkout with stubbed Linux detection. This prevents developer build artifacts from affecting the stop-before-wipe and no-artifact launch assertions. The change adds no public API changes; 17 tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,9 @@ the frozen-backend fallback mirror it for their toolchains.
 - The crash-isolated ASR sidecar and its download preflight now agree on which model to load — setting the shared faster-whisper model variable applies to both variants instead of the sidecar quietly using a different one (#1556)
 - "Ready" now requires the deep health probe (a working database-backed route), not just the identity probe — a backend whose install broke underneath can no longer be announced up while every real request fails (#1548)
 - Supervisor restarts after repeat crashes now back off (immediate, then 5s, then 15s) instead of respawning back-to-back, so a tight crash loop can't burn the whole restart budget in seconds (#1548)
+- The Linux desktop cleanup regression test now isolates build artifacts, so an existing developer build can no longer change its result (#1566)
 
 ### CI
-- The Linux desktop cleanup regression test now isolates build artifacts, so an existing developer build can no longer change its result (#1566)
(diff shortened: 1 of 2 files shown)


debpalash:
Landed on main via #1598 (`de5d8481`); closing this superseded source PR.

PR #1597, opened 2026-08-20
fix(sharing): ship LAN web UI in desktop installs

### LINKED ISSUE #1589: [Bug] MSI/portable LAN sharing redirects to localhost because frontend/dist is not packaged
### Before filing

- [x] I searched [existing issues](https://github.com/debpalash/VoiceStudio/issues?q=is%3Aissue) and this isn't a duplicate.
- [x] I'm on the latest release (or `main`) — older builds may already be fixed.

### What happened?

On VoiceStudio v0.5.0 using the Windows MSI/portable install, Network Sharing starts correctly on the LAN port, but remote browsers are redirected to:

http://localhost:3901

This makes the LAN sharing feature unusable from another device.

## Cause

The installed portable project contains the backend, but does not include:

`frontend/dist`

backend/main.py checks for that folder and, when it is missing, falls back to:

`RedirectResponse(url="http://localhost:3901")`

That works as a development fallback, but it is incorrect for a remote LAN client.

The desktop app itself still works because Tauri has its own bundled frontend, while the separate Uvicorn LAN listener needs frontend assets that Python can serve.

## Confirmed workaround

Building the v0.5.0 frontend and copying the generated:

`frontend/dist`

into the installed project fixes the problem.

After restarting VoiceStudio, the LAN URL returns HTTP 200 and the application loads normally from another device.

Remote API requests also correctly use the host machine for inference, so the network-sharing backend itself appears to work correctly once the frontend assets are present.

### Steps to reproduce

1. Install and launch VoiceStudio v0.5.0 using the Windows MSI/portable setup.
2. Enable Network Sharing.
3. On another device on the same LAN, open:

   http://<host-ip>:3901/

4. The remote browser is redirected to:

   http://localhost:3901

5. Check the installed project and note that `frontend/dist` is missing.

6. Build and copy `frontend/dist` into the installed project, restart VoiceStudio, and repeat step 3.

7. The same LAN URL now loads correctly and returns HTTP 200.

### What did you expect instead?

The MSI/portable installation should include, or otherwise expose, the production frontend assets required by the LAN-sharing server.

A user enabling Network Sharing should be able to open:

http://<host-ip>:3901/

from another device without manually building the frontend.

### Operating system

Windows (x64)

### How did you install it?

Desktop app (installer / AppImage)

### Version

v0.5.0

### Compute device

NVIDIA GPU (CUDA)

### Active TTS/ASR engine

_No response_

### Logs / diagnostic bundle

```text

```

## Suggested fix

Package the production frontend assets in a filesystem location accessible to the Python backend, or pass the packaged Tauri resource path to the backend and have backend/main.py serve the frontend from there.

Changing the localhost redirect alone would not solve the issue because the underlying frontend assets would still be missing.

debpalash:
Fixes #1589

## Summary
- bundle the production SPA as a filesystem resource in every desktop package
- install it beside the packaged backend on first setup, upgrades, and repair paths
- keep the documented source-free LAN sharing behavior true for MSI/portable installs

## Root cause
Tauri embedded `frontend/dist` for its own WebView, but the separate Python LAN server runs from the installed `project/backend` tree and can only serve a sibling `project/frontend/dist`. That filesystem tree was neither a bundle resource nor copied by bootstrap, so `/` fell through to the development-only `localhost:3901` redirect.

## Tests
- fail-before/pass-after bundle resource contract: `tests/test_packaged_lan_frontend.py`
- fail-before/pass-after Rust bootstrap copy/update regression
- `cargo check --lib`
- `cargo test --lib packaged_frontend_is_installed_for_the_lan_server`
- 22 targeted Python docs/changelog/version/resource tests
- full `cargo test --lib`: 139 passed; one existing parallel-sensitive setup test failed, then passed in isolation

## Type
Bug fix; Windows packaging report, implemented through the shared macOS/Windows/Linux Tauri resource/bootstrap path.

## Release cadence
Continuous to `main`; no version bump.

Desktop packages now bundle and install `frontend/dist` beside the packaged backend during setup, upgrades, and repair. This lets the Python LAN server serve the production SPA to remote devices without a source checkout or development server. Review resource-root resolution and replacement of existing frontend files for packaging-path or bootstrap failure risks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ the frozen-backend fallback mirror it for their toolchains.
 - The OmniVoice guide now covers combining style attributes with a reference clip (consistent instruct stabilizes cloning; the reference wins conflicts), inline pronunciation control (pinyin / CMU phonemes), and corrects the claim that the default engine can't do voice design — it can, from attributes (#1565)
 
 ### Fixed
+- Network Sharing from Windows MSI/portable installs now serves the bundled web interface to LAN devices instead of redirecting them to their own `localhost` (#1589) — thanks @TWIISTED-STUDIOS!
 - A remote browser hitting an API-key-configured server's admin 403 now gets the API-key login form instead of endless console 403s, while desktop and PIN-only/no-key servers keep the plain loopback error so guests are never offered a login no key can satisfy (#1568) — thanks @paoloantinori!
 - The crash-isolated ASR sidecar and its download preflight now agree on which model to load — setting the shared faster-whisper model variable applies to both variants instead of the sidecar quietly using a different one (#1556)
 - "Ready" now requires the deep health probe (a working database-backed route), not just the identity probe — a backend whose install broke underneath can no longer be announced up while every real request fails (#1548)
--- a/docs/sharing.md
+++ b/docs/sharing.md
@@ -16,6 +16,10 @@ For another device on the same network — e.g. opening the web UI on your phone
 
 You can also drive this from **Settings → Sharing & Remote Access**.
 
+Desktop installers include the web interface used by the LAN address; another
+device does not need VoiceStudio installed and the host does not need a source
+checkout or a separate frontend development server.
+
 ### How the PIN works
 - A fresh 6-digit PIN is generated each time you enable sharing; it is never written to disk.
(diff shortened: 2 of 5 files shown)


debpalash:
@coderabbitai review

debpalash:
Landed on main via #1598 (`de5d8481`); closing this superseded source PR.

PR #1614, opened 2026-08-20
fix(openai-compat): reuse cached engine instances in _resolve_engine

paoloantinori:
## What

`_resolve_engine` in `/v1/audio/speech` constructed a fresh backend instance per request (`return cls()`) whenever the client passes an explicit engine ID as `model`. For SubprocessBackend engines (pockettts, omnivoice-subprocess, ...) each request therefore:

- spawned a new sidecar process,
- paid a full torch import and engine model reload (~28s floor per request for pockettts, measured on an M3 Pro LAN deploy),
- registered one more `SubprocessBackend` atexit hook (the leak `get_engine_instance`'s docstring explicitly warns about),
- and left the discarded sidecar alive until the 300s idle reaper collected it.

The fix routes the explicit-ID path through the existing cached-singleton seam, `get_engine_instance_for` — the same reuse the `tts-1`/`tts-1-hd` path already gets via the active-engine instance.

## Behavior preserved

- Unknown model IDs → same 400 with the actionable list.
- Unavailable engines → same 400 naming the availability message.
- `tts-1`/`tts-1-hd` and the `OmniVoiceBackend` → active-engine special case untouched.
- Engine idle unload / reaper lifecycle unchanged (the singleton participates in it like any other).

## Testing

- New `tests/test_openai_speech_engine_cache.py`: two resolves of the same explicit ID return the same instance with exactly one construction; unknown IDs still 400.
- Neighbor speech-route suites pass (22 passed across engine-cache + binary-guard-1172 + load-budget-1033 + pool-queue-1190).
- End-to-end on the LAN deploy follows in the task notes: warm pockettts requests drop from ~28s to single-digit seconds.

`/v1/audio/speech` now reuses cached engine instances for explicit engine IDs and unloads the outgoing engine when the ID changes. This prevents repeated subprocess creation, Torch imports, model loads, and `atexit` registrations. Tests verify reuse, unloading, and unchanged 400 responses for unknown IDs.


paoloantinori:
Valid finding — fixed in 13c14e2c: explicit-ID switches now mirror get_active_tts_backend's MM2-01 rule (unload the outgoing engine first, best-effort), with a regression test. OmniVoiceBackend explicit requests also unload the outgoing explicit engine before resolving to the active one.

debpalash:
Reviewed — the diagnosis and the fix are both right, and the measured ~28s → single-digit is a real win. Two notes, one of which is me retracting a concern:

**Not a problem (checked):** I went looking for a race on the new module-global `_explicit_engine` — two requests with different explicit IDs, one calling `_unload_explicit_engine()` while the other holds the instance. It isn't reachable today: `create_speech` is `async def` and `_resolve_engine` contains no `await`, so the check-and-set runs atomically with respect to other coroutines on the loop, and line 312 is its only call site.

That safety is invisible in the code, though, and it's one refactor deep — dropping `async` (which sends the route to the threadpool) or adding an `await` inside `_resolve_engine` would silently introduce the race. Could you add a line to the `#:` comment saying the dict is unsynchronized *because* the only caller is a single-threaded async route? Cheap insurance for whoever touches it next.

**Blocking:** `CHANGELOG.md` needs an `### Fixed` entry under `[Unreleased]` — the repo treats keeping that section current as a hard rule, and this is squarely user-facing (a warm pockettts request going from ~28s to single-digit seconds is the headline). One line, with `(#1614)` and your credit. Style is quiet/scannable one-liners; the existing entries are the template.

Happy to land it as soon as the changelog entry is in and CI reports.

debpalash:
Both items resolved — the shared-seam restructure is better than the comment I asked for (no router-local state left to reason about), and the changelog entry is in. One maintainer commit on top: your entry ran 415 chars against the 400-char one-liner limit `tests/test_changelog_style.py` enforces, so I trimmed it in place rather than bounce CI. Merging when the gates report.

PR #1624, opened 2026-08-21
fix(dub): keep each speaker's words on merge/split, add per-line subtitle editing

### LINKED ISSUE #1612: [Feature] A leap forward in subtitle management
### Before filing

- [x] I searched [existing issues](https://github.com/debpalash/VoiceStudio/issues?q=is%3Aissue) and [discussions](https://github.com/debpalash/VoiceStudio/discussions) for this idea.

### What problem does this solve?

I'm writing this tip because, as a passionate subtitler and dubbing artist, I think it would make the program a game-changer.

Without naming names, the world's best free subtitle software is a step forward in managing individual sentences.

Let me explain: in that program, you can:

- Adjust the duration of individual sentences
- Adjust the start timing of individual sentences
- Delete individual sentences
- Add a new sentence
- Merge individual sentences with the preceding or following sentences (which I believe VoiceStudio already does)

Well, it would be fantastic to be able to work with subtitles this way in VoiceStudio, too.

This is because, if you want perfect dubbing, the most important thing is the subtitles: if the subtitles are bad, then the dubbing won't be perfect either.

An example of the need for these changes:

Character A is speaking, but a couple of words end up in the next sentence spoken by character B. In this case, moving those few words that ended up in the next line back to the previous line causes problems. First, the seconds of that sentence may not be enough to accommodate the additional words. Second, when cloning the voice, incorrect timing that doesn't match the video could affect the choice of voice timbre. This way, some words might end up with the wrong character's speech.

All these problems would be addressed with this new, more comprehensive approach to the more meticulous management of subtitles.

At present, however, yes, some words can be changed, but the timing (which is crucial) cannot be changed at all.

Thank you for your attention.

### Proposed solution

Enable the following features:

- Adjust the seconds of a single phrase's duration
- Adjust the start timing of a single phrase
- Delete a single phrase
- Add a new phrase
- Merge single phrases with preceding or following phrases (which I believe VoiceStudio already does)

### Alternatives considered

_No response_

### Area

Video dubbing

MhAhmadAli:
Closes #1612.

## The bug

Moving a few stray words across a speaker boundary has one path in the dub table: merge the two lines, then split them again at the right word. That round-trip **lost the second speaker**.

`segmentMerge` built the merged row from `...a` and kept only `b.end`, dropping b's `speaker_id` / `profile_id` / `direction` / `gain` / `target_lang`; `segmentSplit` then spread the merged row's fields onto *both* halves. The second character's words came out dubbed in the first character's voice — exactly what the reporter predicted without being able to name the cause.

## The fix

`merge_parts` records, on the merged row, which attribution covers which stretch of **text**. A split restores the attribution covering the split offset, so words carried across a boundary keep the voice of whoever actually speaks them.

**Text offsets, not timestamps.** Timestamps were the first attempt and are wrong: nothing stops a user retiming a line so it overlaps its neighbour, and after merging, two parts then cover the same instant — a time-keyed lookup returns whichever it visits first and hands the words to the wrong speaker. The split point the user chooses *is* a text position, and text ranges cannot overlap by construction.

All five per-row identity fields are carried, not just `speaker_id` — fixing one would leave the same bug wearing a different field's name. Segments that were never merged carry no record and behave exactly as before.

`merge_parts` is client-side bookkeeping: `DubSegment` ignores unknown fields and `segmentGenInputs` does not hash it, so neither the request contract nor regen fingerprints change.

## Also in the report

Three of the five asks already shipped (start-time editing, delete, merge-with-next). The rest:

- **Duration** — the end time is now editable alongside the start. Both fields commit through the same move/resize path the timeline drag handles use, including the speed compensation. The numeric start previously wrote `start` raw and skipped it, so the two UIs disagreed about what changing a time means.
- **Merge with previous** — `Ctrl/Cmd+Shift+M`, or the row's ⋯ menu.
- **Insert a line** — takes the silent gap before the next cue and continues the same speaker, voice and language, so a phrase the transcript missed no longer needs a re-transcribe.

## Verification

Driven end to end against the running app, not only unit tests:

```
── row 1 merged into row 0 (Merge with previous) ──
  0  0:00.5  0:05.8  5.3s  Anna  Hello there, I was just about to head out. But wait, you forgot your coat.

── split back apart at the word boundary ──
  0  0:00.5  0:03.6  3.1s  Anna  Hello there, I was just about to head out.
  1  0:03.6  0:05.8  2.2s  Ben   But wait, you forgot your coat.
```

That run is what caught the timestamp-keyed first draft, which passed 38 unit tests and still failed in the app because every fixture used tidy non-overlapping ranges. There is now a regression test that retimes a line into its neighbour before merging, and the unit fixtures deliberately overlap in time.

- 53 targeted tests (23 unit, 18 hook, 12 row), fail-before/pass-after
- `backend/tests/` 288 passed; typecheck, oxlint (0 errors), oxfmt clean; node:test 71/71
- New strings translated into all 21 locales (the ratchet in `test_locale_parity.py` has zero headroom); keyboard cheatsheet updated for the new shortcut
- `CHANGELOG.md` under `[Unreleased]`

The full `tests/` suite was still running locally when this was opened, on hardware sharing CPU with other work; CI is the authority. Changes are frontend-only plus CHANGELOG and locale files.

## Not addressed here

`detectOverlaps` is wired only into `SegmentTrack.jsx`, never into the segment table — so an overlap from any source (drag, hand-edited times, insert into a gapless transcript) is invisible in the table. Pre-existing and broader than this issue; flagged rather than fixed.

Adds subtitle timing edits, silent-gap insertion, and previous/next-line merging for dubbing workflows. Preserves five attribution fields when merged text is later split and clears stale metadata after text or language changes. Review text-offset attribution and overlapping or boundary timing cases.


MhAhmadAli:
Review findings triaged. Both Greptile P1s were **already fixed by @debpalash** in `79342376` before I got to them — noting that so the thread isn't double-counted:

- **Filtered merges target hidden rows** → `visibleMergeAvailability(segments, visibleSegments, segment)` requires the displayed neighbour to *be* the source neighbour. I'd independently written a visible-id-set variant; theirs is cleaner and I dropped mine rather than duplicate it.
- **Translation retains stale merge offsets** → `merge_parts: undefined` at both text-replacement sites (`useDubWorkflow` translate result, `dubSlice` language swap).

CodeRabbit's three (`parseTime` strictness, source-order merge bounds, untrimmed `merge_parts` offsets) were addressed in `588f2e2c`. I verified the trim fix in particular: `trimmedTextRange` derives offsets from the slice that survives the trim, which is the drift CodeRabbit described.

## One thing I added on top (`7542b72f`)

Clearing `merge_parts` at each call site is correct, but it has to be *remembered* at every site that replaces text — and it was missed at both existing ones, which is what Greptile caught. So `partsFor` now validates the record against the line it is attached to:

```js
export function describesText(parts, text) {
  if (!parts || !parts.length) return false;
  return parts[parts.length - 1].textEnd === (text || '').length;
}
```

A valid record always ends exactly where its text does — an invariant both `mergedParts` and `clipParts` maintain. A record that no longer matches is describing replaced text, so it's discarded in favour of the row's own attribution. This is a backstop for the *next* path that replaces text, not a replacement for the explicit clearing.

Two fixtures carried records that didn't span their text; they violated the invariant and are corrected.

@debpalash happy to drop this commit if you'd rather keep the per-site clearing as the single mechanism — it's additive, not a correction to your fix.

## Verification

287 files / 2327 tests pass, plus your `DubSegmentTableMerge` and per-language cases. typecheck, oxlint (0 errors) and oxfmt clean.

Issue #1635, opened 2026-08-22
[Crash] Backend died (exit code 1)


Xohaibxobi:
## Environment

**Version:** `0.5.0`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36 Edg/151.0.0.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 4070 (12 GB VRAM)`
**CPU:** `Intel64 Family 6 Model 151 Stepping 2, GenuineIntel`
**RAM:** `31.7 GB`
**Disk free:** `20.8 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 0 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-08-22T10:13:10.000Z (27 min ago)
**Exit:** `exit code 1`
**Uptime before crash:** 13 s
**Backend version:** `0.5.0`

```
… (truncated)
use.
2026-08-22 15:13:07,302 INFO [omnivoice.model] Preload skipped: k2-fsa/OmniVoice is not installed locally — the model will load (and download if requested) on first use.
2026-08-22 15:13:07,303 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager started
2026-08-22 15:13:07,303 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager started2026-08-22 15:13:07,303 INFO [omnivoice.api] MCP server mounted at /mcp

2026-08-22 15:13:07,303 INFO [omnivoice.api] MCP server mounted at /mcp
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:3900 (Press CTRL+C to quit)
INFO:     Application startup complete.
ERROR:    [Errno 10048] error while attempting to bind on address ('127.0.0.1', 3900): only one usage of each socket address (protocol/network address/port) is normally permitted
INFO:     Waiting for application shutdown.
2026-08-22 15:13:07,904 INFO [omnivoice.api] Shutdown: cleaning up…
2026-08-22 15:13:07,904 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager shutting down
[...]

PR #1657, opened 2026-08-24
fix(core): macOS fallback for the os.waitid ownership probe (#1656)

### LINKED ISSUE #1656: [Bug][macOS] main after #1635: 'module os has no attribute waitid' — TTS sidecar management broken on every request after the first
## Describe the bug

On current `main` (a8371baa, #1635), every macOS deployment breaks with:

```
ERROR [omnivoice.openai_compat] OpenAI TTS failed: module 'os' has no attribute 'waitid'
ERROR [omnivoice.subprocess_backend] sidecar idle reaper error
```

The first synthesis after a start succeeds; subsequent requests (and the sidecar idle reaper) fail. Reproduced on a macOS arm64 LAN server (source checkout, launchd): first `/v1/audio/speech` 200 in 46s (cold load), second request 500 in 0.04s with the error above.

## Root cause

`backend/core/contained_subprocess.py` (added in #1635) calls `os.waitid(os.P_PID, pid, WEXITED | WNOHANG | WNOWAIT)` in `_posix_exited_unreaped()` and `_signal_owned_group()`.

**CPython on macOS does not expose `os.waitid`** — it is compiled in only where `HAVE_WAITID` is set, and CPython's macOS builds do not set it:

```
$ python3 -c 'import os; print(hasattr(os, "waitid"))'
False        # Homebrew Python 3.11.16, darwin
False        # /usr/bin/python3
```

So the code path raises `AttributeError` on all macOS builds. Linux is unaffected (`waitid` exists there), and CI's macOS smoke jobs evidently don't exercise the poll/reaper path (it triggers on the second spawn/probe, not the first).

This also violates the platform default-parity rule: the desktop/backend lifecycle feature shipped in default mode is broken on a whole platform.

## Repro

1. macOS (arm64 or Intel), main at or after a8371baa.
2. `POST /v1/audio/speech` twice with any subprocess engine (e.g. `model: pockettts`).
3. First request 200, second request 500 `module 'os' has no attribute 'waitid'`; `sidecar idle reaper error` appears in the log within minutes.

## Fix angles

- Gate the WNOWAIT probe: where `os.waitid` is missing, fall back to a macOS-safe equivalent. The probe's contract is "has the direct child exited but not been reaped"; possible implementations: a `ctypes` call to `waitid(2)` from libSystem (Darwin does have the syscall; CPython just doesn't wrap it), or a restructure that uses `waitpid(WNOHANG)` in a context where reaping by this owner is acceptable.
- Whichever way, an `hasattr(os, 'waitid')` platform test would have caught this pre-merge; worth adding alongside the fix.

## Workaround used locally

We reverted a8371baa on our deployment branch; consecutive requests work again (200s, 6.7s / 3.3s warm).

paoloantinori:
Fixes #1656.

## What

#1635's `OwnedPopen` probes "exited but not yet reaped" with `os.waitid(P_PID, …, WEXITED|WNOHANG| WNOWAIT)`. **CPython on macOS does not expose `os.waitid` at all** (`HAVE_WAITID` is not set in its builds), so on macOS every poll after the first spawn raised `AttributeError`: second+ synthesis requests 500 with `module 'os' has no attribute waitid'`, and the sidecar idle reaper errors. Repro in the issue: first request 200, second request 500 in 0.04s.

This PR keeps the waitid path on Linux (byte-identical behavior) and adds a macOS fallback:

- `_posix_exited_reaping()`: `waitpid(WNOHANG)` probe that *reaps* the wrapper — returns the real exit code when exited, `None` while running, and refuses (`ECHILD` → no signal) for foreign reapers. A leader reaped through **our own** Popen reports its known code instead of refusing (own-handle reaping is not the foreign-reaper case the refusal guards).
- The lost pre-reap group kill is covered by the supervisor's control-pipe EOF, which per the #1635 design already terminates the nested group.
- `_signal_owned_group` tolerates the missing attribute (the `_returncode` checks in terminate/kill plus `ProcessLookupError` bound the window).

## Why not ctypes on Darwin's waitid(2)

I tried it before falling back: Darwin's `waitid` with `WNOHANG|WNOWAIT` **reports running children as eligible** with junk siginfo (verified empirically: running child → `si_code=CLD_EXITED, si_status=0`), so the Linux "`si_pid != 0` means waited" reading is wrong there and the probe cannot be implemented over it. `sysctl KERN_PROC_PID` was the other candidate; its `p_stat` offset is not stable enough to ship. The reaping probe is the honest approximation; if you want exact WNOWAIT semantics on Darwin it likely needs the Rust side.

## Testing

- New `backend/tests/test_contained_subprocess_waitid_fallback.py` (5 tests) simulates the macOS platform (`monkeypatch.delattr(os, 'waitid', raising=False)` — on Linux CI it deletes the real attr; on macOS it is already absent) and pins: running child polls `None`, real exit codes surface (exit 3 → 3), `kill()`+`wait()` work, own-handle reaping reports the code, foreign reaping is refused.
- Full upstream `test_contained_subprocess.py` still passes (7 tests) and `test_sidecar_install.py` (36) passes.
- End-to-end on a macOS arm64 LAN server running this branch: 3 consecutive `/v1/audio/speech` pockettts requests → 200 at 27.8s (cold) / 5.9s / 5.0s, zero waitid errors; pre-fix the second request 500'd.

Adds a macOS fallback for `OwnedPopen` when CPython lacks `os.waitid`, using `waitpid(WNOHANG)` to preserve ownership and exit-code handling. This fixes repeated TTS failures and sidecar reaper errors on macOS while preserving Linux behavior. Review the residual PID-reuse risk during group signaling after foreign reaping.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ the frozen-backend fallback mirror it for their toolchains.
 - The OmniVoice guide now covers combining style attributes with a reference clip (consistent instruct stabilizes cloning; the reference wins conflicts), inline pronunciation control (pinyin / CMU phonemes), and corrects the claim that the default engine can't do voice design — it can, from attributes (#1565)
 
 ### Fixed
+- macOS no longer loses TTS after the first request when Python lacks `os.waitid`; subprocess ownership now uses a safe `waitpid` fallback without risking reused process groups (#1656) — thanks @paoloantinori!
 - Desktop startup, Retry, reset, uninstall, shutdown, and crash recovery now share one backend lifecycle owner; quitting interrupts first-run installers and gracefully drains then force-cleans the full backend process tree, so overlaps cannot duplicate or orphan it (#1635) — thanks @Xohaibxobi!
 - Large Stories and Audiobook projects now persist in IndexedDB instead of overflowing the `omnivoice.app` localStorage envelope, with quota-safe migration and orderly exit/reload flushing (#1636) — thanks @leodzai!
 - OmniVoice and its crash-isolated subprocess now route to AMD ROCm GPUs instead of warning and falling back to CPU (#1629) — thanks @j4r3kb!
--- a/backend/core/contained_subprocess.py
+++ b/backend/core/contained_subprocess.py
@@ -163,10 +163,16 @@ class OwnedPopen:
             return
         except AttributeError:
             # macOS CPython has no os.waitid (#1656): the ECHILD refusal is
-            # approximated by acting only while this handle still owns an
-            # unreaped leader (callers check _returncode first), and
-            # ProcessLookupError below covers the already-exited group.
-            pass
+            # approximated by refusing once the leader is fully gone (kill(0)
+            # ESRCH means reaped, so the pid must not be signalled), and
+            # ProcessLookupError below covers a group that exited meanwhile.
+            # kill(0) on a *reused* pid still succeeds — that residual window
+            # is inherent to the platform (no WNOWAIT equivalent); callers
+            # bound it by checking _returncode before signalling.
+            try:
+                os.kill(self.pid, 0)
+            except ProcessLookupError:
+                return
         try:
             os.killpg(self.pid, sig)
         except ProcessLookupError:
--- a/backend/core/contained_subprocess.py
+++ b/backend/core/contained_subprocess.py
@@ -162,16 +162,16 @@ class OwnedPopen:
         except ChildProcessError:
             return
         except AttributeError:
-            # macOS CPython has no os.waitid (#1656): the ECHILD refusal is
-            # approximated by refusing once the leader is fully gone (kill(0)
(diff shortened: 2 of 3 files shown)


paoloantinori:
Two follow-ups:

- CodeRabbit's ownership-verification finding: fixed in the new head — the no-waitid path of `_signal_owned_group` now refuses via `kill(pid, 0)` once the leader is fully gone (ESRCH = reaped), with a test covering the foreign-reap case. The residual reused-pid window is documented as inherent (no WNOWAIT equivalent on the platform).
- The Windows smoke failure (`test_upload_write_does_not_block_revocation_or_publish_after_it`, 0.44s against a 0.2s timing budget) looks like runner slowness in a subsystem this PR doesn't touch — the diff is POSIX-only and that test exercises the Windows Job path unchanged. I can't rerun checks on this repo; the push above re-triggers CI, so we'll see if it reproduces.

PR #1769, opened 2026-09-02
feat(dub): realtime dub preview

mvanhorn:
## Summary

ROADMAP's "Real-time dub preview (stream TTS as you edit)" (feature-magic track, After Phase 4.1): with an opt-in toggle (default **off**) on the dub segment table, editing a segment's translated text debounces ~400 ms and streams that one line over the **existing** `/ws/tts` WebSocket (binary PCM16 sentence chunks) with the segment's CAST voice — you hear the edit without pressing Generate. The next keystroke or a segment switch closes the previous socket, so sockets never pile up.

Walkthrough

*Walkthrough recorded against a protocol-faithful mock `/ws/tts` (paced PCM16 tone frames — the recording box has no GPU): toggle on → typing streams the line (pulsing row speaker + player bar), typing again restarts the stream, clicking the speaker stops it, and a voiceless line gets the actionable CAST toast instead of a socket.*

This deliberately builds on what already ships: the `/ws/tts` protocol is untouched, incremental re-dub and `preview: true` (`num_step=8`) pipeline previews are untouched, and no engine work is involved. `services.rvc.apply_rvc` is not called anywhere on this path.

## Changes

- **`hooks/useDubLivePreview.js`** (new): debounce → abort previous socket → stream, with an intent counter cancelling obsolete sessions. Every stream runs inside the same process-wide `withTtsInflight` admission `/generate` and Convert hold (busy → the standard localized "generation already running" toast; the backend side already serializes on the GPU pool via `run_on_gpu_pool_guarded`). Segment with no CAST voice → no socket, actionable "assign a CAST voice" toast. Voice resolution goes through `segmentGenInputs` — the same `preset:` → instruct expansion the dub generate body uses, so preview and Generate can never disagree.
- **Playback**: binary WS frames feed the existing Web Audio streaming chunk player (`createStreamingChunkPlayer` gains `appendPcm16Bytes`; base64 decode refactored onto the same `pcm16BytesToFloat32` core). Identical default behavior on macOS/Windows/Linux — plain AudioContext, no MediaSource.
- **Ear-only**: nothing is persisted as job audio. Export still runs the full-quality generate path (`finalizeTtsBeforeExport`), and existing dub generate/export tests stay green.
[...]

PR #1768, opened 2026-09-02
feat(batch): watch-folder auto-ingest

mvanhorn:
## Summary

Adds an **opt-in Watch folder** to the Batch queue: pick a directory once, and every new video file that lands in it is auto-enqueued through the existing `enqueueBatchJob` / `POST /batch/enqueue` File upload — no Add-to-queue interaction needed. Fully local, no new outbound calls, and **filesystem paths never ride an HTTP request**: the backend only ever receives uploaded bytes, exactly as it does today.

Walkthrough

*(Walkthrough recorded against a stubbed `/batch/*` backend — the same seams the tests mock — in a Chromium browser via the File System Access path; the mp4 is dropped into the watched folder externally mid-recording and auto-enqueues ≈ 5 s later.)*

## Changes

**Frontend**

- New `components/WatchFolderBar.jsx` (extracted from BatchQueue, which stays focused on the job list): idle "Watch folder" button (double-click-guarded while the picker is open) → active chip with folder name, live auto-added count, Pause/Resume, and Stop. The poll loop (~5s), dedup, and stop-on-unmount live here; the enqueue itself stays with BatchQueue.
- New `utils/watchFolder.js`: pure helpers + two watch backends.
  - Dedup identity is **name+size+mtime** (`entryKey`), with a two-scan stability guard (`createIngestTracker`): a file is only released once two consecutive scans agree on its key, so a large mp4 still being copied in is never uploaded half-written. Pre-existing folder contents are primed as seen — only *new* arrivals enqueue. Non-video files are ignored. Transient read/enqueue failures hand the entry back (`retry`) for a later poll; pauses hand it back untouched (`unsee`).
  - **Tauri backend** (all three OSes): no fs plugin ships in this repo, so the folder is picked natively in the Rust process and registered under a random session token; scans and byte reads resolve that token over IPC (poll ≈5s per the no-watch-API fallback). The webview never sends a directory path back, and nothing path-shaped ever reaches the Python backend.
  - **Web**: browsers with the File System Access API (Chromium) poll a `showDirectoryPicker()` handle directly; elsewhere the button no-ops with an actionable message pointing at Add Videos (web-only limitation, desktop behaviour is identical on macOS/Windows/Linux).
[...]

PR #1767, opened 2026-09-02
feat(dub): project-level drag casting board

mvanhorn:
## Summary

The ROADMAP feature-magic item **"Project-level casting view (drag voices to speakers)"**: the dub editor's CAST strip gains an expandable **Casting Board** — speaker rows (with the auto-clone chip when the extractor found a usable passage) and draggable voice chips (Default, clone profiles, design presets). Dropping a chip on a speaker writes **exactly the fields the CAST `<select>` already writes** (`profile_id` on matching segments plus `merge_parts`/`merge_parts_original` attribution), so the dropdowns stay in sync live and job persistence is unchanged — no new DB table, no new endpoint. The dropdowns are NOT replaced; both views sit over the same store fields.

Walkthrough

*The GIF is the real `CastingBoard` component (the exact code in this diff) with the app's real stylesheet and i18n, driven by native OS-level input over a store-stub harness — the same seams the co-located tests use. It shows the toggle expanding the board, a native HTML5 drag of "Jake — deep announcer" onto SPEAKER_2 (row + top-strip dropdown update together, toast confirms), then the keyboard path on SPEAKER_3: Enter opens the listbox, arrows walk the options, Enter casts "Mia — bright host".*

## Changes

- **`frontend/src/utils/segments.js`**: the CAST select's inline mutation extracted as `assignSpeakerProfile(segments, speakerId, profileId)` (extract-over-copy — the select and the board now share it verbatim), plus `castSpeakers()` for the deduped speaker list including speakers that only survive inside merged rows' parts.
- **`frontend/src/components/dub/CastingBoard.jsx`** (new, exactly at the 300-line soft cap): the pre-existing per-speaker `<select>` strip moved verbatim from `DubLeftColumn` (unchanged markup — existing CAST tests pass untouched), plus the board: HTML5 DnD (chips carry the profile id via `dataTransfer`; the Default chip travels as a sentinel since `''` isn't distinguishable from "no data") **and** a keyboard path — Enter on a speaker row opens a `role="listbox"` with roving focus (arrows/Home/End/Enter/Escape).
- **`frontend/src/components/dub/DubLeftColumn.jsx`**: shrinks 880 → 791 lines (the file was over the hard cap; this PR reduces it and does not grow it).
- **`frontend/src/index.css`**: new rules extend the existing `.dub-cast-*` cluster only — no new CSS file.
[...]

PR #1766, opened 2026-09-02
feat(audiobook): synced-lyrics player

mvanhorn:
## Summary

The audiobook result panel's bare `<audio controls>` becomes a **synced-lyrics player**: the chapter text renders under the native controls and the word under `audio.currentTime` is highlighted karaoke-style, so a listener can follow the narration (and click any word to jump there). Fully local — no new outbound calls, no ASR pass, no backend changes at all; the Tauri-safe `downloadMedia` path (#1218) is untouched.

Walkthrough

*(Walkthrough recorded against a stubbed TTS engine — the same seam the tests mock — so the flow runs without GPU/model downloads: Create → ready → play with the word sweep crossing the chapter boundary → click-to-seek on “hallway” → Download hover.)*

## Changes

**Timing (the interesting part)**

- No word timestamps exist at synth time — TTS engines don't emit them, and `synthesize_chapter`'s outputs pass through two content-addressed cache layers plus a remote-render lane where any per-word sidecar would go stale or missing. But the render stream **already persists what the player needs**: every `chapter` SSE event carries `duration_s` (the same numbers that build the m4b's FFMETADATA chapters). The tab now keeps them.
- Within each chapter, words are **even-split over `[start, end]`** — a direct port of the karaoke burn-in's old-job fallback (`karaoke_ass.even_split_words`, #1764).
- After a reload (the output filename survives in the store; stream events don't), the whole book even-splits over the audio element's own duration, chapter spans falling out proportionally by word count.

**Frontend**

- New pure helper `frontend/src/utils/audiobookLyrics.js` (183 lines): `evenSplitWords` (mirrors `backend/services/karaoke_ass.py`), `scriptChapters` (mirrors `longform_parser.py`'s H1 split + chapter drop rules so cue indices line up with the stream's chapter list — intro-before-first-heading, pause-only chapters kept, control-token-only chapters dropped), `buildLyricsTimeline` (skips failed chapters — they're absent from the muxed audio — and degrades to the proportional split on any script/stream drift), `activeWordIndex` (binary search; a pause keeps the previous word lit, matching the karaoke sweep).
[...]

PR #1775, opened 2026-09-02
fix(i18n): ja "Cleaning…" is denoising, not housekeeping

j30231:
## Summary

`clone.cleaning` is the status shown on the mic button while the backend denoises a
just-recorded reference clip. `ja.json` translated "Cleaning…" as 掃除中… — *sweeping a
room* — which is the housekeeping sense of "clean", not the audio one.

The backend is unambiguous about what the string covers
(`backend/api/routers/system.py`):

```python
@router.post("/clean-audio")
async def clean_audio(audio: UploadFile = File(...)):
    """Accept a raw mic recording, run demucs vocal isolation, return clean WAV."""
```

`ja.json` already uses ノイズ除去 for the sibling key `recording.loaded_raw`
("Recording loaded (raw — denoising unavailable)"), so the two strings describing the
same step contradicted each other. This aligns them.

## Changes

- `frontend/src/i18n/locales/ja.json` — `clone.cleaning`: 掃除中… → ノイズ除去中…

## Type

- [x] 🐛 Bug fix

## Testing

- `pytest tests/test_locale_parity.py` → 248 passed
- No keys added or removed, so the missing-key ratchet is untouched.

## Checklist

- [ ] I've updated relevant documentation (if applicable)
- [ ] Version files are in sync (if version bump)
- [x] If this PR changes runtime behavior, the regression fixture still loads green — string-only change, no runtime behavior touched

---

The same string is wrong in several other locales the same way (`ko` 청소 중…,
`zh-CN` 清理中…, `de` Reinigung…, `fr` Nettoyage…, `es` Limpieza…). ko is fixed in a
separate PR; I left the ones I can't judge to native speakers.

The Japanese `clone.cleaning` translation now reads `ノイズ除去中…` to describe audio denoising. The changelog credits this fix in the Unreleased section. Locale parity, CJK, and changelog style tests passed, so no merge risk was identified.


debpalash:
Reviewed: the change is correct (ノイズ除去中 is the denoising sense the `clone.cleaning` status means). I merged current `main` into the branch and added the Unreleased changelog credit line; locale parity, CJK, and changelog style tests pass locally. Merging once the CI run is approved and green. Thanks @j30231!

debpalash:
I moved the entry back to the 0.5.2 section — sorry for the ping-pong, here is the reasoning so it is not just my preference against yours.

`v0.5.2` is **prepared but not tagged**: the latest release is v0.5.1 (`gh release list`, and there is no `v0.5.2` tag on the remote). The `## [0.5.2] — 2026-09-02` section is therefore the upcoming release, and everything merging right now ships in it — which is why #1765, #1766, #1767, #1768 and #1769 all recorded their entries there. `## [Unreleased]` stays empty until 0.5.2 is actually cut, otherwise the release body that `release.yml` extracts for the tag would omit your fix.

CLAUDE.md's "keep Unreleased current" is the right instinct in the general case, so your read was reasonable — it just does not apply while a version section is open and untagged. Also dropped the Highlights line: the one-liner under Fixed matches how the other small locale and copy fixes in 0.5.2 are recorded.

No change to your translation. Merging once CI is green.

PR #1789, opened 2026-09-03
fix(desktop): resolve exports authorization dir from the running backend

### LINKED ISSUE #1781: [Bug] 403 Forbidden: Invalid or expired desktop authorization
## Error

```
403 Forbidden: Invalid or expired desktop authorization
ApiError: 403 Forbidden: Invalid or expired desktop authorization
    at apiFetch (http://localhost:3901/src/api/client.ts:496:10)
    at async apiJson (http://localhost:3901/src/api/client.ts:505:14)
    at async handleNativeExport (http://localhost:3901/src/App.jsx:741:4)
```

## Environment

**Version:** `0.5.2`
**OS:** `Windows-10-10.0.19045-SP0`
**Python:** `3.11.16`
**Compute device:** `cpu`
**CPU:** `Intel64 Family 6 Model 165 Stepping 2, GenuineIntel`
**RAM:** `15.6 GB`
**Disk free:** `5.8 GB`
**Active TTS engine:** `voxcpm2`

## Backend reachability

**Last backend response:** 1 s before this report

## Recent actions

```
16:33:31 view:studio
16:35:56 view:voice
16:36:15 view:studio
16:36:23 view:voice
16:39:54 view:studio
16:42:47 generate:start (audio)
16:53:43 export
17:00:12 export
17:00:50 export
```

debpalash:
## Root cause

Exports (and every other native-picker capability — `models_dir`, `ffmpeg`, `ffprobe`, `soni_input`, `soni_output_dir`) could 403 with `Invalid or expired desktop authorization`. Tauri writes a one-shot capability file into the data dir **it** resolves (`commands::path_authorization_dir` → `setup::resolved_data_dir` → configured data_dir / portable base / `setup::default_data_dir`), but `backend/core/path_authorization.py::consume` scans `<its own DATA_DIR>/.path-authorizations`, where `DATA_DIR` comes from `backend/core/config.py::get_app_data_dir()` (`OMNIVOICE_DATA_DIR` or a per-OS default). When the two disagree, the token is never found and every export 403s.

This happens whenever the backend is spawned out-of-process without `OMNIVOICE_*` env — e.g. dev mode, where the UI is served from Vite (`localhost:3901`) and the backend is started by `scripts/dev-backend.mjs`, which strips `OMNIVOICE_*` from the child env (`scripts/desktop-common.mjs`) — or in a packaged build when a custom data folder or portable mode is applied after the backend already started.

Confirmed against the root-cause analysis already posted on #1781.

## Fix

- `commands::path_authorization_dir` (frontend/src-tauri/src/commands.rs) now asks the **running backend** where its data dir actually is, via a new `backend::backend_data_dir` (frontend/src-tauri/src/backend.rs) that reads `data_dir` from `GET /system/info`, falling back to Tauri's own resolution (the historical behavior) when the backend is unreachable (not started yet, old backend predating the field, or a non-VoiceStudio responder). This is a single resolver shared by every `authorize_host_path` capability kind, so the fix covers the whole class in one place — not just `dub_export`.
- `core.path_authorization.consume` (backend/core/path_authorization.py) now distinguishes, server-side, a **missing capability store** (the data-dir mismatch symptom — directory doesn't exist at all) from a **missing token** (an ordinary expired/already-consumed capability — directory exists, token isn't in it). Both still return a generic, path-free 403 over HTTP (no filesystem disclosure, no widened `_KINDS`, no webview-supplied path ever trusted), but the mismatch case now logs a diagnosable server-side warning instead of looking byte-identical to a stale token.

## Security

- The capability file is still unguessable, one-shot, `0600`, written only by the native Tauri process, and never resolved from anything the webview supplies.
- The backend still never accepts a filesystem path over HTTP — `backend_data_dir` only ever *reads* `data_dir` from the backend's own `/system/info`, matching the existing `backend_healthy`/`running_backend_version` probe pattern already in `backend.rs`.
- `_KINDS` is unchanged (six kinds, same as before).
- Client-facing error messages stay generic; the new diagnostic detail is server-log only.

## Tests

- Rust (`cd frontend/src-tauri && cargo test --lib`): 209 passed, 0 failed — including 4 new tests: `backend::tests::backend_data_dir_reads_system_info_and_degrades_safely`, `commands::host_path_authorization_tests::prefers_the_running_backends_advertised_data_dir`, `commands::host_path_authorization_tests::falls_back_to_tauris_own_resolution_when_the_backend_is_unreachable`.
- Python (`.venv/bin/python -m pytest tests/ -q`): 6562 passed, 31 skipped, 8 xfailed, 1 xpassed, 0 failed — new `tests/test_path_authorization.py` (14 tests) covers `consume()` across all six capability kinds plus the fail-before/pass-after regression for the store-vs-token error distinction (verified failing against the pre-fix code, passing after).
- Frontend (`bun run test`): 2552 passed (305 files), 0 failed — unaffected by this change, run for completeness per repo convention.

## Docs / Changelog

- No documented behavior changed (nothing in `README.md`/`docs/**` describes this internal IPC mechanism), so no docs-sync update needed.
- One-liner added under `### Fixed` in the `## [0.5.2]` CHANGELOG section (v0.5.2 is prepared but not yet tagged).

Fixes #1781.

https://claude.ai/code/session_01HR6J9zKQop9TGGVUwjypnF

Tauri now uses the backend’s absolute `data_dir` for desktop authorization files, with safe fallback resolution, fixing export and native-picker `403` errors caused by directory mismatches. Backend logs distinguish a missing authorization store from an invalid token while preserving generic HTTP errors and security checks. Review the backend path trust boundary and cross-platform Rust tests.


debpalash:
Reviewed and pushed one hardening commit on top.

**Finding:** `backend_data_dir` accepted any non-empty `data_dir` string. But `get_app_data_dir()` returns `OMNIVOICE_DATA_DIR` verbatim, so the advertised value can legitimately be **relative** — a source or Docker setup pointing at `omnivoice_data` is enough. Tauri would then join that string onto its own working directory while the backend resolved it against its own, silently recreating the exact split this PR closes. Same bug class, new trigger.

**Fix:** require `Path::is_absolute()`; a relative advertisement is treated as unusable and the caller falls back to Tauri's own resolution. That also removes any say a foreign responder squatting the port would have over where capability files get written, which matters because this is the first place Tauri takes a filesystem path from an HTTP response.

Added two regression cases (`omnivoice_data`, `./data`). `cargo test --lib`: 209 passed.

I also verified the rest independently: the fallback chain preserves historical behaviour when the backend is unreachable, `_KINDS` is untouched, the 403 stays generic and path-free over HTTP while the store-vs-token distinction is logged server-side only, and all six capability kinds route through the one resolver.

debpalash:
Found the Windows failure, and I need to correct something I said earlier: it was **not** my `is_absolute()` commit. The Windows leg was already red on the first pushed head (122679a9), before that commit existed.

**Root cause.** The new unit tests obtained an `AppHandle` by building a real Tauri app with `tauri::test::mock_builder`. On the Windows runner that aborts the whole test binary — it exits before the harness prints one line, and cargo reports only "test exited abnormally", no assertion and no panic message, which is why the log looked empty. It had no precedent here because the crate's existing MockRuntime harness lives in `tests/backend_lifecycle.rs`, an **integration** test that this CI step's `cargo test --lib` never runs. Putting one in a lib test was new.

**Fix.** Split the decision into `authorization_dir_from(advertised, fallback)` and test that directly. Same coverage of the branch #1781 turns on, with no app construction, no `OMNIVOICE_PORT` mutation, and no loopback stub — so it also runs on every platform instead of needing platform-conditional fixtures. `path_authorization_dir` is now two argument expressions with no logic of its own, and the HTTP side keeps its own tests in `backend.rs`.

`cargo test --lib`: 210 passed. The `is_absolute()` filter and the escape decoder both stay.

Issue #1774, opened 2026-09-02
[Bug] Generation exceeded the compute-time limit. The backend is still running; try a


DSpinelli01:
## Error

```
Generation exceeded the compute-time limit. The backend is still running; try a shorter passage or raise the generation timeout.
StreamingPreviewError: Generation exceeded the compute-time limit. The backend is still running; try a shorter passage or raise the generation timeout.
    at f (http://tauri.localhost/assets/main-app-Ccm2oqM6.js:16:31792)
    at qa (http://tauri.localhost/assets/main-app-Ccm2oqM6.js:17:54)
    at async o (http://tauri.localhost/assets/generate-DmpB6q_w.js:1:263)
    at async http://tauri.localhost/assets/main-app-Ccm2oqM6.js:17:5019
```

## Environment

**Version:** `0.5.1`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36 Edg/151.0.0.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cpu`
**CPU:** `Intel64 Family 6 Model 165 Stepping 3, GenuineIntel`
**RAM:** `15.9 GB`
**Disk free:** `218.5 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 1 s before this report

## Recent actions

```
16:44:14 view:launchpad
17:11:00 view:studio
17:13:07 generate:start (design)
17:23:08 generate:stream-retryable-abort
```

Harsh4r0ra:
Looks like the client-side timeout is shorter than backend CPU inference time for the `omnivoice` engine — the error explicitly says the backend is still running when the abort fires. Recent actions show `generate:start` → abort after ~10 minutes.

Possible fixes: (1) expose/raise the generation timeout setting in the desktop app rather than hardcoding it, (2) chunk long passages automatically before sending to CPU-bound engines, (3) surface compute device (CPU vs GPU) more prominently since CPU inference on `omnivoice` seems to hit this a lot.

Happy to test a build with a configurable timeout if that's the direction.

Harsh4r0ra:
[...]

Issue #1778, opened 2026-09-03
[Bug] Generation exceeded the compute-time limit. The backend is still running; try a


praveen-teqbahn:
## Error

```
Generation exceeded the compute-time limit. The backend is still running; try a shorter passage or raise the generation timeout.
handleEvent@http://localhost:3901/src/utils/streamingTts.js:356:36
_streamGenerateSpeech@http://localhost:3901/src/utils/streamingTts.js:371:26
```

## Environment

**Version:** `0.5.2`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.6.2-arm64-arm-64bit`
**Python:** `3.11.14`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M3`
**RAM:** `16 GB`
**Disk free:** `98.6 GB`
**Active TTS engine:** `omnivoice`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 2 s before this report

## Recent actions

```
09:12:47 view:launchpad
09:17:01 view:studio
09:18:27 generate:start (audio)
09:26:38 generate:stream-retryable-abort
```

debpalash:
This is the guarded compute-time budget doing its job, not a crash: the backend abandons a job after `OMNIVOICE_GENERATE_TIMEOUT_S` (300 s on GPU/MPS, 600 s on CPU) plus 1 s per 40 characters past the first 1200 (see docs/performance.md). Here the abort fired after roughly 8 minutes, on an Apple M3 (MPS) with the omnivoice engine.

To tell whether that is expected for the input or a stall, please add: the passage length in characters, the voice mode, and the backend log lines between `generate:start` and the abort (Settings → Logs → Backend) — a stall shows no progress lines before the timeout.

Maintainer note: the message says "raise the generation timeout", but there is no Settings control for it; only the env var. Related report: #1774.

debpalash:
Tracked as #1787 now. The abort after roughly 8 minutes on your M3 was the compute budget expiring, not a crash — but the message told you to raise a timeout that has no control in the app, which is the actual defect. #1787 covers exposing it in Settings, fixing the copy, and warning about expected duration up front.

Closing in favour of #1787 so the fix is tracked in one place.

PR #1797, opened 2026-09-03
fix(settings): expose the generation compute-time budget in Settings

### LINKED ISSUE #1787: Generation-timeout error tells users to raise a setting that has no UI, and the knob is env-var-only
## Problem

When a synthesis exceeds its compute budget the user sees:

> Generation exceeded the compute-time limit. The backend is still running; try a shorter passage or raise the generation timeout.

There is no way to raise the generation timeout from the app. The only control is the `OMNIVOICE_GENERATE_TIMEOUT_S` environment variable (`backend/services/model_manager.py`), which a desktop-installer user has no obvious way to set — and on Windows the documented advice is explicitly *not* to use `setx`. So the message names a remedy the UI does not offer.

This is generating duplicate reports rather than self-service fixes: #1774 (10 minutes, CPU) and #1778 (8 minutes, Apple M3 / MPS) are both the budget working as designed, filed because the suggested workaround was undiscoverable.

## Fix

Three parts, all needed:

1. **Expose the budget in Settings.** There is already a mechanism for exactly this: `PERSISTENT_KEYS` in `backend/api/routers/system.py` plus `POST /system/set-env`, which persists to `prefs.json` and restores at startup, with the existing Settings env panel as the surface. Add `OMNIVOICE_GENERATE_TIMEOUT_S` and `OMNIVOICE_CPU_GENERATE_TIMEOUT_S`, validated as positive numbers the way `_PORT_KEYS` validates ports.

   **Implementation note:** both budgets are captured at import time (`GPU_JOB_TIMEOUT_S` / `CPU_JOB_TIMEOUT_S` at module level, plus `_CONFIGURED_GPU_JOB_TIMEOUT_S` which `generate_timeout_s` compares against to detect an explicit override). Setting the env var at runtime will not take effect until the values are read per-dispatch or the backend restarts. Pick one and make it real; do not ship a control that silently does nothing until restart without saying so.

2. **Fix the copy.** `backend/core/public_errors.py`'s `generation_timeout` detail must not advise an action the user cannot take. Once part 1 lands it can point at the Settings control; until then it should name the concrete remedies that exist (shorter passage, a faster engine, GPU if available).

3. **Say how long this will take before it starts.** The budget already scales with input length and compute device, and `generatePreflight.js` already mirrors the backend's length rule, so the app knows enough to warn up front — the CPU-vs-GPU expectation is what reporters keep missing. Surfacing progress or an estimate during generation also lets a user tell "slow on CPU" from "wedged" without filing a report, which is what @Harsh4r0ra asked for in #1774.

## Acceptance

- A user can change the generation budget from Settings, and the change actually affects the next generation (or the UI states plainly that a restart is required).
- The timeout error never recommends an action with no path in the UI.
- Regression tests: the env key is accepted and validated, a non-numeric value is refused, and the effective budget reflects a changed value.

Reported via #1774 and #1778; closing those in favour of this.

debpalash:
## Summary

Fixes #1787 (closes duplicates #1774, #1778).

When a synthesis exceeded its compute budget, the error told users to "raise the generation timeout" — a remedy with no UI path; the only control was `OMNIVOICE_GENERATE_TIMEOUT_S`, env-var-only, with no in-app way to set it (and Windows guidance explicitly avoids `setx`).

- **Settings → Performance & Device → Compute-time budget**: new panel with two rows, `OMNIVOICE_GENERATE_TIMEOUT_S` (accelerated/GPU-family hosts) and `OMNIVOICE_CPU_GENERATE_TIMEOUT_S` (CPU hosts). Validated server-side in `/system/set-env` the same way `_PORT_KEYS` validates ports (reject non-numeric, ≤0, or above a 6h ceiling — generous enough for a legitimate audiobook-length CPU render, bounded enough that a fat-fingered extra digit can't hide a wedge for days). Persisted through the existing `PERSISTENT_KEYS` → `prefs.json` path.
- **Implementation-note decision**: both budgets are captured at **import time** in `backend/services/model_manager.py` (`GPU_JOB_TIMEOUT_S`, `CPU_JOB_TIMEOUT_S`, `_CONFIGURED_GPU_JOB_TIMEOUT_S`). Rather than rewire that to a per-dispatch env read (risking the `universal_override` CPU-downgrade-guard semantics and ~9 existing tests that `monkeypatch.setattr` those module globals directly), I kept the import-time capture and made the restart requirement explicit and honest: every row in the new panel carries the existing `RestartBadge` (the same "Applies now" vs "Restart required" affordance every other env-backed Settings row uses), and `prefs.json` restoration already runs (`env_prefs` startup step) *before* `services.model_manager` is first imported (`ml_imports` step), so a saved value **does** take effect — on the next restart. This is one of the two options the issue explicitly sanctions ("...or keep the import-time capture and have the UI state plainly that a restart is required").
- **Copy fix**: `backend/core/public_errors.py`'s `generation_timeout` detail now points at the new Settings control instead of naming an unreachable action.
- **Upfront expectation**: the existing CPU long-text preflight hint (`generatePreflight.js`, from #1302) now also fires for **MPS** (Apple Silicon) hosts — `generate_timeout_s()` gives MPS the GPU-family budget, not the CPU one, so a long render there can legitimately exceed it too. This is exactly the #1778 report class, which the CPU-only check didn't cover.

## Localization

9 new keys (8 `settings.*` for the new panel, 1 `engines.mpsLongText`) land translated — not English copies — in all 21 locale files, verified by `tests/test_locale_parity.py`.

## Docs

`docs/performance.md` (env-var table gains the CPU row + Settings pointer; "warnings before a slow generation" table gains the MPS row) and `docs/install/troubleshooting.md` (Settings path noted alongside the env vars) updated in this PR per the docs-sync rule.

## Test plan

- [x] `tests/test_generate_timeout_settings_1787.py` (new, 14 tests): `/system/set-env` validation (non-numeric/zero/negative/over-max rejected, valid accepted + persisted to `prefs.json`), effective-budget-after-restart behaviour (module reload picks up the persisted env var, both GPU and CPU keys), `/system/info` surfaces both fields, and the fixed copy no longer names an unreachable action.
- [x] `frontend/src/components/settings/GenerateBudgetPanel.test.jsx` (new, 6 tests) + extended `generatePreflight.test.js` (4 new MPS tests).
- [x] Full suites green: `pytest tests/ -q` → 6576 passed (baseline 6562 + 14 new), `bun run test` → 2593 passed (baseline 2582 + 11 new), `bun run format:check` clean.
- [x] Existing `tests/test_generate_timeout_730.py` (device-aware guidance / override semantics) untouched and still passing — the import-time capture and `universal_override` branch are unchanged.

https://claude.ai/code/session_01HR6J9zKQop9TGGVUwjypnF

Adds Settings controls for GPU and CPU generation compute-time budgets with validation, persistence, restart guidance, external-override warnings, translations, and regression tests. Updates timeout guidance and long-text preflight warnings for CPU and MPS devices. Values apply after backend restart, so verify the restart flow and effective-budget reporting.


debpalash:
Addressed both CodeRabbit/Greptile P1 findings:

**1. Independent rows (took option 1 — genuinely independent, not a UI relabel).** An explicit `OMNIVOICE_CPU_GENERATE_TIMEOUT_S` now always wins for CPU dispatches in `generate_timeout_s()`, even when `OMNIVOICE_GENERATE_TIMEOUT_S` is also explicit (new `cpu_explicit` check, mirroring the existing `GPU_JOB_TIMEOUT_S != _CONFIGURED_GPU_JOB_TIMEOUT_S` pattern). Setting *only* the GPU var keeps its historical "universal override" behavior byte-for-byte (existing test unchanged, still asserts the old value) — the only behavior change is the previously-broken "both set" combination, which was never a documented/relied-upon feature. `test_worker_deadlines.py::test_universal_timeout_override_wins_on_cpu_worker` was pinning the old broken behavior and has been corrected + split into two tests. The CPU row's Settings note now says the accelerated budget is still the fallback when the CPU row is untouched, so the remaining legacy coupling is never silently assumed.

**2. External-override honesty.** `core.prefs` gains `restore_env()`/`is_env_shadowed()`, snapshotting which keys were already present in the environment (shell, `.env`, Docker, …) before our own `setdefault()`-based restore runs — a reliable predictor for every future restart too, since Settings never writes to a shell profile or `.env` file. `/system/info` and `/system/set-env`'s response both carry `*_shadowed` flags; a shadowed row shows an "Overridden externally" badge instead of "Restart required", and a save into it shows a distinct warning toast instead of a plain success.

6 new backend tests (`tests/test_generate_timeout_settings_1787.py`, `tests/test_generate_timeout_730.py`, `tests/test_worker_deadlines.py`) + 2 new frontend tests all fail on the pre-fix code. Full suites green: 6587 backend passed (was 6576), 2595 frontend passed (was 2593), `format:check` clean.

PR #1830, opened 2026-09-06
fix(moss_tts_v15): select device via torch.accelerator instead of CUDA hardcode

li-lizhe:
## Problem

The MOSS-TTS-v1.5 engine hardcodes device selection to:

```python
device = "cuda" if torch.cuda.is_available() else "cpu"
dtype = torch.bfloat16 if device == "cuda" else torch.float32
```

On an Ascend NPU (torch_npu) host, `torch.cuda.is_available()` is **False**, so the entire model silently falls back to CPU inference in fp32 — never using the NPU accelerator — even though the accelerator is available and bf16 is supported.

The same bug affects any non-CUDA accelerator: Intel XPU, AMD ROCm, etc.

## Root cause

`torch.cuda.is_available()` is CUDA-specific and does not detect other accelerators. The correct device-agnostic API is `torch.accelerator.current_accelerator()` which returns the active backend regardless of type.

## Fix

Replace the CUDA-or-CPU hardcode with `torch.accelerator.current_accelerator().type`:

- Use `torch.accelerator.current_accelerator()` to detect the active device (CUDA / NPU / XPU / MPS / CPU)
- MPS is still excluded (MOSS is untested on Apple Silicon per the comment)
- dtype is bf16 for any GPU-class accelerator, fp32 on CPU

## Verification

Verified on Ascend 910B (torch 2.14, torch_npu, 4 NPU):

|          | Before (CUDA hardcode) | After (accelerator API) |
|----------|----------------------|------------------------|
| device   | `"cpu"` (wrong)      | `"npu"` (correct)      |
| dtype    | `float32` (slow)     | `bfloat16` (fast)      |
| CUDA_avail | `False`            | `False`                |

No changes to the MPS exclusion or the existing bf16-on-CPU-safe logic.

MOSS-TTS-v1.5 and device routing now support CUDA, ROCm, XPU, and registered NPU accelerators, with CPU fallback for unavailable or failed probes and MPS. Accelerators use bfloat16, while CPU uses float32; older PyTorch environments without `torch.accelerator` remain supported. Verify optional driver failures and non-CUDA backends.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
+- MOSS accelerator routing now matches runtime device selection, including native XPU and registered NPU detection (#1830) — thanks @li-lizhe!
+
 
 ## [0.5.2] — 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
-- MOSS accelerator routing now matches runtime device selection, including native XPU and registered NPU detection (#1830) — thanks @li-lizhe!
+- MOSS accelerator routing and status match runtime selection, with CPU fallback when device probing fails (#1830) — thanks @li-lizhe!
 
 
 ## [0.5.2] — 2026-09-02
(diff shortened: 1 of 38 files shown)


li-lizhe:
Addressed the review feedback (greptile P1 + coderabbit Stability).

`torch.accelerator.current_accelerator()` can return `None` on CPU-only PyTorch builds (no accelerator compiled in), so `accel.type` would crash model loading. I now call `current_accelerator(check_available=True)` and fall back to `"cpu"` when it returns `None`, preserving the bf16-on-accelerator / fp32-on-CPU behavior.

Verified on Ascend 910B (torch 2.14, torch_npu): `current_accelerator(check_available=True)` returns `npu`, `.type == "npu"`, dtype resolves to bf16 as intended.

PR #1831, opened 2026-09-06
fix(tts-engines): select device via torch.accelerator in confucius4 and dots_tts

li-lizhe:
## Problem

Both Confucius4 and DOTS-TTS engine sidecars hardcode device selection via
`torch.cuda.is_available()`:

- `confucius4/main.py`: `device = "cuda" if torch.cuda.is_available() else "cpu"`
- `dots_tts/main.py`: `default_precision = "bfloat16" if torch.cuda.is_available() else "float32"`

On an Ascend NPU (torch_npu) host, `torch.cuda.is_available()` is **False**,
so the Confucius4 model runs on CPU (ignoring NPU) and DOTS-TTS defaults to
fp32 instead of bf16.

The same bug affects any non-CUDA accelerator: Intel XPU, AMD ROCm, etc.

## Root cause

`torch.cuda.is_available()` is CUDA-specific and does not detect NPU, XPU,
or other backends. The correct device-agnostic API is
`torch.accelerator.current_accelerator()`.

## Fix

- **confucius4/main.py**: Use `torch.accelerator.current_accelerator().type`
  for device selection. MPS is excluded (upstream untested on Apple Silicon).
- **dots_tts/main.py**: Use the same accelerator API for precision selection.
  bf16 on any GPU-class accelerator, fp32 on CPU.

## Verification

Verified on Ascend 910B (torch 2.14, torch_npu, 4 NPU):

| Engine       | Before                     | After                        |
|-------------|----------------------------|------------------------------|
| confucius4  | device="cpu" (wrong)       | device="npu" (correct)       |
| dots_tts    | precision="float32" (slow) | precision="bfloat16" (fast)  |
| cuda_avail  | False                      | False                        |

Confucius4 now detects CUDA, ROCm, XPU, and NPU accelerators with CPU fallback when detection fails or no accelerator is available. DOTS-TTS uses `bfloat16` on CUDA, NPU, and XPU, and `float32` on CPU and MPS, while preserving explicit precision overrides. Verify runtime support for non-CUDA `bfloat16` before deployment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
+- Confucius accelerator routing matches device selection, and dots.tts keeps safe CPU precision on non-CUDA accelerators (#1831) — thanks @li-lizhe!
+
 
 ## [0.5.2] — 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
-- Confucius accelerator routing matches device selection, and dots.tts keeps safe CPU precision on non-CUDA accelerators (#1831) — thanks @li-lizhe!
+- Confucius accelerator routing tolerates failed device probes, and dots.tts keeps safe default precision on non-CUDA hosts (#1831) — thanks @li-lizhe!
 
 
 ## [0.5.2] — 2026-09-02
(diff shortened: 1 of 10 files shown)


li-lizhe:
Addressed the review feedback (greptile P1 + coderabbit Functional Correctness).

In both `confucius4` and `dots_tts`, `torch.accelerator.current_accelerator()` can return `None` on CPU-only builds, so `.type` would crash. Both now call `current_accelerator(check_available=True)` and fall back to `"cpu"` when `None`.

For `dots_tts`, I additionally exclude MPS from bf16 (DotsTtsRuntime is untested on MPS), selecting fp32 there, and keep bf16 for CUDA/NPU/XPU.

Verified on Ascend 910B (torch 2.14, torch_npu): `current_accelerator(check_available=True)` returns `npu`, device resolves to npu / dtype bf16.

Issue #1860, opened 2026-09-06
[macOS] Studio header kicker text renders underneath the traffic-light window controls


psiberfunk:
## Environment

- Version: 0.5.2-150 (Desktop Release run #150)
- Exact source commit: `53ff367c1fdde695f17673707cd46e2b27d41546`
- Host: macOS 26.6.2, Apple Silicon arm64, 48 GB RAM, device=mps, managed Python 3.11.15 backend

## What happens

The studio window's header renders its kicker text (e.g. `STUDIO`) at the very left edge of the window, underneath the macOS close/minimize/zoom traffic-light buttons. When the window is active, the saturated traffic lights fully cover the text; when the window is backgrounded, the dimmed traffic lights let it show through partially legible — behavior consistent with the text sitting *under*, not beside, the controls.

## Root cause

`frontend/src-tauri/tauri.conf.json:25-26`:

```json
"decorations": false,
"titleBarStyle": "Overlay",
```

On macOS this makes the OS overlay the native traffic-light controls on top of the web content, rather than reserving a titlebar strip — so the page itself is responsible for leaving roughly 78px of left inset for that zone.

`frontend/src/components/Header.jsx:261` opens the header's left block with no macOS-specific left padding:

```jsx
<div className="flex items-center gap-[14px] justify-self-start min-w-0">
```

The pulsing status dot and the uppercase kicker `t(view.kickerKey)` (lines 264-269) both start at `x=0` inside that block, so on a sufficiently wide window they render directly under the traffic-light zone.

The codebase already knows this zone exists and handles it correctly elsewhere: `frontend/src/App.jsx:1318-1320` has a dedicated invisible drag strip commented as matching "the macOS traffic-light zone" for the wizard screens. The studio header (`Header.jsx`) just doesn't apply the same inset to its own content.

Note the kicker carries `max-[1501px]:hidden` (`Header.jsx:267`), so this only shows on wide windows where the kicker renders at all — which is why it isn't universally visible/reported.

## What I expected

The header's left block should reserve the traffic-light width (macOS-only; other platforms use `decorations` normally and don't need this) so the kicker never renders underneath native window chrome.

## Suggested fix

Add a platform-conditional left inset (roughly matching the drag-strip's traffic-light allowance already used in `App.jsx:1318-1320`) to `Header.jsx`'s left block on macOS, rather than hiding the kicker — the text should sit beside the traffic lights, not disappear.

PR #1865, opened 2026-09-07
fix(header): hide the custom window controls on macOS

### LINKED ISSUE #1864: [macOS] Both native traffic lights AND custom Windows-style min/max/close buttons render in the header
## Environment

- Host: macOS (Apple Silicon and Intel — not platform-version-specific)
- Affects the packaged desktop app (Tauri)

## What happens

On macOS, the app header renders its own Windows-style minimize/maximize/close buttons on the right side of the title bar, *in addition to* the native macOS traffic-light cluster (red/yellow/green) on the left. Both sets of controls are visible and both work, which looks sloppy and duplicated — a user has two ways to close the window, styled for two different operating systems, on screen at once.

Only the native traffic lights should appear on macOS. Only the custom minimize/maximize/close buttons should appear on Windows/Linux.

## Root cause

`frontend/src/components/Header.jsx:153`:

```js
const showWindowControls = typeof window !== 'undefined' && '__TAURI_INTERNALS__' in window;
```

This only checks "are we running inside the Tauri runtime" — true on every OS build, macOS included. It never checks *which* OS is running, so the custom control row (added in #1481, "use the app header as titlebar") renders unconditionally whenever the app is packaged, duplicating the native traffic lights that macOS already draws via `frontend/src-tauri/tauri.conf.json`'s `"decorations": false` + `"titleBarStyle": "Overlay"`.

Windows/Linux need the custom row because `decorations: false` there just removes all chrome — there's nothing native to duplicate. macOS is the only platform where this flag was never added.

## Suggested fix

Gate `showWindowControls` on platform, reusing the existing `navigator.platform` macOS-detection convention already used in `frontend/src/components/settings/HotkeyTab.jsx` and `SettingsSearch.jsx`:

```js
const isMacLike =
  typeof navigator !== 'undefined' && /Mac|iPad|iPhone|iPod/.test(navigator.platform || '');
const showWindowControls =
  typeof window !== 'undefined' && '__TAURI_INTERNALS__' in window && !isMacLike;
```

This is unrelated to #1860 (kicker text rendering under the traffic lights) — that issue is about layout/inset on macOS; this one is about the custom Windows-style controls rendering at all on macOS.

psiberfunk:
Fixes #1864.

## Problem

On macOS, the header rendered its own Windows-style minimize/maximize/close buttons *in addition to* the native traffic-light cluster the OS already draws. `frontend/src-tauri/tauri.conf.json` sets `decorations:false` + `titleBarStyle:"Overlay"` on every platform — on macOS this still overlays the native traffic lights on the web content; on Windows/Linux it draws nothing, which is why the custom row exists at all (added in #1481).

`Header.jsx`'s `showWindowControls` only checked `'__TAURI_INTERNALS__' in window` — "are we in Tauri" — true on every OS, so the custom row rendered unconditionally and duplicated macOS's native controls.

## Fix

Added an `isMacLike` check (same `navigator.platform` regex already used in `HotkeyTab.jsx` / `SettingsSearch.jsx`) and gated `showWindowControls` on `!isMacLike`. Windows/Linux behavior is unchanged.

## Testing

- `frontend/src/test/HeaderNavStyle.test.jsx`: made the existing "native window controls" test explicit about platform (`Win32`), and added a new regression test asserting the custom controls are absent on macOS (`MacIntel`). Verified fail-before (assertion fails against the unpatched `showWindowControls`) / pass-after.
- `cd frontend && bun run test` → 309 files / 2609 tests passed (full frontend suite, no regressions).
- `uv run pytest tests/test_locale_parity.py -q` → 248 passed (no new user-facing strings).
- `bun run lint` → no new warnings in `Header.jsx` or the test file.

## What a reviewer should check

- This touches the same `showWindowControls`/`isMacLike` neighborhood as #1863 (a different, already-open fix for #1860 — kicker text rendering *under* the traffic lights on macOS). The two are independent bugs in adjacent code; if both land, whichever merges second will need a small rebase to reconcile the two `isMacLike` declarations. Left #1863 untouched per repo policy.

The header now hides custom window controls on macOS to prevent duplicate controls beside native traffic lights. Windows and Linux retain custom controls. Review the `navigator.platform` detection and expanded macOS Tauri configuration for compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
+- On macOS, the header no longer shows Windows-style minimize/maximize/close buttons alongside the native traffic lights (#1864)
+
 
 ## [0.5.2] — 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
-- On macOS, the header no longer shows Windows-style minimize/maximize/close buttons alongside the native traffic lights (#1864)
+- On macOS, the header no longer shows Windows-style minimize/maximize/close buttons alongside the native traffic lights (#1864) — thanks @psiberfunk!
 
 
 ## [0.5.2] — 2026-09-02
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
+- The capture widget can hide after recording and recover from being left visible while idle (#1865) — thanks @psiberfunk!
+
 - macOS retains the shared desktop window sizing, resize limits, and file-drop behavior when native chrome is applied (#1865) — thanks @psiberfunk!
 
 - On macOS, the header no longer shows Windows-style minimize/maximize/close buttons alongside the native traffic lights (#1865) — thanks @psiberfunk!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,9 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
-- On macOS, the header no longer shows Windows-style minimize/maximize/close buttons alongside the native traffic lights (#1864) — thanks @psiberfunk!
+- macOS retains the shared desktop window sizing, resize limits, and file-drop behavior when native chrome is applied (#1865) — thanks @psiberfunk!
+
+- On macOS, the header no longer shows Windows-style minimize/maximize/close buttons alongside the native traffic lights (#1865) — thanks @psiberfunk!
 
 
 ## [0.5.2] — 2026-09-02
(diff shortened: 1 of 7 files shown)


psiberfunk:
Addressed the review feedback:

- **Greptile — missing contributor credit**: fixed, `CHANGELOG.md` now credits `@psiberfunk`.
- **CodeRabbit — Cross-Platform Default Parity warning**: judged not applicable, left as-is. `showWindowControls` differing by OS is the intended fix, not a divergence: the underlying feature (minimize/maximize/close) is present and default-on for every platform, only the *chrome* differs — native traffic lights on macOS (drawn by the OS itself via `titleBarStyle: "Overlay"`), custom buttons on Windows/Linux (which get no OS chrome at all under `decorations: false`). This is the same "platform-specific implementation for OS-level chrome, identical user-visible capability" pattern CLAUDE.md already carves out (e.g. the GPU-backend clarification) — every desktop app presents OS-native window controls, so uniform *appearance* across platforms would itself be the parity violation. No settings toggle is warranted since there's no platform where the capability is missing or opt-in.
- **CodeRabbit — Docstring Coverage warning**: noise for this diff (a two-line conditional in JSX); no docstring convention exists in this codebase for that, so left alone.

PR #1916, opened 2026-09-08
fix(setup): isolate start-truncated paths from the bidi algorithm

### LINKED ISSUE #1848: [setup] System-check details are bidi-reordered by an RTL left-truncation hack
## Environment

- Version: 0.5.2-150 (Desktop Release run #150)
- Exact source commit: `53ff367c1fdde695f17673707cd46e2b27d41546`
- Host: macOS 26.6.2, Apple Silicon arm64, 48 GB RAM, device=mps, managed Python 3.11.15 backend
- Installed from the official run-150 asset `VoiceStudio_0.5.2-150_aarch64.dmg` (sha256 `355a218c81c84ee51b178e33d3a85395b46bb760b225193a54cf2dd3553d50cb`) into `/Applications/VoiceStudio.app`, after a full verified uninstall/clean of all prior VoiceStudio state (clean-room first run).

Related: #1845 (blank always-on-top dictation pill), filed from the same clean-room run.

## What happens

On onboarding Step 1 "System check", every preflight detail string whose first character is a digit or `/` renders with its leading run moved to the visual end of the line:

| Backend emits | Screen shows |
|---|---|
| `48.0 GB total` | `GB total 48.0` |
| `219.5 GB free at /Users/aless/.cache/huggingface` | `GB free at /Users/aless/.cache/huggingface 219.5` |
| `/Users/aless/.cache/huggingface` | `Users/aless/.cache/huggingface/` |

Details beginning with a letter render correctly on the same screen (`Python 3.11.15`, `macOS-26.6.2 (arm64)`, `Reachable (63 ms)`, `omnivoice -> mps (accelerated)`) — that split (letter-first vs. digit/slash-first) is the diagnostic signature of a Unicode bidi reordering, not a data or i18n bug.

## What I expected

The detail text to render left-to-right, in the order the backend emits it, with only the *start* of an overlong string ellipsized (the visual point of the `dir="rtl"` + `truncate` combo), never with characters moved to the opposite end of the line.

## Root cause

`frontend/src/pages/SetupWizard.jsx:122` renders each System Check detail as:
```jsx
<span className="truncate text-xs text-fg-muted" dir="rtl" title={c.detail}>
  {c.detail}
</span>
```
`dir="rtl"` is being used as a left-truncation hack, paired with Tailwind's `truncate`, so long paths ellipsize at the *start* instead of the end. Under the Unicode bidi algorithm, an RTL paragraph places a leading run of European numbers (or other bidi-neutral characters, including `/`) at the visual *right* edge of the line instead of the left — which is exactly the digit/slash-first vs. letter-first split observed above.

The backend is not at fault — it emits correctly ordered, LTR strings. Confirmed at `backend/api/routers/setup/wizard.py`:
- line 387: `f"{ram:.1f} GB total"` (pass case) / lines 375, 383 (fail/warn cases, same order)
- line 403: `f"{free:.1f} GB free at {cache}"` (pass case) / line 399 (fail case)
- line 413 area: the `hf_cache_writable` check's detail is the bare cache path, also correctly ordered.

Same class, same hack at a second call site: `frontend/src/components/FirstRunSetup.jsx:149`:
```jsx
<code className="truncate font-mono text-[0.66rem] text-fg-muted" title={path} dir="rtl">
  {path}
</code>
```
used to left-truncate install-destination paths. A correct fix must cover **both** call sites.

(Line numbers verified against current `origin/main` at commit `53ff367c1fdde695f17673707cd46e2b27d41546`, which is where `main` currently sits.)

*Inference, not verified on those platforms:* this is pure CSS/Unicode-bidi behavior with no platform-specific code path (no `#[cfg(target_os = ...)]` branching involved, nothing WebView-engine-specific), so it is expected to reproduce identically on Windows and Linux. Only macOS was actually exercised in this clean-room run.

## Suggested fix

Preserve left-side truncation without triggering bidi reordering of the content, e.g.:
- `unicode-bidi: plaintext` on the span/code element (lets the browser derive directionality from the content itself instead of forcing RTL), or
- keep `dir="rtl"` on the truncating wrapper for the ellipsis-position behavior, but wrap the actual text in an inner `<bdi>` or `dir="ltr"` element so the characters themselves aren't reordered, or
- do the left-truncation in JS (compute an already-elided string) and drop `dir="rtl"` entirely.

Apply at both `frontend/src/pages/SetupWizard.jsx:122` and `frontend/src/components/FirstRunSetup.jsx:149`.

Chang-Jin-Lee:
## Summary

`dir="rtl"` on the preflight detail line and on the storage-path row is there to
move the ellipsis to the *start* of the line, so a long path keeps its tail on
screen. It also makes the line an RTL paragraph, and the Unicode bidi algorithm
places a leading run of European numbers or neutrals at the visual right edge of
one. Every detail beginning with a digit or a `/` therefore rendered with its
head at the end:

| Backend emits | Before | After |
|---|---|---|
| `48.0 GB total` | `GB total 48.0` | `48.0 GB total` |
| `219.5 GB free at /home/user/.cache/huggingface` | `GB free at /home/user/.cache/huggingface 219.5` | unchanged from the source string |
| `/home/user/.cache/huggingface` | `home/user/.cache/huggingface/` | `/home/user/.cache/huggingface` |

Letter-first details on the same screen (`Python 3.11.15`, `Reachable (63 ms)`)
were never affected, which is the signature of bidi reordering rather than bad
data. The backend emits these strings correctly ordered.

Closes #1848.

## Changes

The text at both call sites now sits in a `<bdi>`. That element is a
bidirectional isolate with an implicit `dir="auto"`, so the run is ordered by
its own content while the box keeps `direction: rtl` for the ellipsis side —
the left-truncation the hack exists for still works.

- `frontend/src/pages/SetupWizard.jsx` — preflight detail.
- `frontend/src/components/FirstRunSetup.jsx` — install-destination path. Both
  had to change; either one alone leaves the other screen broken, which is why
  #1848 filed them as one defect.
- `frontend/src/test/bidiStartTruncation.test.jsx` — regression test, below.
- `CHANGELOG.md` — Unreleased → Fixed.

No i18n keys, no schema, no engine surface, no new network call.

## Type

- [x] 🐛 Bug fix

## Testing

jsdom resolves no bidi, so the ordering itself was measured in headless
Chromium 153 against the same nesting the preflight row uses. The numbers are
the x offset of the first and last word of each string, relative to its line
box:

| text | variant | head | tail | |
|---|---|---|---|---|
| `48.0 GB total` | bare | x=49 | x=20 | reversed |
| `48.0 GB total` | `<bdi>` | x=0 | x=46 | |
| `/home/user/.cache/huggingface` | bare | x=175 | x=107 | reversed |
| `/home/user/.cache/huggingface` | `<bdi>` | x=0 | x=112 | |
| `Python 3.11.15` | bare | x=0 | x=44 | letter-first, never broken |
| an overflowing path | `<bdi>` | x=-142 | x=288 | head still clipped off the start |

That last row is what keeps the fix honest: the isolate restores the order
without giving up the start-side ellipsis.

The committed test pins the DOM shape that result depends on, and adds a source
scan that fails any future `dir="rtl"` element owning its text directly — the
bug shipped at two call sites written months apart, so a third is the expected
failure mode. Both assertions go red with the two source edits reverted, naming
`components/FirstRunSetup.jsx:149` and `pages/SetupWizard.jsx:122`.

Ran locally on Linux:

- `bun run test` — 329 files, 2732 tests pass.
- `bun run lint`, `bun run format:check`, `bun run typecheck:ci`,
  `npx vite build --mode development` — clean. The two `oxlint` warnings on
  `FirstRunSetup.jsx` (`max-lines`, `only-export-components`) predate this
  branch.
- `pytest tests/test_changelog_style.py tests/test_locale_parity.py tests/test_no_hardcoded_cjk.py` — 542 pass.

- `uv run pytest backend/ -x -q` — 358 pass. Nothing here touches Python, so
  that is a baseline rather than evidence, but it is green.

Cross-platform: pure CSS and Unicode bidi, no `#[cfg(target_os)]` and no
engine-specific path, so the same markup behaves identically on WebKitGTK,
WKWebView and WebView2. Verified on Linux/Chromium; #1848 was reported on
macOS.

## Checklist

- [ ] I've updated relevant documentation — no doc describes this markup
- [ ] Version files are in sync — no version bump
- [x] No runtime-behavior change for `tests/fixtures/omnivoice_data/`; this is
      presentational markup only

Wraps setup details and installation paths in `<bdi>` while retaining RTL start-side truncation. This prevents numeric- and slash-prefixed strings from bidi reordering while preserving backend text order. Regression tests cover both affected locations; human review should verify browser-engine behavior across supported platforms.


Chang-Jin-Lee:
On CodeRabbit's inconclusive **Cross-Platform Default Parity** check, since it is
a hard rule and I would rather answer it than leave it hanging.

The mechanism is UAX#9, which every shipping engine implements the same way. #1848
was reported on macOS/WKWebView and I reproduced the identical reordering on
Linux/Chromium, so two of the three engines already agree on the broken rendering
before any fix. `<bdi>` is the isolate that same spec defines for it, and it has
been in WebKit since Safari 6.1, Blink since Chrome 16, and Gecko since Firefox 10.
The diff carries no platform branch, no conditional style and no engine sniff, so
all three get the same markup.

What I could not do is measure WebKit and WebView2 directly. `playwright install
webkit` needs `libevent-2.1-7t64` and `libavif16` installed with root, which this
environment does not allow. If you want those two numbers in the PR before merge,
say the word and I will get them.

Issue #1937, opened 2026-09-09
[Bug] 503 Service Unavailable: [starting] VoiceStudio


MarcoPropaga:
## Error

```
503 Service Unavailable: [starting] VoiceStudio is still starting (Loading ML runtime (PyTorch)…). Retry shortly.
$@tauri://localhost/assets/client-CNTJBHFd.js:5:13265
```

## Environment

**Version:** `0.5.1`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.6.2-arm64-arm-64bit`
**Python:** `3.11.15`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M3 Pro`
**RAM:** `18 GB`
**Disk free:** `35.3 GB`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 2 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-09-09T00:15:46.000Z (8 s ago)
**Exit:** `exit code 1`
**Uptime before crash:** 12889 s
**Backend version:** `0.5.1`

```
… (truncated)
.model] Native TTS preload skipped: OmniVoice uses crash isolation on this host.
2026-09-08 17:35:39,746 INFO [mcp.server.streamable_http_manager] StreamableHTTP session manager started
2026-09-08 17:35:39,746 INFO [omnivoice.api] MCP server mounted at /mcp
2026-09-08 17:35:39,821 INFO [omnivoice.api] Deferred startup complete — all routes live.
INFO:     127.0.0.1:53064 - "WebSocket /ws/events" [accepted]
INFO:     connection open
2026-09-08 17:35:43,389 INFO [omnivoice.events] WS client connected (1 total)
2026-09-08 17:36:09,746 INFO [omnivoice.model] GPU pool: MPS detected, using 1 worker (shared system memory)
2026-09-08 17:36:09,766 INFO [omnivoice.api] Capture ASR backend selected: parakeet-mlx
2026-09-08 17:36:14,816 INFO [omnivoice.watermark] Watermark prefetch skipped: AudioSeal checkpoint is not cached
2026-09-08 17:51:10,002 INFO [omnivoice.asr] Idle timeout reached. Unloading capture ASR (ParakeetMLXBackend) to free memory.
[...]

Issue #1936, opened 2026-09-08
[Backend] Backend failed to start


MarcoPropaga:
## Error

```
Port 3900 is already in use by another application, and VoiceStudio could not free it. Quit whatever is using that port (another copy of VoiceStudio, or an app that claimed it) and try again.
@tauri://localhost/assets/main-app-Ccm2oqM6.js:16:1862
vd@tauri://localhost/assets/main-app-Ccm2oqM6.js:9:126192
@tauri://localhost/assets/main-app-Ccm2oqM6.js:9:131186
en@tauri://localhost/assets/main-app-Ccm2oqM6.js:9:15067
Cd@tauri://localhost/assets/main-app-Ccm2oqM6.js:9:127419
up@tauri://localhost/assets/main-app-Ccm2oqM6.js:10:28473
cp@tauri://localhost/assets/main-app-Ccm2oqM6.js:10:28295
```

## Environment

**Version:** `0.5.1`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 45 s before this report

## Recent actions

```
17:29:38 view:launchpad
17:30:29 view:studio
17:32:30 generate:start (audio)
```

Chang-Jin-Lee:
@MarcoPropaga — #1933, #1935 and this one are the same condition on the same machine (macOS, 0.5.1, identical platform string, filed 16:50 → 17:32), and #1937's `503 [starting]` is the tail of it. Three of the four can be closed as duplicates once this is triaged.

**A workaround so you are not stuck while this is decided.** The port is almost certainly held by an orphaned VoiceStudio backend from an earlier run, which has no window to quit. In Terminal:

```
lsof -ti tcp:3900 | xargs kill
```

Then relaunch. If `lsof` prints nothing, something else really does own 3900 and the message is accurate for your case.

**Why the app cannot get you out of it.** `bootstrap.rs:414` reports this after `backend::free_port_or_report()` returns false, and that function calls `kill_orphan_on_port()` first — which never signals anything:

```rust
pub fn kill_orphan_on_port(port: u16) {
    if port_in_use(port) {
        log::warn!("Refusing to signal the unowned listener on port {port}; …");
    }
}
```

[...]

PR #1996, opened 2026-09-10
fix(backend): name who actually holds port 3900 (#1933)

### LINKED ISSUE #1933: [Bug] The local VoiceStudio backend could not start, so this request had nowhere to go
## Error

```
The local VoiceStudio backend could not start, so this request had nowhere to go. The app reported:

Port 3900 is already in use by another application, and VoiceStudio could not free it. Quit whatever is using that port (another copy of VoiceStudio, or an app that claimed it) and try again.

Open the details for the full output, or use Retry / Clean & Retry in Settings → Logs → Backend.
$@tauri://localhost/assets/client-CNTJBHFd.js:5:12429
```

## Environment

**Version:** `0.5.1`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`

## Backend reachability

**Last backend response:** 2 min before this report
**First failure:** 2026-09-08T20:17:33.812Z
**Attempts before giving up:** 4
**Mode at failure time:** `desktop`
**Transport error:** `Load failed`

## Recent actions

```
16:50:09 view:transcriptions
16:50:15 view:projects
16:50:38 view:voice
16:54:33 view:dub
16:54:34 view:studio
16:55:07 generate:start (audio)
17:01:05 generate:start (audio)
17:04:12 generate:start (audio)
17:08:03 generate:start (audio)
17:08:21 view:settings
17:13:54 view:catalogue
17:14:50 view:projects
17:14:52 view:transcriptions
17:14:53 view:gallery
17:14:54 view:audiobook
17:14:55 view:stories
17:14:57 view:dub
17:14:59 view:studio
17:15:49 generate:start (audio)
17:17:33 generate:start (design)
```

debpalash:
Fixes #1933. Thanks @Chang-Jin-Lee for the diagnosis on #1936.

The port-conflict failure asserted "already in use by **another application**" without ever asking who held the port. In the reports behind #1933 — and its duplicates #1935, #1936, #1937, all the same machine inside one 45-minute window — the holder was the user's *own* orphaned backend from an earlier run. The app told them to quit a copy of VoiceStudio that has no window, and nothing in the message would have worked.

**The check already existed.** `running_backend_version(port)` asks `/system/info` who is there, and is already trusted for the more consequential decision of whether to attach to a healthy same-version backend. It just was not consulted on this path.

`port_holder()` now returns one of three answers and `port_conflict_message()` words the failure from it:

- **our own backend at this version** (or one too old to report one) — say so, say it has no window to quit, and give the terminal command that ends it;
- **our own backend at a different version** — name the version, which is what identifies it, and give the same command;
- **anything else** — the existing wording, now actually justified.

The terminal command is offered *only* for a listener that identified itself as ours. An unidentified one keeps the conservative wording: a user must never be told to go kill a process that may not be theirs. A listener that accepts a connection but does not answer `/system/info` counts as unidentified, which is the reading that cannot do harm.

All three sites that reported this — take-ownership, respawn, and the early-exit path on `EXIT_PORT_IN_USE` — go through the one builder now.

**Deliberately unchanged.** `kill_orphan_on_port` still refuses to signal a PID discovered through lsof/netstat. That refusal is correct: the reuse race is real and a matching foreign service must never be terminated. This changes what the user is told, not what the app is willing to kill.

**Verification.** Five Rust tests — one per branch, the suffix behaviour, and one pinning every wording against the `detectHints` matcher. That regex is what turns these English strings into the localised `bootstrap.hint_port`, and an earlier draft of one message silently lost the translation by saying "is held by". The frontend test that pinned the old literals now pins the new ones. 241 Rust lib tests and 2851 vitest tests green.

https://claude.ai/code/session_017ypcgSsh5j2PEonSJiAU1S

Port-3900 conflict handling now identifies the listener and provides actionable recovery commands across startup, respawn, and teardown paths. This addresses issue `#1933` and preserves process-killing behavior. Human review should verify platform-specific detection and localization matching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,8 @@ the frozen-backend fallback mirror it for their toolchains.
 
 ### Fixed
 
+- A port-3900 conflict now names who is actually holding it, and gives the command that ends an orphaned backend, instead of telling you to quit an app that has no window (#1933) — thanks @Chang-Jin-Lee!
+
 - Windows desktop launches no longer freeze at "Loading ML runtime (PyTorch)": the parent-liveness watchdog polls the stdin pipe instead of leaving a read pending, which deadlocked numpy's OpenBLAS initializer (#1952)
 - `bun desktop-prod` and `bun desktop-fresh` find Rust and uv from a terminal opened before they were installed, as `bun desktop` already did; a missing Rust toolchain fails up front with the install steps (#1952)
 - Windows desktop launches no longer freeze at "Loading ML runtime (PyTorch)": the parent-liveness watchdog polls the stdin pipe instead of leaving a read pending, which deadlocked numpy's OpenBLAS initializer (#1955)
--- a/frontend/src-tauri/src/backend.rs
+++ b/frontend/src-tauri/src/backend.rs
@@ -416,6 +416,99 @@ pub fn free_port_or_report(port: u16) -> bool {
     false
 }
 
+/// Who is holding the port, as far as an HTTP request can tell (#1933).
+///
+/// The port-conflict failure used to assert "already in use by **another
+/// application**" without asking. That is wrong in the common case: the holder
+/// is usually the user's own orphaned backend from an earlier run, which has
+/// no window to quit — so the message sent them to close a copy of VoiceStudio
+/// they cannot see, and gave them nothing that would work.
+///
+/// The identity check already existed (`running_backend_version`, used to
+/// decide whether to attach to a healthy same-version backend, a far more
+/// consequential decision). It just was not consulted here.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PortHolder {
+    /// Nothing answered — either genuinely free, or a listener that accepts a
+    /// connection but does not respond in time.
+    Unknown,
+    /// A VoiceStudio backend, carrying the version it reports. Empty string
+    /// when it predates the `app_version` field.
+    OurBackend(String),
+    /// Something answered `/system/info` and it was not us.
+    Foreign,
+}
+
+/// Ask the listener on `port` who it is. One HTTP request, the same one
+/// `running_backend_version` makes.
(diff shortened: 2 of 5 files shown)


debpalash:
Bot review triage.

**Fixed — Greptile P1 (security), blind kill.** The one-liner did not preserve the identity `port_holder` established: `lsof -ti tcp:PORT` matches connected clients as well as the listener, and `findstr :3900` matches `:39001` and established connections. Both platforms now show the listening process and kill by a PID the user has looked at.

**Fixed — CodeRabbit, body sniff.** `running_backend_version` accepts any `/system/info` body containing `model_checkpoint` or `data_dir`. Fine for deciding whether to attach; not fine for an answer that ends in a kill command. `port_holder` now requires the `x-omnivoice-backend` header, the same gate `startup_progress` applies. Two tests against a real loopback responder.

**Fixed — the integration test.** `port_conflict_is_named_as_a_port_conflict` pinned a literal sentence while its own comment said the contract was the `detectHints` matcher. The reworded message still matched the matcher and still failed the test. It asserts the actual contract now.

**Not a finding — Greptile, "Port Errors Bypass Localization".** Every Rust-side failure message in this codebase is English by design, rendered verbatim in the error box, with `detectHints` deriving a localised hint shown beside it. That is the documented architecture, and the comment stating it is quoted in `portInUseHint.test.js` and in the message builder's own docs. The localization hard rule governs CJK in the frontend translation layer, not Rust diagnostics. Introducing structured variants here would be a new architecture for one message while every neighbouring one keeps the old, which is worse than either.

The gap that IS real: the localised `bootstrap.hint_port` says "Close other instances of VoiceStudio or apps using that port", which is the advice this PR exists to correct. Worth a follow-up to that key's text in all 21 locales; out of scope here because it changes strings this PR does not own.

Issue #1949, opened 2026-09-09
[Bug] Pronunciation dictionary: IPA and CMU entry types are silently ignored — they never reach any engine


AlphanB:
### Before filing

- [x] I searched [existing issues](https://github.com/debpalash/VoiceStudio/issues?q=is%3Aissue) and this isn't a duplicate.
- [x] I'm on the latest release (or `main`) — older builds may already be fixed.

### What happened?

Settings → Pronunciation offers three entry types in the spelling-notation dropdown: **Respelling**, **IPA**, **CMU**. Only Respelling does anything. IPA and CMU entries save without error, get a badge in the list and can be toggled on — but they are discarded before term matching and never reach any engine.

Two symptoms, one root cause:

1. **"Test a sentence" reports no match for an IPA/CMU entry.** With an enabled entry `Alphan` → `AA0 L P HH AA0 N` (type CMU), typing `Alphan` into the test field shows *"No entries match; spoken as written."* That message is wrong — the term does match; the entry is dropped because of its type.
2. **Switching the entry to Respelling makes the preview fire, but then the phoneme string is handed to the model as graphemes**, because respelling is plain text substitution. So there is no path through the UI that gets a validated IPA/CMU string to the engine as phonemes.

### Root cause — `backend/services/pronunciation.py:212`

```python
# Phase 1: only respelling rows substitute text. IPA/CMU rows without a
# respelling fall through (Phase 2 lowers them to engine markup); we do
# NOT feed a raw IPA string into the grapheme stream.
if etype != "respelling":
    continue
```

`entries_for_language()` filters IPA/CMU rows out *before* it matches terms. That function is the single input to both the synth path (`backend/api/routers/generation.py:1681`) and the `/pronunciation/test` dry run (`backend/api/routers/pronunciation.py:251`), which is why both go quiet at the same time.

Verified against `main` @ `18f56a79`:

```python
>>> rows = [{'term':'Alphan','replacement':'AA0 L P HH AA0 N','type':'cmu','language':'*','enabled':1}]
>>> entries_for_language(rows, 'tr')
{}
>>> apply_pronunciation('Merhaba, ben Alphan.', rows, 'tr')
'Merhaba, ben Alphan.'
```

[...]

PR #2002, opened 2026-09-10
fix: surface IPA/CMU pronunciation entries as not-yet-applied (#1949)

utkarsha741:
Addresses #1949, scoped to the minimum fix as discussed in my claim comment.

## What this does
- `entries_for_language()` now optionally returns which IPA/CMU entries matched but were skipped (Phase 1 only substitutes respelling rows)
- `/pronunciation/test` surfaces this as a new `skipped_terms` field, additive to the existing response shape
- Settings → Pronunciation now badges IPA/CMU entries as "Not yet applied" and notes it in the test preview

## What this doesn't do
Phase 2 (actual CMU/IPA lowering to engine markup) is intentionally out of scope here — happy to look at that separately once this lands.

## Testing
- All 39 existing tests in `tests/test_pronunciation.py` pass unmodified, including `test_ipa_row_never_enters_the_grapheme_stream`
- Added 4 new frontend tests covering the badge and the preview message (21/21 passing in `PronunciationPanel.test.jsx`)
- Manually verified via the backend API that a CMU entry now reports `skipped_terms` instead of silently reporting no match

### DIFF
--- a/backend/api/routers/pronunciation.py
+++ b/backend/api/routers/pronunciation.py
@@ -249,12 +249,13 @@ def test_substitution(req: PronTestRequest):
             "FROM pronunciation_entries"
         ).fetchall()
     substituted = apply_pronunciation(req.text, rows, req.language)
-    applied = entries_for_language(rows, req.language)
+    applied, skipped = entries_for_language(rows, req.language, include_skipped=True)
     return {
         "input": req.text,
         "substituted": substituted,
         "changed": substituted != req.text,
         "applied_terms": sorted(applied.keys(), key=len, reverse=True),
+        "skipped_terms": skipped,
     }
 
 
--- a/backend/services/pronunciation.py
+++ b/backend/services/pronunciation.py
@@ -176,7 +176,9 @@ def _lang_prefix(language: Optional[str]) -> Optional[str]:
     return s[:2]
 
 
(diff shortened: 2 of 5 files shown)


debpalash:
Thanks @utkarsha741 — the part of this that was new is landing in #2003 with your credit in the changelog.

**What was already on main.** #1949 was fixed by #1984 a few hours before this arrived. The backend skip detection and the test-preview line are there under different names: `inert_entries_for_language()` in `backend/services/pronunciation.py`, surfaced as `inert_entries` on the `/pronunciation/test` response, rendered by the panel as `pronunciation.test_inert`. Merging the `skipped_terms` version would have given one behaviour two implementations and two response fields.

Sorry you spent effort on ground that had just been covered — the open-PR and recently-merged queue is worth a glance before starting, and that is on the project to make more visible, not on you.

**What was genuinely new, and is landing.** The per-row "not applied yet" badge. #1984 only made this visible in the test preview, which a user sees only if they run a test; the entry list is where they look at what they saved, and there an IPA row still looked like every other working row. Good catch.

Two changes on the way in: the string is in all 21 locales rather than `en` alone, since a partially translated panel is its own bug, and a regression test that fails with the badge condition disabled.

Closing this in favour of #2003.

PR #2003, opened 2026-09-10
feat(pronunciation): badge a stored-but-inert entry in the list (#1949)

debpalash:
Takes the part of #2002 by @utkarsha741 that #1984 did not already cover.

An IPA or CMU row saves, validates and toggles on, and is then dropped before term matching — Phase 1 only substitutes respelling. #1984 made that visible in the "Test a sentence" preview, which the user sees only if they run a test. The entry **list** is where they look at what they have saved, and there it still looked like every other working row.

So the row carries the same fact: a warning badge beside the type and scope badges, on IPA and CMU only. Badging a respelling row would be the opposite lie — those do take effect.

**What I did not take.** The rest of #2002 is already on main under different names: it re-added the backend skip detection and the test-preview line as `skipped_terms`, where `inert_entries_for_language` and `inert_entries` have shipped since #1984. Landing that half would have been a second implementation of one behaviour, with two response fields for it.

**Verification.** The string is in all 21 locales, not just `en`, so the badge is not an English island in a translated panel. Fails before, passes after: with the badge condition disabled the new test cannot find it. 2858 vitest tests pass, and 536 locale-parity and CJK guard tests.

Adds a translated warning badge for enabled non-`respelling` pronunciation entries that are stored but not applied during term matching. Tests cover enabled, disabled, `respelling`, and unfamiliar entry types across all 21 locales. Human review should confirm translations and future pronunciation types do not require different handling.


debpalash:
Bot review triage — both bots found the same thing independently, and half of it is right.

**Taken: the enabled gate.** `e.type !== 'respelling'` badged rows the user had switched OFF. A disabled entry is indeed not applied, but for a reason the toggle already shows, and labelling it "not applied yet" reads as a defect rather than their own choice. `inert_entries_for_language` excludes disabled rows for exactly that reason, so the badge now matches it. Test fails without the gate.

**Not taken: naming `ipa` and `cmu` explicitly.** The backend's rule is that everything which is not respelling is inert today, and mirroring it keeps the two in step. An explicit list would silently stop badging a notation added later — an entry that saves, validates, toggles on and quietly does nothing, which is precisely the invisibility #1949 exists to remove. A test now pins that direction with an unfamiliar type.

utkarsha741:
Thanks for taking this the rest of the way and catching the enabled-entry bug — appreciate the credit! Good to know about #1984, will check for prior art before diving in next time.

Issue #1371, opened 2026-08-04
[Bug] The local OmniVoice backend crashed (exit code -1073740791) 2 s ago and is being


NullAILab:
## Error

```
The local OmniVoice backend crashed (exit code -1073740791) 2 s ago and is being restarted — this request could not reach it. Open the crash notice for the error output, or check Settings → Logs → Backend.
ApiError: The local OmniVoice backend crashed (exit code -1073740791) 2 s ago and is being restarted — this request could not reach it. Open the crash notice for the error output, or check Settings → Logs → Backend.
    at Q (http://tauri.localhost/assets/client-CxP2DadA.js:3:11096)
    at async s (http://tauri.localhost/assets/generate-DL4nRL4S.js:1:408)
    at async Ci (http://tauri.localhost/assets/main-app-DQafD-09.js:16:27910)
    at async o (http://tauri.localhost/assets/generate-DL4nRL4S.js:1:219)
    at async http://tauri.localhost/assets/main-app-DQafD-09.js:17:3650
```

## Environment

**Version:** `0.4.2`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36 Edg/150.0.0.0`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 19 s before this report
**First failure:** 2026-08-04T20:16:39.889Z
**Attempts before giving up:** 4
**Mode at failure time:** `desktop`
**Transport error:** `Failed to fetch`

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-08-04T20:17:06.000Z (13 s ago)
**Exit:** `exit code -1073740791`
**Uptime before crash:** 18 s
**Backend version:** `0.4.2`

```
… (truncated)
oaded successfully.
2026-08-04 22:16:58,654 INFO [omnivoice.model] Preload complete — model ready.
2026-08-04 22:16:58,845 INFO [speechbrain.utils.quirks] Applied quirks (see `speechbrain.utils.quirks`): [disable_jit_profiling, allow_tf32]
2026-08-04 22:16:58,845 INFO [speechbrain.utils.quirks] Excluded quirks specified by the `SB_DISABLE_QUIRKS` environment (comma-separated list): []
[...]

PR #2072, opened 2026-09-13
fix(docker): install cuDNN 8 compat libraries in the CUDA Docker image

### LINKED ISSUE #2050: [Install] No `cudnn8_compat` present in Docker image
### Where did it fail?

Python / uv environment bootstrap

### The error

```text
Trying to synthesize the demo voice from a freshly initialized state with the `omnivoice` engine, produces a warning that `cudnn8_compat` is missing and restarts the app after a timeout.

I can see that `scripts/setup.py` installs it, but I don't see it in the Docker file or the torch requirements.

This occurred after an upgrade of the same environment from `0.5.1`.

I can reproduce with the example `deploy/docker-compose.yml`.
```

### Operating system

Docker

### How are you installing it?

Docker image

### Version

0.5.2

### Network conditions (model/dependency downloads)

Normal / unrestricted

### Logs / diagnostic bundle

```text
2026-09-11 19:25:59,831 WARNING [core.cudnn8] CTranslate2 ASR engines unavailable: CUDA is active but libcudnn_ops_infer.so.8 cannot be loaded (libcudnn_ops_infer.so.8: cannot open shared object file: No such file or directory). WhisperX and faster-whisper are CTranslate2, which requires cuDNN 8; loading it is not optional and its absence aborts the backend process outright rather than raising (#1371). Reinstall the compat libraries with `uv pip install --target <venv>/lib/pythonX.Y/site-packages/cudnn8_compat nvidia-cudnn-cu12==8.9.7.29`, or pin a non-CTranslate2 engine with OMNIVOICE_ASR_BACKEND=pytorch-whisper.
```

### What have you already tried?

A simple:
```Dockerfile
RUN uv pip install --python /opt/conda/bin/python3 --no-deps --target "/opt/conda/lib/python3.11/site-packages/cudnn8_compat" nvidia-cudnn-cu12==8.9.7.29
```
in a derived Docker image seems to fix this for me.

Additionally the automatic drag and drop of the diagnostic bundle does not seem to work for me. Please find it attached here.

[omnivoice-diagnostics-20260911-192641.zip](https://github.com/user-attachments/files/32128946/omnivoice-diagnostics-20260911-192641.zip)

basil-k-aji-dev:
Fixes #2050

The CUDA base image is `pytorch/pytorch:2.8.0-cuda12.8-cudnn9-runtime`, so it ships cuDNN 9. CTranslate2 — WhisperX and faster-whisper — links cuDNN 8, and per the note in `backend/core/cudnn8.py` its absence aborts the backend process outright rather than raising (#1371).

`scripts/setup.py` side-loads the cuDNN 8 libraries for source installs, but the Dockerfile never did. Docker users therefore lose every CTranslate2 ASR engine, which surfaces as the demo synthesis timing out and:

```
WARNING [core.cudnn8] CTranslate2 ASR engines unavailable: CUDA is active but
libcudnn_ops_infer.so.8 cannot be loaded
```

## Change

Install the same `nvidia-cudnn-cu12==8.9.7.29` shim during the image build.

Two details worth flagging:

**The target is derived from `sys.prefix`, not hardcoded.** `backend/core/cudnn8.py::compat_dirs()` searches `<sys.prefix>/lib/pythonX.Y/site-packages/cudnn8_compat/nvidia/cudnn/lib`, and `sys.prefix` differs between the conda-based CUDA image and the ROCm venv. Deriving it keeps the install and the lookup in agreement if the base image's Python version or layout moves. I verified the computed target is exactly the parent of the directory `compat_dirs()` looks in.

**A post-install assert fails the build if no `.so.8` libraries landed.** Without it, a silently empty install would reappear as this same runtime warning, which is what made the original report hard to place.

Guarded to `GPU_FLAVOR=cuda` — ROCm does not use cuDNN — and `--no-deps` keeps the install to the cuDNN wheels alone, leaving the base image's torch stack untouched, consistent with the existing torch-clobber guard directly above.

## Quality gates

- **Cross-platform parity:** no change to default behavior on any desktop platform; this only affects the CUDA container build, and the ROCm path is explicitly skipped.
- **Engine back-compat:** additive. Already-installed engines need no reinstall or re-download — engines that were previously unusable in Docker start working.
- **i18n / DB schema / security posture:** not touched, no user-facing strings, no migration, no new HTTP surface.
- **Local-first:** a build-time package install from PyPI, the same mechanism as the existing `uv pip install` steps in this Dockerfile. No new runtime outbound calls.

## Testing

`docker build --check -f deploy/Dockerfile .` passes — the only warning is `InvalidBaseImagePlatform`, because I am on arm64 and the base image is amd64.

**What I could not run:** I could not build the image or run `uv run pytest backend/ -x -q` on this machine — it is an arm64 Raspberry Pi, and the CUDA base image is amd64-only. So the install step itself is unexercised here and I would rely on CI for it. What I did verify directly is the path logic, since that is the part most likely to be wrong:

```
Dockerfile target : <sys.prefix>/lib/pythonX.Y/site-packages/cudnn8_compat
compat_dirs()     : <sys.prefix>/lib/pythonX.Y/site-packages/cudnn8_compat/nvidia/cudnn/lib
target is the parent of the searched lib dir: True
```

The pinned version matches the one `scripts/setup.py` already installs, so it is not a new dependency choice.

The reporter in #2050 confirmed a hardcoded variant of this fix works in a derived image; this PR generalises it and adds the verification step.

The CUDA Docker image now installs and verifies `nvidia-cudnn-cu12==8.9.7.29` in the `cudnn8_compat` path required by CTranslate2-based WhisperX and faster-whisper engines, with documentation and regression coverage added. The install runs only for `GPU_FLAVOR=cuda` and uses `--no-deps`. Full image builds and backend tests were not run because the available machine is arm64 while the CUDA image is amd64.


basil-k-aji-dev:
Thanks — title updated to `fix(docker): ...`.

On the local-first warning, I think it is being applied to the wrong layer, but I would rather lay out the reasoning than quietly dismiss it.

The gate in CONTRIBUTING reads:

> **Local-first:** no new outbound calls except GitHub Issues (opt-in reporting) and HuggingFace model downloads. Never log or persist secrets or absolute home paths.

That sits among runtime properties — logging, persisted secrets, home paths — so I read it as constraining what the running application does on a user's machine, not what the image build fetches. This change adds nothing at runtime; it only puts files into the image.

At build time the Dockerfile already reaches the network in four places:

- `deploy/Dockerfile:20` — `bun install --frozen-lockfile` (npm registry)
- `deploy/Dockerfile:48` — `apt-get install` (Debian archives)
- `deploy/Dockerfile:64` — `python3 -m pip install uv` (PyPI)
- `deploy/Dockerfile:90` — `uv pip install --constraint deploy/torch-constraints.txt .` (PyPI)

My line 119 is the same mechanism as 64 and 90, pinned to the exact version `scripts/setup.py` already installs, so it is not a new outbound channel or a new dependency decision — it is the step that was missing relative to the source install.

Vendoring the wheel as the bot suggests would mean carrying ~700 MB of platform-specific CUDA binaries in-tree, which seems a worse trade than one pinned install alongside the four that already run.

Happy to be overruled — if you would prefer it vendored, sourced from a prebuilt artifact, or made opt-in behind a build arg, say which and I will rework it.

Issue #1931, opened 2026-09-08
[Bug] RTX 50-series (Blackwell sm_120): backend crashes in ml_imports — torch 2.8.0 pin lacks sm_120 kernels, and torchaudio 2.9 removed set_audio_backend()


kn3nan3-gif:
## What happened?

Running VoiceStudio from source on an **RTX 5070 Ti Laptop GPU (Blackwell, compute capability sm_120)** the backend never becomes ready — it crashes during the `ml_imports` startup phase, so the desktop app stays stuck on "starting backend" and `/health` returns 503.

There are two confirmed root causes, both triggered by the fact that Blackwell (sm_120) users are forced off the repo's pinned torch version.

### Root cause 1 — pinned torch 2.8.0 has no sm_120 kernel support

`uv.lock` pins `torch 2.8.0+cu128`. On an RTX 50-series GPU, `import torch` dies with a native access violation (`exit code 0xffffffff`) — the known sm_120 issue: Blackwell needs torch 2.8 *nightly* / 2.9+ with cu128.

Upgrading to `torch 2.9.1+cu128` (edit `[tool.uv] constraint-dependencies` in `pyproject.toml` → `uv lock` → `uv sync`) fixes the `import torch` crash. CUDA then works (verified: `torch.cuda.is_available()` is True and a GPU matmul runs, device capability `sm_(12, 0)`).

### Root cause 2 — `torchaudio.set_audio_backend()` was removed in torchaudio 2.9

Once on torch 2.9.1, startup now crashes at `backend/main.py:627`:

```
AttributeError: module 'torchaudio' has no attribute 'set_audio_backend'
FATAL: backend startup failed during 'ml_imports': module 'torchaudio' has no attribute 'set_audio_backend'
```

torchaudio 2.9.x removed `set_audio_backend()` entirely (`soundfile` has been the only backend since 2.0). The call at `backend/main.py:627` is legacy and should be guarded or removed:

```python
if hasattr(torchaudio, "set_audio_backend"):
    torchaudio.set_audio_backend("soundfile")
```

## Steps to reproduce

1. On a Windows machine with an RTX 5070 / 5070 Ti / 5080 / 5090 (sm_120).
2. `git clone https://github.com/debpalash/VoiceStudio && cd VoiceStudio && bun install && bun run desktop-prod`
3. The backend crashes in `ml_imports`; `/health` stays 503 and `/startup/progress` reports `ml_imports` active indefinitely (or the process exits).

[...]

PR #2083, opened 2026-09-14
fix(audio): fall back to soundfile when torchaudio 2.9 routes I/O through TorchCodec

Moep90:
## Problem

On torch >= 2.9 without TorchCodec installed, **every audio read and write fails**:

```
ImportError: TorchCodec is required for save_with_torchcodec.
ImportError: TorchCodec is required for load_with_torchcodec.
```

torchaudio 2.9 routes `save()`/`load()` through TorchCodec, which needs FFmpeg
*shared libraries* on the system (not the bundled static binaries `media_tools`
fetches). Where those are absent, TTS generation returns 500, and voice cloning,
watermarking and the dub pipeline fail on the reference read.

This is a direct follow-up to #1931. That issue established that RTX 50-series
(Blackwell, sm_120) owners **have no choice** but to move off the torch 2.8.0
pin, which has no sm_120 kernels — and its fix guarded `set_audio_backend()` in
`main.py` but stopped there, leaving `load()`/`save()` unprotected. So this is
not an ARM-only or exotic-host issue: it reaches the exact cohort #1931 named.

## Root cause

Two independent sites, same cause:

1. `_safe_torchaudio_save` had no fallback when `torchaudio.save` demands TorchCodec.
2. `load_audio` already had a pydub fallback written for exactly this situation,
   but caught `(RuntimeError, OSError)` — and the failure is an `ImportError`,
   so the fallback never ran.

Note `backend="soundfile"` does **not** avoid either: torchaudio 2.9 accepts
that argument and ignores it. `load_audio` was already passing it.

## Fix

- `_safe_torchaudio_save`: on that `ImportError`, write through the module's own
  audited `_safe_soundfile_write` instead of failing. The tensor is already
  normalised (dtype/clamp/contiguity/shape) at that point, so the fallback
  inherits those guarantees rather than duplicating them.
- `_safe_soundfile_write`: optional `format` passthrough. soundfile cannot infer
  a container from a file-like object, and `/v1/audio/speech` writes to `BytesIO`.
- `load_audio`: add `ImportError` to the existing `except`, so the fallback the
  original author wrote actually fires.

`torchaudio.save` stays the primary path everywhere — on torch 2.8, or any host
with FFmpeg shared libs, behaviour is byte-for-byte unchanged.

## Tests

Five new regression tests, all monkeypatching the TorchCodec `ImportError`, so
they exercise the fallback on **any** torchaudio version (CI on 2.8 included):

```
with the fix:     5 passed
without the fix:  5 failed
```

Existing `tests/backend/services/test_audio_io.py`: **34 passed**, no regression
from the `format` parameter.

## Verified on

aarch64 (NVIDIA GB10) with torch 2.9.1+cu130 and no FFmpeg shared libraries —
the environment that reproduces the bug. End-to-end after the fix: TTS in
wav/flac/mp3/opus, voice cloning, and dub upload → transcribe → translate →
render.

## Out of scope

- `backend/tests/test_atomic_wav.py` has 2 failures on a torch-2.9-without-
  TorchCodec host, from its own direct `torchaudio.load()` calls in assertions
  (lines 29, 44, 49) — not reachable by library code. This PR takes that file
  from 4 failures to 2; fixing the remaining assertions is a separate change.
  CI (torch 2.8) is unaffected either way.
- `[tool.uv.sources]` excludes linux-aarch64 from the CUDA index, so `uv sync`
  there silently installs CPU torch. Separate concern, separate PR.

Adds soundfile and pydub fallbacks when torchaudio 2.9+ cannot use TorchCodec or FFmpeg libraries. Corrects pydub normalization for 8-, 16-, 24-, 32-bit, and float audio and supports explicit formats for file-like outputs. Review fallback behavior for uncommon formats and direct `torchaudio.load()` calls that remain out of scope.


Moep90:
Both bots flagged the pydub fallback normalization. Fixed in 6dafebee — the finding is real, but the prescribed remedy is not what the code needs, so recording the evidence.

**The bug.** `load_audio`'s fallback divided every decode by `32768.0`, correct only for 16-bit. Measured peak of a 0.5-amplitude sine, pydub 0.25.1:

| subtype | pydub `sample_width` | peak `/32768.0` (before) | peak `/max_possible_amplitude` (after) |
|---|---|---|---|
| PCM_U8 | 1 | 0.00195 (256x too quiet) | 0.500 |
| PCM_16 | 2 | 0.500 | 0.500 |
| PCM_24 | 4 | 32768.0 (32768x too loud) | 0.500 |
| PCM_32 | 4 | 32768.0 | 0.500 |
| FLOAT | 4 | 32768.0 | 0.500 |

Nothing downstream clamps, so a non-16-bit reference clip silently became noise. Catching `ImportError` is what makes this matter: the fallback used to run only for formats torchaudio could not open, but on torchaudio >= 2.9 without TorchCodec it is the *only* path, which turns a loud failure into a silent wrong result.

**Why no 24-bit special case.** Both reviews ask for one, on the grounds that pydub keeps 24-bit as sign-extended 32-bit data. pydub's own comment says the same ("The values are not scaled up to the 32 bit range", `audio_segment.py:244-246`) — but the loop below it writes the pad byte *first*, so it lands in the LSB and the sample becomes `(old << 8) | pad`, occupying the full int32 range. The table above confirms it: PCM_24 and PCM_32 land on the identical value. One width-derived divisor is therefore correct for every width, and pydub already exposes it as `max_possible_amplitude`. A `sample_width`-conditional branch would be dead code.

**Tests.** `tests/test_load_audio_torchcodec_fallback.py` gains a parametrized case over all five subtypes. Fail-before / pass-after with the one-line revert:

```
before: 4 failed, 1 passed   (PCM_U8 0.001953, PCM_24/32/FLOAT 32768.000000)
after:  7 passed
```

**Unrelated correction, same push (23df0e80).** Three comments I added in the first commit claimed torch 2.8.0 carries no `sm_120` kernels. That is wrong for this repo's pin: `pyproject.toml:282` resolves torch from the cu128 index, and on an `sm_120` device that build reports

```
arch_list  ['sm_70', 'sm_75', 'sm_80', 'sm_86', 'sm_90', 'sm_100', 'sm_120']
capability (12, 0)   cuda_matmul OK
```

matching `CU128_ARCHS` in `tests/test_cuda_arch_compat.py`. The comments now state the forcing function that is actually verifiable: torch 2.8.0 publishes no aarch64 wheel, so arm64 CUDA hosts reach torchaudio 2.9 with no 2.8.0 option at all. Comments only, no behaviour change. The same stale claim appears in several pre-existing files (`backend/main.py:633`, `tests/test_torchaudio_removed_api_guard.py:9`, `tests/test_torchaudio_legacy_backend_1931.py:9`); left alone as out of scope, happy to send a separate PR.

Moep90:
Reproduced and verified on x86_64, so this is not arm64-specific. Fedora, RTX PRO 5000 Blackwell (sm_120), torch/torchaudio 2.9.1+cu128 installed over the repo checkout, **torchcodec deliberately not installed** — `pyproject.toml` and `uv.lock` untouched.

Baseline on the pinned torch 2.8.0+cu128, same machine: `torchaudio.save` to a BytesIO succeeds (4080 bytes). The bug is a 2.9 regression, not a pre-existing gap.

After moving only the torch trio to 2.9.1, raw torchaudio raises on both sides:

```
raw save   ImportError -> TorchCodec is required for save_with_torchcodec. Please install torchcodec to use this function.
raw load   ImportError -> TorchCodec is required for load_with_torchcodec. Please install torchcodec to use this function.
```

Both messages are verbatim what the arm64 host produced. In that same interpreter, with the branch applied:

```
tests/test_load_audio_torchcodec_fallback.py + tests/backend/services/test_audio_io.py   39 passed
tests/backend/services/test_audio_io.py -k cuda                                           6 passed, 26 deselected
```

So the fallbacks carry the suites on a real torchaudio 2.9 with TorchCodec genuinely missing, on CUDA, on x86_64. The environment that makes stock torchaudio raise is the one where the branch is green.

PR #2143, opened 2026-09-16
fix: Argos language names and BCP-47 tags map to ISO 639-1

### LINKED ISSUE #2140: Dubbing to Chinese fails on Argos: "No Argos package available for en -> Chinese" — target_lang is never normalized to Argos's ISO 639-1 codes
### Environment
- VoiceStudio v0.5.2 (macOS installer), Apple Silicon / macOS 26
- Dubbing → Translation engine: **Argos** (offline default), target language: **Chinese**

### What happens
Every segment fails with:

```
13/13 段失败：No Argos package available for en -> Chinese
```

### Root cause
In `backend/api/routers/dub_translate.py`, the Argos branch picks the target code raw:

```python
to_code = seg.target_lang if seg.target_lang else req.target_lang
```

But Argos packages are keyed by bare ISO 639-1 codes (`en` → `zh`). Meanwhile the value that arrives can be:

1. a human name — `"Chinese"` (restored projects / per-segment overrides can carry names),
2. the UI dropdown's **own** value — the dub target-language list (`frontend/src/utils/languages.js`) uses BCP-47-ish codes, so picking Chinese sends `"cmn-Hans"`,
3. a regional code like `"zh-CN"`.

None of these match `"zh"`, so **both** the installed-package lookup and the auto-install index lookup fail — even when the user picks the dropdown's own value. Every other provider in the same file normalizes first (`TRANSLATE_CODES`, `FLORES_CODES`, `LANG_NAMES`); Argos is the only one fed raw values.

### Verified against a live backend (v0.5.2)
| `target_lang` | Result |
|---|---|
| `"Chinese"` | `No Argos package available for en -> Chinese` (the reported error) |
| `"cmn-Hans"` | `No Argos package available for en -> cmn-Hans` (dropdown value, still broken) |
| `"zh"` | works — auto-installs `en→zh` and translates correctly |

### Suggested fix
Normalize both `from_code` and `to_code` before lookup, e.g.:

```python
_ARGOS_CODE_ALIASES = {"cmn": "zh", "zho": "zh", "in": "id", "iw": "he", "fil": "tl",
                       "chinese": "zh", "chinese (traditional)": "zh"}
_NAME_TO_ARGOS_CODE = {}
for _code, _name in LANG_NAMES.items():
    _NAME_TO_ARGOS_CODE.setdefault(_name.lower(), _code.split("-")[0])

def _norm_argos_code(raw):
    key = (raw or "").strip().lower()
    if not key:
        return raw
    if key in _ARGOS_CODE_ALIASES:
        return _ARGOS_CODE_ALIASES[key]
    name_code = _NAME_TO_ARGOS_CODE.get(key)
    if name_code:
        return name_code
    base = key.split("-")[0]
    return _ARGOS_CODE_ALIASES.get(base, base)
```

then `from_code = _norm_argos_code(src_lang)` and `to_code = _norm_argos_code(seg.target_lang if seg.target_lang else req.target_lang)`. Related but distinct: #477 fixed the *source*-code variant (`en -> en`); this is the *target*-code side of the same class.

I have this implemented and tested locally (happy to open a PR if useful).

gyanu2507:
Fixes #2140

Argos Translate ships pairs as `zh`, not `Chinese`. The dub UI sends the display name (and sometimes `cmn-Hans` / `zh-CN`). `get_translation_from_codes` then returns None and every subtitle stays in the source language.

`argos_lang_code` now maps those names and BCP-47 tags down to ISO 639-1 before lookup. Covered in `tests/test_dub_translate.py`.

Argos language normalization now maps names, BCP-47 tags, script codes, and aliases to ISO 639-1 codes, including `Chinese`, `cmn-Hans`, and `zh-CN` to `zh`. This fixes installed-package and auto-install lookups for valid Argos languages. Tests cover the new mappings; no additional merge risk is identified from the supplied evidence.


rollroyces:
Hey @gyanu2507 — your fix looks clean. I opened #2152 as a small follow-up that adds the missing test cases CodeRabbit flagged (Chinese Traditional / Mandarin / zho / in→id / iw→he / fil→tl + a ValueError test for empty/whitespace input). All 47 dub_translate tests pass locally. If you'd rather pull these into this branch instead of merging a separate PR, the only commit is on `rollroyces:fix/argos-lang-names-tests` and cherry-picks cleanly onto your tip. Either way, thanks for the diagnosis and the fix.

PR #2152, opened 2026-09-16
test(dub): cover remaining argos_lang_code aliases and empty input (#2143)

rollroyces:
## Summary

Extends the test coverage of the `argos_lang_code` normalizer introduced in #2143 so every new branch and every new error path is exercised.

## Changes

- **`tests/test_dub_translate.py`** — Extends the existing parametrized normalization test with the remaining aliases (`Chinese (Traditional)`, `Mandarin`, `zho`, `in`→`id`, `iw`→`he`, `fil`→`tl`) and adds a separate parametrized test asserting `ValueError` for empty and whitespace-only input. All assertions document what the input represents (ISO 639-1, BCP-47, deprecated ISO 639-1, ISO 639-2/T, display name).

No production code changes — pure test addition on top of #2143.

## Why this PR

CodeRabbit on #2143 flagged a 🟡 Minor — Quick win: "this parametrization omits `zho`, `in`, `iw`, `fil`, `Chinese (Traditional)`, `Mandarin`, and blank input, so regressions in newly added aliases or validation will pass CI." This PR adds the missing cases so the test suite matches the branches the function introduces.

Issue #2140 (Argos `target_lang` not normalized) is fixed by #2143; this PR only hardens its test coverage. The `Closing issues` line is intentionally left blank for that reason.

## Type

- [x] Test only

## Testing

- `uv run --frozen pytest tests/test_dub_translate.py -q` → 47 passed (was 33 on `main` before #2143, +14 new test cases)
- `uv run --frozen pytest tests/test_dub_translate.py -k argos_lang_code -v` → 15 passed (11 existing parametrized cases + 4 new `ValueError` cases across the empty/whitespace branch)

The 14 pre-existing failures in the wider `tests/` suite (`test_audiobook_expressive`, `test_longform_synth_params`, `test_socks_proxy`, `test_broken_dependency_diagnosis_1415`, etc.) reproduce on `main` and on `gyanu2507:fix/argos-lang-names` without this PR — they require local model installation and are unrelated.

## Checklist

- [x] Forked the repo and branched from `main` (branched from `gyanu2507:fix/argos-lang-names`, which is itself branched from `main`)
- [x] One feature/fix per PR
- [x] PR title uses Conventional Commit format with scope, referencing the issue it extends
- [x] All template sections filled
- [x] Tests added that fail without the source change in #2143 and pass with it

Adds coverage for `argos_lang_code` normalization, including ISO aliases, BCP-47 tags, display names, and invalid inputs. The targeted test file reports 50 passing tests. A human should verify the existing Traditional Chinese script mapping because this PR preserves that behavior.


rollroyces:
### Review-bot replies

**Greptile P1 — "Traditional Chinese Uses Wrong Script":** Valid finding, but it lives in `backend/services/translation_engines.py`, which is gyanu2507's source change from #2143, not this PR's. The right fix is a script-aware mapping (e.g. `Chinese (Traditional)` → Argos `zho_Hant`, the same `_nllb_language` mapping already used for NLLB) — that's a product decision and a behaviour change to the Argos contract, not a test-coverage tweak. I'm leaving it for @debpalash / @gyanu2507 to scope, since I'm trying to keep this PR test-only on top of #2143.

If you want me to take a swing at the script-aware fix here, say so and I'll open a separate branch with the Argos lookup mapping `zh-Hant`/`Chinese (Traditional)` → `zho_Hant` and add tests for it. Otherwise this PR stays narrow.

**CodeRabbit 🟡 — "redundant blank-input branch":** Fixed in 50c6bfb — dropped the explicit `if not raw: raise` line (the `re.fullmatch` two lines below raises the same message for any non-code input, so the shortcut was dead code). Expanded the docstring to enumerate every alias shape. Widened the empty-input test to also assert `"Chinese ("` (a non-empty but invalid label-shaped input) so the test asserts behaviour unique to the regex branch instead of passing for the wrong reason.

Local check: `uv run --frozen pytest tests/test_dub_translate.py -q` → 50 passed (was 47 before the latest commit).

PR #2190, opened 2026-09-17
fix(encoding): read pyproject.toml and quant_map.json as UTF-8

kevin9327:
Read bundled pyproject.toml and quant_map.json explicitly as UTF-8 so source startup and generation work on CJK Windows code pages. The corresponding guards and regression fixtures use the same portable decoding.

Current main is merged and the changelog conflict is resolved with contributor credit. Both CodeRabbit and Greptile reviews were checked; neither has an outstanding code finding.

Validation: 50 offline locale-decoding, version, GGUF and changelog tests passed. The contributor’s regressions exercise cp932/cp936/cp949/cp950 without requiring a Windows host.

Bundled `pyproject.toml` and `quant_map.json` reads now use explicit UTF-8 decoding, with portable regression coverage for CJK Windows code pages. This prevents source startup and generation failures caused by locale decoding. No outstanding review findings remain.


kevin9327:
Heads-up on the CHANGELOG: `## [Unreleased]` is empty on `main`, so this PR and my siblings (#2190 / #2191 / #2192) each recreate the Highlights + Fixed scaffold and will collide there. Whichever lands first, I'll re-file the others onto its block — or just drop my two lines at merge time and I'll follow up.

PR #2191, opened 2026-09-17
fix: harden imports, media exports, diagnostics and NPU cleanup

kevin9327:
EPUB imports, media exports and error reports could fail or lose useful information on otherwise supported installations. This maintenance batch preserves the contributor fixes and their review corrections:

- Decode EPUB chapters using BOMs, declared encodings and BOM-less UTF-16/UTF-32, with safe fallback for unsupported codecs (#2191).
- Resolve the app's bundled FFmpeg for video watermarks and dubbing keyframes, preserving existing failure fallbacks (#2192).
- Preserve and scrub backend exception classes through Electron streaming failures and the shared bug-report builder (#2197).
- Clear Ascend NPU allocator caches and recognize dedicated NPU memory while preserving other accelerator paths (#2194).

Current main is merged into every source branch. This PR includes their merge commits in the order above, so landing it also absorbs #2192, #2197 and #2194 without losing contributor history. Review findings were corrected on the source branches before combining. Documentation and changelog credits are included.

Validation: 119 offline backend/changelog tests, 9 shared report-builder tests, 26 Electron generation tests, and both Electron TypeScript checks pass on the combined tree. Full combined CI and post-merge main checks remain required. No physical Ascend hardware or live model generation was used.

Review disposition: the changelog entries stay under `### Fixed`, which CLAUDE.md explicitly permits beneath a short Highlights list. The request to move every entry into Highlights conflicts with that governing rule; the deterministic changelog tests pass.


kevin9327:
Both Greptile P1s were real and are fixed in 28ef19d.

**UTF-32 BOMs** — the table lives in `services/text_upload.py` and is shared, so `/dub/import-srt` and the `.txt`/`.md` branch of `/audiobook/import` carried the same corruption. Fixed once, longest mark first, and all three paths inherit it. `tests/test_text_upload_encoding.py` now parametrises `utf-32-le-bom` / `utf-32-be-bom` alongside the existing encodings (4 new fail-before cases), plus an EPUB case.

**Non-text codec** — confirmed: `codecs.lookup("hex_codec")` resolves, `b"...".decode("hex_codec")` raises `LookupError: 'hex_codec' is not a text encoding`, and it was uncaught. The `codecs.lookup` pre-check is gone entirely — the decode itself decides, and an encoding that cannot produce text falls back the way an undeclared document does. Parametrised over `x-not-a-real-charset` and `hex_codec`.

Fail-before on the six new cases, then 289 passed across `test_longform_import`, `test_text_upload_encoding`, `test_longform_{e2e,parser,limits,render,jobs,resume}`, `test_srt_parser` and `test_dub_parse_subtitle_text`.

Heads-up on the CHANGELOG: `## [Unreleased]` is empty on `main`, so this PR and my #2190 / #2192 each recreate the Highlights + Fixed scaffold. Whichever lands first, I'll re-file the other two onto its block — just say the word (or drop my two lines and I'll follow up).

kevin9327:
CodeRabbit's BOM-less UTF-16 finding is valid and is fixed in 6398f01.

UTF-16 without a mark is out of spec (XML 1.0 §4.3.3 requires one), but real EPUBs carry it — and in such a document the declaration is unreadable anyway, because the ASCII byte patterns never match NUL-interleaved bytes, so the entry fell through to the UTF-8/Windows-1252 guess and narrated the markup. XML 1.0 Appendix F reads the opening `<` instead: its byte pattern names the width and byte order. Widest first, since `<\0\0\0` also starts `<\0`, so UTF-32 can't be misread as UTF-16.

Parametrised over utf-16-le/be and utf-32-le/be, each asserting the fixture really has no mark. 4 fail-before cases; 293 passed across the longform, text-upload and subtitle-parser suites.

Decoder order is now: byte-order mark → no-mark width sniff → declared encoding → guess (UTF-8, else Windows-1252).

Issue #2177, opened 2026-09-17
[Bug] Generation failed. Check the selected engine and try again.


OtterBeWorking:
## Error

```
Generation failed. Check the selected engine and try again.
Backend error class: RuntimeError
StreamingPreviewError: Generation failed. Check the selected engine and try again.
    at f (http://tauri.localhost/assets/streamingTts-Wz3Wpui2.js:1:6035)
    at O (http://tauri.localhost/assets/streamingTts-Wz3Wpui2.js:2:54)
    at async o (http://tauri.localhost/assets/generate-CLzVoZMB.js:1:263)
    at async http://tauri.localhost/assets/main-app-Cz6IkjIW.js:16:31825
```

## Environment

**Version:** `0.5.3`
**Platform:** `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/153.0.0.0 Safari/537.36 Edg/153.0.0.0`
**OS:** `Windows-10-10.0.26200-SP0`
**Python:** `3.11.15`
**Compute device:** `cuda`
**GPU:** `NVIDIA GeForce RTX 5060 Ti (15.9 GB VRAM)`
**CPU:** `AMD Ryzen 5 5600 6-Core Processor`
**RAM:** `31.9 GB`
**Disk free:** `658.7 GB`
**Active TTS engine:** `cosyvoice`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 0 s before this report

## Recent actions

```
09:06:59 view:catalogue
09:16:54 engine:tts=cosyvoice
09:16:58 engine:tts=cosyvoice
09:17:16 view:studio
09:19:57 generate:start (audio)
09:21:12 generate:stream-retryable-abort
09:21:24 generate:start (audio)
09:21:30 generate:stream-retryable-abort
```

rollroyces:
Thanks for the report. Recent merges on `main` have likely fixed this class of failure — your error is the symptom of an engine subprocess that crashed at startup or went silent mid-generation. The diagnostic now names the cause (exit code, stderr tail, deadline vs hang) instead of a generic 500, so a fresh build will say *what* failed.

Could you update VoiceStudio to the latest release (or `main`) and rerun the same generate? If the new build reproduces the same failure, please paste the full error text — it now carries enough context to identify the next class of bug. If it succeeds, this issue is stale and can be closed as already-fixed.

Relevant fixes:
[...]

Issue #1943, opened 2026-09-09
[Bug] 500 Internal Server Error: VoiceStudio hit an internal error; check the backend


elMan-L:
## Error

```
500 Internal Server Error: VoiceStudio hit an internal error; check the backend log for details. The connection to the video server dropped mid-download (often a transient CDN/network blip or a regional rate-limit). Just retry — VoiceStudio already cleaned up the partial download. If it keeps failing, check your network/VPN.
$@tauri://localhost/assets/client-CNTJBHFd.js:5:13265
```

## Environment

**Version:** `0.5.1`
**Platform:** `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)`
**OS:** `macOS-26.6.2-arm64-arm-64bit`
**Python:** `3.11.12`
**Compute device:** `mps`
**GPU:** `Apple Silicon (MPS)`
**CPU:** `Apple M3`
**RAM:** `16 GB`
**Disk free:** `11.2 GB`
**Active TTS engine:** `mlx-audio`

## Backend reachability

**Deployment mode:** `desktop`
**Last backend response:** 0 s before this report

## Last backend crash (auto-captured — may predate this bug)

**When:** 2026-09-09T03:30:35.000Z (22 min ago)
**Exit:** `exit code 3`
**Uptime before crash:** 4 s
**Backend version:** `0.5.1`

```
… (truncated)
    Started server process [96543]
INFO:     Waiting for application startup.
2026-09-09 11:30:31,762 INFO [omnivoice.api] Startup watchdog armed: thread dump if startup exceeds 300s (#632).
2026-09-09 11:30:31,763 WARNING [omnivoice.run_sentinel] run_sentinel.json belongs to a live process (pid 5801) — another instance shares this data dir; skipping crash detection and sentinel ownership for this run.
INFO:     Application startup complete.
ERROR:    [Errno 48] error while attempting to bind on address ('127.0.0.1', 3900): address already in use
INFO:     Waiting for application shutdown.
2026-09-09 11:30:32,806 INFO [omnivoice.hf_progress] hf_progress: installed tqdm patch on huggingface_hub.utils.tqdm
2026-09-09 11:30:33,045 INFO [omnivoice.setup.models] Loaded 36 models from ~/Library/Application Support/com.debpalash.omnivoice-studio/project/backend/config/models.yaml
[...]

PR #2195, opened 2026-09-18
fix(generate): name the cause of a streaming generation failure

shivsin25:
## Summary

Related to #2177. Recognized GPU build mismatches, Windows application-control blocks, and audio-file failures now provide actionable recovery guidance instead of the generic streaming failure message. GPU build and policy failures are terminal, preventing an automatic second render of the same passage.

This improves diagnosis; the reporter's underlying CosyVoice failure is still unconfirmed without backend logs. It does not claim to fix every RuntimeError or establish a GPU mismatch on the reported machine.

## Changes

- Classify CUDA's unsupported-kernel signature; allow existing Windows policy and audio I/O diagnoses through the privacy-safe streaming response. Unknown errors and OOM behavior remain unchanged; exception text is never copied into public error frames.
- Translate stable diagnosis topics in Electron and web HTTP/streaming clients, including Electron pipeline failures, with real translations in all 21 locales. Canonical English API fallback guidance remains available to older clients.
- Preserve the backend exception class in Electron streaming errors, matching the web client.
- Address review findings: import backend modules in runtime test fixtures, replace private dynamic imports and empty exception handling with a public terminal-topic predicate, and link recovery documentation. Credit @shivsin25 in the changelog.
- Merge current main into the contributor branch before validation.

## Type

- [x] Bug fix
- [x] Tests
- [x] Documentation

## Testing

- Backend diagnosis/privacy tests: 65 passed with offline Hugging Face mode and an empty model cache.
- Backend plus locale, changelog, and source-string checks: 589 passed (existing locale-baseline warnings only).
- Electron targeted tests: 46 passed; locale checks and both TypeScript projects passed.
- Web targeted tests: 57 passed; TypeScript passed.
- Regression checks reproduced failures before the localization and Electron error-class fixes and pass afterward.
- Full cross-platform CI passed: backend/frontend/Electron tests, production and renderer smokes, Linux native packaging, all four OS smoke jobs, all three legacy shell checks, and Windows MSI authoring. Security checks and both bot review checks passed on c101367d.

## Checklist

- [x] Tested locally
- [x] Updated recovery documentation and changelog
- [x] No version or dependency changes
- [x] No release publishing requested by this PR

## Review disposition

CodeRabbit's updated review found no actionable code issues and passed the 21-locale completeness check. Greptile's remaining thread concerns the canonical English API fallback: updated Electron and web clients render translated topic guidance, while keeping that fallback preserves compatibility with older clients. Regression tests cover both HTTP response shapes and streaming errors.

The automated parity warning is not a feature divergence: classification and terminal handling do not branch on the client or backend OS. A Windows worker's policy diagnosis is handled identically by macOS, Windows, Linux, and web clients. Adding speculative mappings for unrelated OS errors would risk incorrect recovery guidance. The test-function docstring percentage is advisory; the new public helpers document their contract.


shivsin25:
The report can't name a cause, and that's the bug rather than missing detail
from the reporter. I reproduced this exact payload from `main`:

    >>> stream_generation_failure(RuntimeError("CUDA error: no kernel image is available..."))
    {'code': 'generation_failed',
     'detail': 'Generation failed. Check the selected engine and try again.',
     'retryable': True, 'error_class': 'RuntimeError'}

The streaming error frame never copies exception text, so everything it can say
comes from the failure taxonomy — and 10 of the 12 classes
`_oom_friendly_reraise` explains on the classic path have no reachable topic, so
they all render this one message. Any of them produces a byte-identical report.

Two things follow from the action log:

- `generate:stream-retryable-abort` twice, 90s apart — the floor message ends in
  "try again", and `retryable: true` is set even for classes that can never
  succeed on a retry.
- Because the message is identical across ~10 causes, `error_class: RuntimeError`
  is the only distinguishing field, and it isn't enough to triage.

#2195 adds the missing topics for the three classes whose triggers are
unmistakable machine strings (`no kernel image` → new `GPU_ARCH_UNSUPPORTED`,
plus allowlisting `WINDOWS_APP_CONTROL_BLOCKED` and `AUDIO_IO_FAILED`), and marks
the build/policy classes terminal so the app stops re-rendering to reach the same
error. The generic-English classes are deliberately left alone — that's the #1943
wrong-remedy risk.

To be clear about scope: I can't tell which `RuntimeError` **this** reporter hit,
and §5b argues against the obvious Blackwell guess (torch 2.8.0+cu128 does list
`sm_120`, and their backend stayed healthy). #2195 fixes the reason it's
unknowable, not that specific failure.

@OtterBeWorking — if you still have it, `backend_err.log`
(Settings → Logs → Backend) around 09:19–09:21 would have the real exception, and
would confirm whether the new `GPU_ARCH_UNSUPPORTED` class covers your case.

OtterBeWorking:
Thanks for digging into this! I don't have backend_err.log anymore (deleted it before I saw this), but I do still have the error_journal.jsonl entry from that exact window:

first_seen: 2026-09-17T09:21:12
last_seen: 2026-09-17T09:30:41
count: 5
RuntimeError: cosyvoice sidecar synthesize error: ErrorDuringImport: problem in cosyvoice.flow.flow_matching - ModuleNotFoundError: No module named 'gdown'

So to close the loop: this isn't GPU_ARCH_UNSUPPORTED — your doubt about the Blackwell guess was right, my torch 2.8.0+cu128 build does list sm_120 and the backend stayed healthy. The actual cause is that backend/engines/cosyvoice_subprocess/requirements.txt drops gdown and wget (its own comments say "not needed to import or run CosyVoice"), but cosyvoice.flow.flow_matching imports gdown directly at module load, so the sidecar dies before it can log anything useful to the Runtime panel — that's also why the panel looked empty and the health check just said "returned-unavailable, details withheld."

I manually installed gdown, wget, pyarrow, pyworld, and downgraded setuptools to <81 (for pkg_resources) directly inside CosyVoice's own .venv. That did not fix it — Test Engine still failed every single time afterward, same as before. So either there's a second import failing behind gdown, or something else entirely is wrong. I haven't been able to get a fresh error out of it since (health-check probes withhold the real exception, and a real generation attempt is what actually populates the error journal — I'm still working on getting a current traceback). Wanted to flag this now so the missing-gdown fix isn't assumed to be a full resolution, and in case it's a useful data point for classify() either way.

PR #2200, opened 2026-09-18
fix(gpt-sovits): api_v2 probe accepts 400/405, default reference for plain TTS, no server-side splitting

### LINKED ISSUE #2102: [Bug] GPT-SoVITS adapter speaks the api.py v1 protocol, so it cannot talk to the api_v2 server the docs tell users to start
## What happened

I started the GPT-SoVITS server with the exact command this repo documents:

```
python api_v2.py -a 127.0.0.1 -p 9880 -c GPT_SoVITS/configs/tts_infer.yaml
```

The server came up healthy. VoiceStudio still reported it as missing:

```
is_available: (False, 'GPT-SoVITS server not reachable at http://127.0.0.1:9880.
Start it with: python api_v2.py -a 127.0.0.1 -p 9880 -c GPT_SoVITS/configs/tts_infer.yaml')
```

Forcing a generation past the probe then failed:

```
OSError: endpoint returned HTTP 404
RuntimeError: GPT-SoVITS API call failed: endpoint returned HTTP 404.
Ensure the server is running at http://127.0.0.1:9880
```

Server side, both were plain 404s on a route that does not exist:

```
INFO:     127.0.0.1:64653 - "GET / HTTP/1.1" 404 Not Found
```

## Root cause

The adapter was written against `api.py` (v1) and the docs tell users to run `api_v2.py`. The two do not share a route or a parameter schema.

**Probe.** `GPTSoVITSBackend.is_available()` (`backend/services/tts_backend.py:2060`) does `GET <base url>`. `api_v2.py` has no `/` route, so a perfectly healthy server answers 404, `open_trusted_endpoint` raises, and the except branch reports "server not reachable". The message is also wrong in kind: the server is reachable and answering.

**Generation.** `generate()` (`:2034`) posts a query string to the base URL with v1 names:

| adapter sends | api_v2 expects |
| --- | --- |
| `POST /` | `POST /tts` |
| query string | JSON body |
| `text_language` | `text_lang` |
| `refer_wav_path` | `ref_audio_path` |
| `prompt_language` | `prompt_lang` |

`api_v2.py:307-320` reads `text_lang` and `ref_audio_path` and returns 400 when either is absent.

## Verification

The server is fine. Posting the v2 schema directly produced a clean 28.6 s English render of the same script the adapter could not get:

```python
body = {"text": ..., "text_lang": "en", "ref_audio_path": ..., "prompt_text": ...,
        "prompt_lang": "en", "text_split_method": "cut5", "media_type": "wav",
        "streaming_mode": False}
urllib.request.Request("http://127.0.0.1:9880/tts", data=json.dumps(body).encode(),
                       headers={"Content-Type": "application/json"}, method="POST")
```

Polish and Spanish were refused by the server with a correct, clear message (`text_lang: pl is not supported in version v2`), which matches the engine's declared `["zh", "en", "ja", "yue", "ko"]`.

## Suggested fix

Pick one protocol and make the docs, the probe and the request agree.

- Target api_v2 (what `docs/engines/gpt-sovits.md` and the unavailability message both tell users to run): probe `GET /tts` or another route that exists, and post JSON to `/tts` with `text_lang` / `ref_audio_path` / `prompt_lang`.
- Separately, the probe should not translate "reachable but 404" into "not reachable" — a connection refused and a routing mismatch need different messages, or the user has no way to tell a stopped server from a protocol mismatch.

## Two smaller things found on the way

Neither is a VoiceStudio bug, but both cost time and could go in the engine doc:

- api_v2 loads reference audio through torchcodec, which needs FFmpeg 4-7. A host with FFmpeg 9 fails with "Could not load libtorchcodec"; `brew install ffmpeg@7` plus `DYLD_FALLBACK_LIBRARY_PATH` fixes it. (`nohup` strips `DYLD_*` on macOS, so the variable must be set on the server process directly.)
- It also needs the upstream NLTK data bundle, or it 400s with `Resource 'averaged_perceptron_tagger' not found`.

## Environment

**Version:** `0.5.2` (main @ `eaf8bb9`)
**OS:** macOS (Darwin 25.5.0), Apple M4 Pro, 24 GB unified memory
**GPT-SoVITS:** `RVC-Boss/GPT-SoVITS` HEAD, v2 pretrained weights, running on CPU

jaketame:
## Summary
Builds on #2180 by @rollroyces (api_v2 protocol; its commit is included here unchanged) and closes the gaps the review bots and a live server turned up. Closes #2102.

- **Probe (P1 on #2180):** `is_available()` still routed `GET /tts` through `open_trusted_endpoint`, which raises on every 4xx/5xx. A healthy api_v2 answers a parameterless GET with 400 or 405, so the server read as "not reachable". New `probe_trusted_endpoint` shares the pinned-connection setup but only returns the status; 404 keeps meaning "api.py v1 server", anything else is reachable.
- **Plain TTS:** api_v2 has no default voice (api.py's `-dr/-dt/-dl` were v1 only) and answers 400 without `ref_audio_path`, so a request with no voice profile failed with an opaque "API call failed: HTTP 400". The adapter now clones a configured default clip (`OMNIVOICE_GPTSOVITS_REF_AUDIO` / `_REF_TEXT` / `_REF_LANG`) when the request has none, and fails before the request, naming those variables, when there is no clip at all. An explicit clip on the request still wins.
- **Splitting:** requests are sent with `text_split_method: cut0`. VoiceStudio already chunks long text (`split_text_into_chunks`); letting the server re-split at every punctuation mark (`cut5`) produced short fragments on which the GPT stage ended early and dropped whole clauses (reproduced 1 in 2 seeds on a two-sentence input; cut0 rendered both seeds completely).

Docs: protocol, the default-reference variables, fine-tuned weights via `tts_infer.yaml`, and the near-silent-output symptom (clip ending in silence / transcript mismatch).

## Verification
- Tests: 400/405 probes → available, 404 → wrong-protocol message, connection refused → not-reachable message, probe helper returns any status and closes the connection, default-reference body, explicit-clip precedence, early no-reference error, request shape with `cut0`.
- Live, against `api_v2.py` with v2ProPlus fine-tuned weights on a Mac, through the real adapter: `is_available()` → True; plain TTS via env default clip renders; clone path renders; no-ref → clear error; pointing `OMNIVOICE_GPTSOVITS_URL` at an `api.py` v1 server → "does not expose api_v2's /tts route". Renders ASR-verified.
- `tests/test_gptsovits_endpoint_security.py`, changelog/engine-docs/CJK guard tests pass locally.

Co-Authored-By: Claude <noreply@anthropic.com>

Migrates the GPT-SoVITS adapter to the `api_v2` `/tts` JSON protocol with status-aware probing, required reference audio, default reference configuration, and unsplit requests. This fixes compatibility with documented `api_v2.py` servers and distinguishes unavailable servers from v1 protocol mismatches. Human review should verify that `cut0` and default reference settings produce acceptable output across supported languages and server configurations.


jaketame:
Addressed Greptile's readiness finding in ec1195af: readiness now accepts only api_v2's own answers to a parameterless GET /tts (200/400/405/422); 404 still reads as an api.py v1 server, and any other status (proxy, auth wall, redirect, 5xx) is reported with its code and treated as unavailable. Tests cover 302/401/403/500/502.

jaketame:
fed3049c: a live api_v2 answers a *bare* GET /tts with 500 (its handler lower-cases text_lang/prompt_lang before validating), which the readiness whitelist would have rejected. The probe now sends the smallest query that reaches api_v2's own validator (text=&text_lang=en&prompt_lang=en → 400 "ref_audio_path is required", no work done); api.py v1 still answers 404 and proxies/auth walls/crashes keep their own codes. Verified against a running server.

